PR #9, opened 2017-08-15
Modified applygain to support visibilities/gains in axes other than 1.

gooddc:


draco/util/tools.py
```diff
@@ -109,11 +109,13 @@ def apply_gain(vis, gain, axis=1, out=None, prod_map=None):
...
 
         # Fetch the gains
-        gi = gain[:, ii]
-        gj = gain[:, ij].conj()
+        gi = gain[[slice(None) for i in range(axis)] + [ii]]
+        gj = gain[[slice(None) for i in range(axis)] + [ij]].conj()
 
         # Apply the gains and save into the output array.
-        out[:, pp] = vis[:, pp] * gi * gj
+        vis_slice = [slice(None) for i in range(axis)] + [pp]
```
jrs65: You should rewrite this to remove the repetition, i.e you construct essentially the same list of slice objects in three places. You can even create this outside the loop for extra kudos.

PR #10, opened 2018-06-21
Gain sims

cahofer:
Tidying up the pipeline. 
There is some code duplication in the mapmaker module but I could not think of a smart way combining the m-mode map maker with svd map maker. If you have ideas, I am open to develop.

### DIFF
--- a/draco/analysis/fgfilter.py
+++ b/draco/analysis/fgfilter.py
@@ -80,7 +80,7 @@ class SVDModeProject(_ProjectFilterBase):
         # Iterate over local m's, project mode and save to disk.
         for lm, mi in mmodes.vis[:].enumerate(axis=0):
 
-            tm = mmodes.vis[mi].reshape(tel.nfreq, 2 * tel.npairs)
+            tm = mmodes.vis[mi].transpose((1, 0, 2)).reshape(tel.nfreq, 2 * tel.npairs)
             svdm = bt.project_vector_telescope_to_svd(mi, tm)
 
             if np.isfinite(svdm).all() is False:
--- a/draco/analysis/fgfilter.py
+++ b/draco/analysis/fgfilter.py
@@ -74,6 +74,8 @@ class SVDModeProject(_ProjectFilterBase):
 
         svdmodes = containers.SVDModes(mode=bt.ndofmax, axes_from=mmodes,
                                        attrs_from=mmodes)
+        svdmodes.vis[:] = 0.0
+
         mmodes.redistribute('m')
         svdmodes.redistribute('m')
 
@@ -82,14 +84,6 @@ class SVDModeProject(_ProjectFilterBase):
 
             tm = mmodes.vis[mi].transpose((1, 0, 2)).reshape(tel.nfreq, 2 * tel.npairs)
             svdm = bt.project_vector_telescope_to_svd(mi, tm)
-
-            if np.isfinite(svdm).all() is False:
-                print "M MODE", mi
-                print np.isfinite(svdm).all(), "NOT FINITE"
-
-            if np.isnan(svdm).any() is True:
-                print "M MODE", mi
-                print np.isnan(svdm).any(), "FOUND A NAN"
 
             svdmodes.nmode[mi] = len(svdm)
             svdmodes.vis[mi, :svdmodes.nmode[mi]] = svdm
@@ -184,6 +178,9 @@ class KLModeProject(_ProjectFilterBase):
         # Construct the container and redistribute
         klmodes = containers.KLModes(mode=bt.ndofmax, axes_from=svdmodes,
                                      attrs_from=svdmodes)
+
+        klmodes.vis[:] = 0.0
+
         klmodes.redistribute('m')
         svdmodes.redistribute('m')
 
--- a/draco/analysis/fgfilter.py
+++ b/draco/analysis/fgfilter.py
@@ -83,6 +83,14 @@ class SVDModeProject(_ProjectFilterBase):
             tm = mmodes.vis[mi].reshape(tel.nfreq, 2 * tel.npairs)
             svdm = bt.project_vector_telescope_to_svd(mi, tm)
 
+            if np.isfinite(svdm).all() is False:
+                print "M MODE", mi
+                print np.isfinite(svdm).all(), "NOT FINITE"
+
+            if np.isnan(svdm).any() is True:
+                print "M MODE", mi
+                print np.isnan(svdm).any(), "FOUND A NAN"
+
             svdmodes.nmode[mi] = len(svdm)
             svdmodes.vis[mi, :svdmodes.nmode[mi]] = svdm
 
--- a/draco/analysis/mapmaker.py
+++ b/draco/analysis/mapmaker.py
@@ -72,7 +72,7 @@ class BaseMapMaker(task.SingleTask):
         bt_freq = self.beamtransfer.telescope.frequencies
         mm_freq = mmodes.index_map['freq']['centre']
 
-        freq_ind = [ find_key(bt_freq, mf) for mf in mm_freq]
+        freq_ind = [find_key(bt_freq, mf) for mf in mm_freq]
 
         # Trim off excess m-modes
         mmodes.redistribute('freq')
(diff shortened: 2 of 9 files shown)


draco/analysis/powerspectrum.py
```diff
@@ -60,31 +60,32 @@ def process(self, klmodes):
...
+        ps = containers.Powerspectrum2D(kperp_edges=pse.kperp_bands,
+                                        kpar_edges=pse.kpar_bands)
 
         npar = len(ps.index_map['kpar'])
         nperp = len(ps.index_map['kperp'])
 
         # Calculate the right unmixing matrix for each ps type
         if self.pstype == 'unwindowed':
-            M = la.inv(fisher)
+            M = la.pinv(fisher, rcond=1e-12)
```
cahofer: Should set rcond to 1e-8! or 1e-7

cahofer:
Hey Richard,

this is my updated pull request. I developed a bunch of files which contain different features. 
So you should regard this branch as the simulation pipeline fixing, map making developing, and complex gain sims restructure branch. Sorry, I will in the future be more organized with my changes!

The BaseMapMaker class is a bit nasty after all, I know we had a lot of copies for the different basis sets but since SVD modes don't have frequencies you have to throw in a bunch of if statements in order for code actually to work.  

Always open to suggestions and to learn better coding ;)

draco/analysis/fgfilter.py
```diff
@@ -74,13 +74,15 @@ def _forward(self, mmodes):
```
jrs65: Changes in this file look good.

draco/analysis/powerspectrum.py
```diff
@@ -60,31 +60,31 @@ def process(self, klmodes):
         q_list = []
```
jrs65: Changes look good here.

draco/analysis/sidereal.py
```diff
@@ -25,7 +25,7 @@
```
jrs65: Changes in the file look good.

draco/analysis/transform.py
```diff
@@ -306,6 +306,15 @@ class MModeTransform(task.SingleTask):
 
     Currently ignores any noise weighting.
```
jrs65: Changes in this file look good.

draco/core/containers.py
```diff
@@ -806,6 +806,53 @@ def input(self):
         return self.index_map['input']
```
jrs65: Looks good hear too.

draco/synthesis/stream.py
```diff
@@ -232,21 +232,19 @@ class MakeTimeStream(task.SingleTask):
 
     _cur_time = 0.0  # Hold the current file start time
```
jrs65: Changes here look good.

draco/synthesis/noise.py
```diff
@@ -53,6 +53,67 @@ def process(self, data):
...
+        nprod = len(data.index_map['prod'])
+
+        #with mpi_random_seed(self.seed):
+
+        # Calculate the number of samples
+        nsamp = int(self.ndays * dt * df)
+        std = self.recv_temp / np.sqrt(2 * nsamp)
+        ndraws = nfreq * ntime * nprod
+
+        std_real = std * np.random.standard_normal(ndraws).reshape(nfreq, nprod, ntime)
```
jrs65: These variable names are confusing. `std_` usually implies it's a standard deviation, there are actual realisations so call them something more appropriate.

draco/synthesis/noise.py
```diff
@@ -53,6 +53,67 @@ def process(self, data):
         return data
 
 
+class GaussianNoise(task.SingleTask):
```
jrs65: Please document the parameters below in an `Attributes` section to this doc string.

draco/synthesis/gain.py
```diff
@@ -43,84 +153,66 @@ class RandomGains(task.SingleTask):
...
 
-            # Generate amplitude and phase fluctuations consistent with the existing data
-            gain_amp = 1.0 + constrained_gaussian_realisation(time, cf_amp, nsamp,
-                                                              prev_time, prev_amp)
-            gain_phase = constrained_gaussian_realisation(time, cf_phase, nsamp,
-                                                          prev_time, prev_phase)
 
-        # Combine into an overall gain fluctuation
-        gain_comb = gain_amp * np.exp(1.0J * gain_phase)
+def generate_fluctuations(time, cf, num_realisations, prev_time, prev_fluc):
```
jrs65: Doc strings should have the format of:

```
"""Single line description.

A much longer description if needed.

Parameters
----------
x : int
  Parameter 1.

...
"""
```

draco/synthesis/gain.py
```diff
@@ -19,10 +18,121 @@
 from ..core import containers, task
 
 
-class RandomGains(task.SingleTask):
-    r"""Generate a random gaussian realisation of gain timestreams.
+class BaseGains(task.SingleTask):
+    r"""Rudimentary class to generate gain timestreams.
```
jrs65: This doesn't need to be a raw string (i.e. `r"""`). That's only needed below as it contains Latex source.

draco/synthesis/gain.py
```diff
@@ -19,10 +18,121 @@
 from ..core import containers, task
 
 
-class RandomGains(task.SingleTask):
-    r"""Generate a random gaussian realisation of gain timestreams.
+class BaseGains(task.SingleTask):
```
jrs65: This base class implementation looks nice.

draco/analysis/mapmaker.py
```diff
@@ -70,58 +73,79 @@ def find_key(key_list, key):
             except ValueError:
                 return None
 
-        # Figure out mapping between the frequencies
-        bt_freq = self.beamtransfer.telescope.frequencies
-        mm_freq = mmodes.index_map['freq']['centre']
+        if self.basis == 'mmodes':
```
jrs65: Don't almost all the things in both clauses of this `if ... else ...` block need doing for both `mmodes` and `svdmodes`? Or at the very least they can be safely done for both.

jrs65: In other words remove this if-else structure if you can.

draco/analysis/mapmaker.py
```diff
@@ -269,10 +291,23 @@ def _solve_m(self, m, f, v, Ni):
...
 
+        a_wiener = self._minimize(m, bmt, bth, vt)
+
+        # Copy the solution into a correctly shaped array output
+        a = np.zeros((bt.telescope.num_pol_sky, bt.telescope.lmax + 1), dtype=v.dtype)
+        a[:, m:] = a_wiener.reshape(bt.telescope.num_pol_sky, -1)
+
+        return a
+
+    def _minimize(self, m, beam, beam_conj, vec):
```
jrs65: Why is this called minimize? The name doesn't seem obvious to me.

draco/analysis/mapmaker.py
```diff
@@ -282,18 +317,114 @@ def _solve_m(self, m, f, v, Ni):
...
+        beam_svnum_conj = beam_svnum.T.conj()
+
+        # Solve for the dirty map alms
+        a = np.dot(beam_svnum_conj, fvec)
+        a = a.reshape(bt.telescope.num_pol_sky, -1)
+
+        return a
+
+
+class WienerMapMakerSVD(WienerMapMaker):
```
jrs65: Does this really need to be a separate sub class? It seems to me that you could probably merge them entirely. You just need to massage the projection matrices to be the same form. Same for `DirtyMapMaker`.

jrs65:
@cahofer this isn't really merged (despite what GitHub says), but I did pull out the changes I feel are ready to go and merged them in #10. Thanks for all those. Lots of bug fixes.

You'll need to carefully put the outstanding changes into a new branch (i.e. the changes in `mapmaker.py` and `transform.py`). 

On a related note, I don't understand necessity for the changes you made to `MModeTransform` (in `transform.py`), given the transform doesn't need any information about the telescope I'd rather not have the dependency in there. This is used for instance when doing the SVDFiltering which does a non-parametric transform in the m-mode space. If you need the array to be only `telescope.mmax` long you should probably ensure that in the place you actually need that (and not here).

cahofer:
Okay thanks! I think the reason I changed `transform.py` is when you have a sidereal time axis that is longer than 2 * `telescope.mmax` + 1 (maybe due to a higher time resolution, which could be convenient when simulating errors directly in sidereal time instead of solar time). The `SVDModeProject` step in that case will crash, because the m-beams are generated only up to `telescope.mmax`, yet in the fourier transform you have higher m's. I never used the `SVDFiltering` so I haven't considered that this could be a problem in another task. 
Maybe an idea is to slice the array in the `SVDModeProject` to `telescope.mmax`.

draco/analysis/mapmaker.py
cahofer: the if else structure can be mostly removed thats correct. 
line 79:  `mm_freq = modes.index_map['freq']['centre']`
line 80: `freq_ind = [find_key(bt_freq, mf) for mf in mm_freq]`
these two lines though are not working for svd modes because their index map has only m, and mode. we could add an frequency axis, although people have to aware that the number of svdmodes is different for every frequency and flatted on the `mode` axis. opinions?

draco/analysis/mapmaker.py
cahofer: maybe not the best name ... wanted to indicate that in this routine you are solving a linear matrix equation

draco/analysis/mapmaker.py
cahofer: Hmmmm, well I thought about this, we could probably merge them, but we would have to have some if else structures. Some of them include:

- 'beam_m' versus  'beam_svd'

- line 356: `svnum, svbounds = bt._svd_num(m)' for svdmodes

- when dealing with svdmodes we need to slice the beam transfer matrices and the data vector according to `svnum[f]`

- For Svd modes we need to create a noise covriance matrix `Ni` that is unity to merge the code.

Any comments?

PR #11, opened 2019-01-28
Updates for beam measurement processing

tristpinsm:
- Moved beam container definitions here from `ch_pipeline`

### DIFF
--- a/draco/analysis/delay.py
+++ b/draco/analysis/delay.py
@@ -111,7 +111,7 @@ class DelaySpectrumEstimator(task.SingleTask):
 
         tel = self.telescope
 
-        ss.redistribute('time')
+        ss.redistribute('freq')
 
         # Construct the Stokes I vis
         vis_I, vis_weight, baselines = stokes_I(ss, tel)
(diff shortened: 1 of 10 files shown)


Meilingdeng:
they all look good to me, except that I do not understand the 'pol' axis, so it will be E_theta/E_phi or E_co/E_cx as the ludwig3 definition, or can be both?

tristpinsm:
@Meilingdeng In principle it can be either. At the moment I haven't put in a place to specify the polarization axes. If we think we will be carrying around datasets with both of these conventions I can add attribute that describes which one it is. But given it is easy to convert between these two we might avoid confusion by always writing theta/phi and converting to ludwig3 when that is helpful

tristpinsm:
another alternative would be to tie these conventions to the choice of coordinates. So if we are using celestial it is theta/phi and for telescope coordinates it could be ludwig3 for example

Meilingdeng:
I see.. to me, I think stick to E_theta/E_phi and convert to Ludwig 3 when
necessary is good, so that no confusion and also flexible.

jrs65:
@tristpinsm how far off do you think this is? It would be good to get the changes merged in. Particularly the work on the regridder which is useful elsewhere.

tristpinsm:
I think this is pretty much done as far as the goal of getting holography transits regridded and packaged into beam containers go (as is https://bitbucket.org/chime/ch_pipeline/pull-requests/19/holography-transit-processing/diff). So if it looks good to you we should go ahead and merge.

jrs65:
@tristpinsm any reason to not merge this one?

tristpinsm:
No, nothing new. It should be ready to merge.

PR #17, opened 2019-03-12
Expand Products fix with stack axis

cahofer:
you have to give it stack 'None' so it won't crash in ExpandProducts


cahofer:
Hey I am adding @nritsche and @tristpinsm for this PR so that it can be merged! 
Thanks

jrs65:
It took me a while to understand exactly why you need to set `stack=None` in here, but it makes sense to me now. Thanks!

PR #18, opened 2019-03-12
noise fix

cahofer:
if we want the weight to be 1/var and var = 50**/nsamp then we have to multiple the 1/std^2 by 0.5.
I think with 2 / std^2 we are off by a factor of 4


cahofer:
Adding @tristpinsm to check the weight fix

PR #20, opened 2019-03-13
Gains restructure

cahofer:
this is a base gains class that is compatible with the temp sims coming up in ch_pipeline


jrs65:
Can you describe the motivation for these changes a bit more? It's not clear exactly why you need to introduce the extra layer of indirection. Is it because some upcoming changes don't separate into amplitude and phase fluctuations?

cahofer:
In the future I would like to have the ability to both simulate phase and amplitude errors which will both depend on temperature errors. I thought the best way to do this is to have a TempGains class that will inherit from BaseGains and in generate_gain the generate_temp_fluctuation and the query of the outside temperature function will only be called once instead of twice if we have one the phase and amp function. What do you think?

jrs65:
Okay. Thanks for the explanation. This seems fine then, but you should probably remove the empty definitions of `_generate_amp/phase` from the `BaseGains` class and only have them in the `RandomGains` one.

cahofer:
How about I move _generate_gain from Random.Gains into BaseGains (since RandomGains and Timingerrors are using both this function)? 

I would then leave _generate_amp and _generate_phase as unimplemented in BaseGains, because all child classes will have to implement these unless you set either amp or phase to False.

cahofer:
Hi @jrs65, sorry it took so long to get this PR tidied up. I was simply confused how to build the base classes in a good efficient way. 

`BaseGains` is a base class that takes in `process` a time stream file and generates gain errors from that. That's how the code existed originally with `RandomGains` (1-2 years ago) and I wanted to keep hold of simulating gain errors this way because it's convenient when we deal with gaussian gain errors that just integrate down like 1/sqrt(days). It gives the simulator a quick way to estimate lots of power spectra to get a sense of the bias and variance with a single configuration file.
This would not be more cumbersome with a single sidereal stream file. 

`SiderealGains` is for simulating any type sidereal gain errors that you have to stack up over x number of years.
I wrote this at some point but ended up never using this specific base class for any of my sims.
We can either keep it, if you think it's useful otherwise we can also trash it. 

The base class I used for all my temperature dependent simulations is in `ch_pipeline.synthesis.tempgains` on the branch `tempgain_sims`.
It is a base class that takes a weather stream file into process and has a bunch of weather specific methods implemented. I believe this class is CHIME specific enough to put it into `ch_pipeline` but the branch is still WIP.

draco/synthesis/gain.py
```diff
@@ -9,18 +7,21 @@
     :toctree:
 
     RandomGains
+    GainStacker
```
jrs65: Add `SiderealGains`

jrs65:
Hey. This is all looking great now. Thanks for all your work. Just two very short things to do before merging:

- Fix up the one comment I have.
- Rebase everything together into a single commit (or a few meaningful commits) with a commit message in the new format. If you've never done a rebase and want some advice, let me know and I'll swing over (it'll take five mins).

jrs65:
Actually. If you think it needs be just one commit, just send me your suggested message in the new format and I'll squash merge it.

Thinking about it, maybe you didn't see that presentation by Rick at the hackathon. If not, ask Tristan to explain!

cahofer:
Hey @jrs65,
proposed commit for merge:

feat(gain.py): add tasks SiderealGains and GainStacker

- Have to ability to simulate gain errors in sidereal time using the base class `SiderealGains`. The child classes will have to implement the methods `_generate_phase` and `_generate_amp`.

- Stack products of gains over sidereal days in `GainStacker` and apply them to a sidereal stream visibility.

PR #21, opened 2019-04-04
stack can only be created when prod is not None

cahofer:
For simulating timestream data, stack can only be created when prod is not None otherwise it crashes on line 687
`TypeError: object of type 'NoneType' has no len()`


jrs65:
Oh sorry. I made exactly the same fix in a branch I forgot to push. This looks good though, so I'm merging now.

PR #22, opened 2019-04-08
Fast collate products

jrs65:
Significant speed ups to collate products:

- Moved calculate redundancy into a Cython function.
- Moved slicing of dataset to be outside tightloop to remove implied MPI
calls.

### DIFF
--- a/draco/analysis/transform.py
+++ b/draco/analysis/transform.py
@@ -135,6 +135,12 @@ class CollateProducts(task.SingleTask):
                                   for upp in self.telescope.uniquepairs],
                                   dtype=[('prod', '<u4'), ('conjugate', 'u1')])
 
+        # Construct the equivalent prod and stack index_map for the telescope instance
+        dt_prod = np.dtype([('input_a', '<u2'), ('input_b', '<u2')])
+        self.bt_prod = (np.array(np.triu_indices(self.telescope.nfeed))
+                        .astype('<u2').T.copy().view(dt_prod).reshape(-1))
+
+
 
     def process(self, ss):
         """Select and reorder the products.
@@ -208,10 +214,6 @@ class CollateProducts(task.SingleTask):
 
         bt_freq = ss.index_map['freq'][freq_ind]
 
-        # Construct the equivalent prod and stack index_map for the telescope instance
-        dt_prod = np.dtype([('input_a', '<u2'), ('input_b', '<u2')])
-        bt_prod = np.array(np.triu_indices(self.telescope.nfeed)).astype('<u2').T.copy().view(dt_prod).reshape(-1)
-
         # Construct the equivalent stack reverse_map for the telescope instance.  Note
         # that we identify invalid products here using an index that is the size of the stack axis.
         feedmask = pack_product_array(self.telescope.feedmask)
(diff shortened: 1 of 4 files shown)


draco/analysis/transform.py
```diff
@@ -200,15 +209,8 @@ def pack_product_array(arr):
...
-                                     for fj in range(fi, self.telescope.nfeed)],
-                                     dtype=[('input_a', '<u2'), ('input_b', '<u2')])
-
-        bt_stack = np.array([(tools.cmap(upp[0], upp[1], self.telescope.nfeed), 0)
-                             if upp[0] <= upp[1] else
-                             (tools.cmap(upp[1], upp[0], self.telescope.nfeed), 1)
-                             for upp in self.telescope.uniquepairs],
-                             dtype=[('prod', '<u4'), ('conjugate', 'u1')])
+        dt_prod = np.dtype([('input_a', '<u2'), ('input_b', '<u2')])
+        bt_prod = np.array(np.triu_indices(self.telescope.nfeed)).astype('<u2').T.copy().view(dt_prod).reshape(-1)
```
tristpinsm: couldn't this also be precalculated in `setup`? It's probably not wasting much time, but it would be a bit more consistent.

draco/analysis/transform.py
```diff
@@ -285,20 +295,20 @@ def pack_product_array(arr):
...
             if feedconj == conj:
-                sp.vis[:, sp_pi] += wss * ss.vis[freq_ind, ss_pi]
+                spv[:, sp_pi] += wss * ssv[freq_ind, ss_pi]
             else:
-                sp.vis[:, sp_pi] += wss * ss.vis[freq_ind, ss_pi].conj()
+                spv[:, sp_pi] += wss * ssv[freq_ind, ss_pi].conj()
 
             # Accumulate variances in quadrature.  Save in the weight dataset.
-            sp.weight[:, sp_pi] += wss**2 * tools.invert_no_zero(ss.weight[freq_ind, ss_pi])
+            spw[:, sp_pi] += wss**2 * tools.invert_no_zero(ssw[freq_ind, ss_pi])
```
tristpinsm: maybe I'm just missing something but I don't see this accumulating variance. At least in the inverse variance case, doesn't this end up accumulating the weight (non-inverted)?

draco/util/_fast_tools.pyx
```diff
@@ -80,3 +82,68 @@ def _unpack_product_array_fast(cython.numeric[::1] utv, cython.numeric[:, ::1] m
...
+    rshape = (redundancy.shape[0], redundancy.shape[1])
+    if rshape != (nstack, ntime):
+        raise RuntimeError("redundancy array shape %s incorrect, expected %s" %
+                           (repr(rshape), repr((nstack, ntime))))
+        
+    # Check that we don't index out of bounds from the prod_map
+    if np.min(prod_map) < 0 or np.max(prod_map) >= ninput:
+        raise RuntimeError("Input index in prod_map out of bounds.")
+
+    with nogil:
```
tristpinsm: since there is no parallelism, why do you need to release the GIL?

draco/analysis/transform.py
jrs65: Good point. I can do that I guess.

draco/analysis/transform.py
jrs65: It inverts the original weights to get the variances and then accumulates them. Maybe I'm not understanding your comment.

draco/util/_fast_tools.pyx
jrs65: It was once a parallel loop, but that didn't end up saving much time. In fact many times it even hurt performance. However, if I can release the GIL I might as well, so I left it in.

draco/analysis/transform.py
tristpinsm: oh I see, they are accumulated according to the weighted scheme... I was indeed missing something. It's just strange that in the inverse variance case you end up adding up the original weights

draco/util/_fast_tools.pyx
tristpinsm: ok, I guess it's not hurting anything

PR #23, opened 2019-04-11
Improve regridding speed

jrs65:
Significantly improves the speed of the regridder by restricting the
length of the inner summation when calculating the band matrix to only
the non-zero elements.


tristpinsm:
I'm going to go ahead and merge so I can get this into my beam processing branch.

PR #25, opened 2019-04-30
Make MModeTransform slice mmodes at tel_mmax if requested

cahofer:
I would like to have the capability to slice the mmodes at the telescope mmax if requested. 
(when we have more time samples in the visibilities than 2*mmax + 1)

### DIFF
--- a/draco/analysis/transform.py
+++ b/draco/analysis/transform.py
@@ -434,7 +434,14 @@ class MModeTransform(task.SingleTask):
     Currently ignores any noise weighting.
     """
     def setup(self, manager=None):
+        """Set the telescope instance if a manager object is given.
 
+        Parameters
+        ----------
+        manager : manager.ProductManager (optional)
+            If the ProductManager is not None it returns an instance of a
+            drift.core.TransitTelescope object.
+        """
         if manager is not None:
             self.telescope = io.get_telescope(manager)
         else:
@@ -460,12 +467,12 @@ class MModeTransform(task.SingleTask):
         weight_sum = sstream.weight[:].sum(axis=-1)
 
         if self.telescope is not None:
-            setmmax = self.telescope.mmax
+            mmax = self.telescope.mmax
         else:
-            setmmax = None
+            mmax = None
 
         # Construct the array of m-modes
-        marray = _make_marray(sstream.vis[:], setmmax)
+        marray = _make_marray(sstream.vis[:], mmax)
         marray = mpiarray.MPIArray.wrap(marray[:], axis=2, comm=sstream.comm)
 
         # Create the container to store the modes in
--- a/draco/analysis/transform.py
+++ b/draco/analysis/transform.py
@@ -432,7 +432,18 @@ class MModeTransform(task.SingleTask):
     """Transform a sidereal stream to m-modes.
 
     Currently ignores any noise weighting.
+
+    Attributes
+    ----------
+    set_telmmax : bool
+        Whether to slice the data at the maximum m the telescope can resolve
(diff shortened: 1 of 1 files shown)


draco/analysis/transform.py
```diff
@@ -432,7 +432,18 @@ class MModeTransform(task.SingleTask):
...
+
+    Attributes
+    ----------
+    set_telmmax : bool
+        Whether to slice the data at the maximum m the telescope can resolve
+        after fourier transforming. Default : false.
     """
+    set_telmmax = config.Property(proptype=bool, default=False)
+
+    def setup(self, manager):
```
jrs65: This introduces a mandatory dependency on the `manager` but that's actually unnecessary if you don't set `set_telmmax`.

I would change this to make manager optional `i.e. manager=None`, and then if `manager` is not None this is equivalent to setting `set_telmmax`. Then I'd remove the `set_telmmax` option entirely.

draco/analysis/transform.py
```diff
@@ -453,8 +464,13 @@ def process(self, sstream):
         # variance for the m-modes
         weight_sum = sstream.weight[:].sum(axis=-1)
 
+        if self.set_telmmax is True:
```
jrs65: With that I said above this clause would just change to `self.telescope is not None`

draco/analysis/transform.py
cahofer: Ok that makes sense. I made the changes and just going to quickly test the code.

cahofer:
Let me know if the changes I implemented were the ones you had in mind. Now the manager is optional.

draco/analysis/transform.py
```diff
@@ -433,6 +433,12 @@ class MModeTransform(task.SingleTask):
 
     Currently ignores any noise weighting.
     """
+    def setup(self, manager=None):
```
jrs65: The docstring for this function should explain the behaviour here.

draco/analysis/transform.py
```diff
@@ -453,8 +459,13 @@ def process(self, sstream):
         # variance for the m-modes
         weight_sum = sstream.weight[:].sum(axis=-1)
 
+        if self.telescope is not None:
+            setmmax = self.telescope.mmax
```
jrs65:
```suggestion
            mmax = self.telescope.mmax
```

draco/analysis/transform.py
```diff
@@ -453,8 +459,13 @@ def process(self, sstream):
...
+        else:
+            setmmax = None
```
jrs65:
```suggestion
            mmax = None
```

draco/analysis/transform.py
```diff
@@ -453,8 +459,13 @@ def process(self, sstream):
...
+
         # Construct the array of m-modes
-        marray = _make_marray(sstream.vis[:])
+        marray = _make_marray(sstream.vis[:], setmmax)
```
jrs65:
```suggestion
        marray = _make_marray(sstream.vis[:], mmax)
```

draco/analysis/transform.py
jrs65: This variable name doesn't make sense. `setxyz` sounds more like a binary parameter that sets some behaviour, whereas it actually represents the `mmax` itself.

draco/analysis/transform.py
jrs65: You might also want to explain the behaviour in the class docstring. Only one location needs a full account though.

cahofer:
Okay, I added a docstring for setup and changed the cosmetic.

PR #29, opened 2019-05-24
feat(io): Add config option to LoadFilesFromParams

fandinomat:
Allows files to be loaded without any axes distributed.
Default value means it is backwards compatible with previous API.

### DIFF
--- a/draco/core/io.py
+++ b/draco/core/io.py
@@ -180,10 +180,12 @@ class LoadFilesFromParams(task.SingleTask):
     ----------
     files : glob pattern, or list
         Can either be a glob pattern, or lists of actual files.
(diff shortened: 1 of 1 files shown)


draco/core/io.py
```diff
@@ -183,6 +183,7 @@ class LoadFilesFromParams(task.SingleTask):
     """
 
     files = config.Property(proptype=_list_or_glob)
+    distribute = config.Property(proptype=bool, default=True)
```
jrs65: Can we call the parameter `distributed` for consistency?

jrs65:
Actually I want to merge this, so I'm just going to make the change myself.

PR #33, opened 2019-06-24
Beamform dev

fandinomat:
Can process sidereal streams and time streams, with options for full polarization, tokes I or co-pol.

### DIFF
--- a/draco/analysis/beamform.py
+++ b/draco/analysis/beamform.py
@@ -1,7 +1,7 @@
 import numpy as np
 from mpi4py import MPI
 
-from caput import config, mpiutil
+from caput import config, mpiutil, mpiarray
 from cora.util import units
 
 from ..core import task, containers, io
@@ -104,18 +104,22 @@ class BeamFormBase(task.SingleTask):
             formed_beam = containers.FormedBeam(
                     freq=self.freq,
                     object_id=self.source_cat.index_map['object_id'],
-                    pol=np.array(self.return_pol))
+                    pol=np.array(self.return_pol),
+                    distributed = True)
         else:
             # Container to hold the formed beams
             formed_beam = containers.FormedBeamHA(
                     freq=self.freq, ha=np.arange(self.nha, dtype=int),
                     object_id=self.source_cat.index_map['object_id'],
-                    pol=np.array(self.return_pol))
+                    pol=np.array(self.return_pol),
+                    distributed = True)
             # Initialize container to zeros.
             formed_beam.ha[:] = 0.
         # Initialize container to zeros.
         formed_beam.beam[:] = 0.
         formed_beam.weight[:] = 0.
+        # Ensure container is distributed in frequency
+        formed_beam.redistribute('freq')
 
         # Indices of local frequency axis. Needed because in principle
         # beamform() allows a subset of frequencies to be processed.
@@ -131,10 +135,6 @@ class BeamFormBase(task.SingleTask):
             # Phasing will actually be done at src position.
             if self.is_sstream:
                 sra_index = np.searchsorted(self.ra, self.sra[src])
-                # Compute hour angle array
-                ha_array, ra_index_range = self._ha_array(
-                    self.ra, sra_index, self.sra[src],
-                    self.ha_side, self.is_sstream)
             else:
                 # Cannot use searchsorted, because RA might not be
                 # monotonically increasing. Slower.
@@ -146,26 +146,27 @@ class BeamFormBase(task.SingleTask):
                 ra_cadence = self.ra[1]-self.ra[0]
                 if transit_diff[sra_index] > 1.5 * ra_cadence:
                     continue
-                # Compute hour angle array
-                ha_array, ra_index_range, ha_mask = self._ha_array(
-                    self.ra, sra_index, self.sra[src],
-                    self.ha_side, self.is_sstream)
+
+            # Compute hour angle array
+            ha_array, ra_index_range, ha_mask = self._ha_array(
+                self.ra, sra_index, self.sra[src],
+                self.ha_side, self.is_sstream)
 
             # The length of the HA axis might not be equal to
             # self.nha for all sources in case self.is_sstream
             # is False. It could be smaller if some HA are
-            # outside the range of the data.
+            # outside the range of the data. nha is the actual size.
             nha = len(ha_array)
             if self.collapse_ha:
                 formed_beam_full = np.zeros(
-                    (self.npol, self.nfreq), dtype=np.float64)
+                    (self.npol, self.local_nfreq), dtype=np.float64)
                 weight_full = np.zeros(
-                    (self.npol, self.nfreq), dtype=np.float64)
+                    (self.npol, self.local_nfreq), dtype=np.float64)
             else:
                 formed_beam_full = np.zeros(
-                    (self.npol, self.nfreq, self.nha), dtype=np.float64)
+                    (self.npol, self.local_nfreq, self.nha), dtype=np.float64)
                 weight_full = np.zeros(
-                    (self.npol, self.nfreq, self.nha), dtype=np.float64)
+                    (self.npol, self.local_nfreq, self.nha), dtype=np.float64)
             # For each polarization
             for pol in range(self.npol):
 
@@ -195,7 +196,7 @@ class BeamFormBase(task.SingleTask):
                         np.sum(sumweight_inrange, axis=-1) * 
                         primary_beam, axis=1)
 
-                    this_formed_beam = (
+                    formed_beam_full[pol] = (
                         np.sum(this_formed_beam * primary_beam, axis=1) *
                         invert_no_zero(this_sumweight))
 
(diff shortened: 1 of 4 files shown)


draco/analysis/beamform.py
```diff
@@ -0,0 +1,515 @@
...
+        'full' : 'XX', 'YY', 'XY' and 'YX' in this order.
+        'copol' : 'XX' and 'YY' only.
+        'stokes' : 'I', 'Q', 'U' and 'V' in this order. Not implemented.
+    timetrack : float
+        How long (in seconds) to track sources at each side of transit.
+        Total transit time will be ~ 2 * timetrack.
+    """
+
+    collapse_ha = config.Property(proptype=bool, default=True)
+    polarization = config.Property(proptype=str, default='full')
```
jrs65: Use the `config.enum` type here, it will validate for you automatically.

draco/analysis/beamform.py
```diff
@@ -0,0 +1,515 @@
...
+
+    Attributes
+    ----------
+    collapse_ha : bool
+        Wether or not to sum over hour-angle/time to complete
+        the beamforming. Default is True, which sums over.
+    polarization : string
+        One of:
+        'I' : Stokes I only.
+        'full' : 'XX', 'YY', 'XY' and 'YX' in this order.
```
jrs65: The polarisation order should be `XX`, `XY`, `YX`, `YY` to be consistent with everywhere else that we use this (e.g. ring maps)

draco/analysis/beamform.py
```diff
@@ -0,0 +1,515 @@
...
+from ..util.tools import polarization_map, baseline_vector
+from mpi4py import MPI
+
+# Constants
+NU21 = units.nu21
+C = units.c
+
+
+class BeamFormBase(task.SingleTask):
+    """ Base class for beam forming tasks.
```
jrs65: Please format docstrings as:
```
class Name(Blah):
    """Short single line summary.

    Longer description of what what it is. Optional.

    Parameters
    -----------
    ...
```

draco/analysis/beamform.py
```diff
@@ -0,0 +1,515 @@
...
+            self.return_pol = ['I', 'Q', 'U', 'V']
+            msg = 'Stokes parameters are not implemented'
+            raise RuntimeError(msg)
+        else:
+            msg = 'Invalid polarization parameter: {0}'
+            msg = msg.format(self.polarization)
+            raise ValueError(msg)
+        # Number of polarizations to process
+        self.npol = len(self.process_pol)
+        self.latitude = np.deg2rad(ephemeris.CHIMELATITUDE)
```
jrs65: This is in `draco` it can't depend on CHIME. You can get the latitude from the `telescope` instance.

draco/analysis/beamform.py
```diff
@@ -0,0 +1,515 @@
+import numpy as np
+from caput import mpiarray, config, mpiutil, pipeline
+from ..core import task, containers, io
+from cora.util import units
+from ch_util import tools, ephemeris
```
jrs65: You can't depend on `ch_util` in `draco`.

draco/analysis/beamform.py
```diff
@@ -0,0 +1,515 @@
+import numpy as np
```
jrs65: Imports should be ordered as:
```
import python_standard_libs

import third_party_packages_eg
import numpy as np

from my_other_package import other_module

from . import my_local_module
```

draco/analysis/beamform.py
```diff
@@ -0,0 +1,515 @@
...
+                    pol=np.array(self.return_pol))
+        else:
+            # Container to hold the formed beams
+            formed_beam = containers.FormedBeamHA(
+                    freq=self.freq, ha=np.arange(self.nha, dtype=int),
+                    object_id=self.source_cat.index_map['object_id'],
+                    pol=np.array(self.return_pol))
+            # Initialize container to zeros.
+            formed_beam.fbeam[:] = 0.
+            formed_beam.ha[:] = 0.  # TODO: Should this be np.nan instead?
```
jrs65: I'd imagine it should be zero. You should typically never introduce a NaN yourself. What's the argument it should be NaN?

If you're worried about how to take care of missing data, you should probably have a weight dataset.

draco/analysis/beamform.py
```diff
@@ -0,0 +1,515 @@
...
+        else:
+            # XY or YX
+            return (_amp(0, dec, zenith)*np.exp(
+                                -((ha-ha0)/_sig(0, freq, dec))**2) *
+                    _amp(1, dec, zenith)*np.exp(
+                                -((ha-ha0)/_sig(1, freq, dec))**2))**0.5
+
+    def _process_data(self, data):
+        """ Store code for parsing and formating data
+            priot to beamforming.
```
jrs65:
```suggestion
            prior to beamforming.
```

draco/analysis/beamform.py
```diff
@@ -0,0 +1,515 @@
...
+
+        source_cat : :class:`containers.SourceCatalog`
+            Catalog of points to beamform at.
+
+        """
+        super(BeamForm, self).setup(manager)
+
+        # Extract source catalog information
+        self.source_cat = source_cat
+        # Number of poinytings
```
jrs65:
```suggestion
        # Number of pointings
```

draco/core/containers.py
```diff
@@ -1315,6 +1315,65 @@ class SpectroscopicCatalog(SourceCatalog):
...
 
 
+class FormedBeam(ContainerBase):
+    """Container for formed beams.
+    """
+    
+    _axes = ('object_id', 'pol', 'freq')
+
+    _dataset_spec = {
+        'fbeam': {
```
jrs65: Any reason to not call it `beam`?

draco/core/containers.py
```diff
@@ -1315,6 +1315,65 @@ class SpectroscopicCatalog(SourceCatalog):
...
+            'axes': ['object_id', 'pol', 'freq'],
+            'dtype': np.float64,
+            'initialise': True,
+            'distributed': False
+        }
+    }
+
```
jrs65: You might consider adding a weight dataset in here.

draco/analysis/beamform.py
fandinomat: I am not sure what the best solution is here.

I can move `polarization_map()` to ch_pipeline.util.tools.py (which I would create).
But then I would depend on ch_pipeline, is that ok?

draco/core/containers.py
fandinomat: Not a problem here. I'll make the change.
(In BeamForm I use the word `beam` to denote the primary beam returned by my Gaussian functions. So I use formed_beam to differentiate.)

draco/analysis/beamform.py
jrs65: No. The point is draco can't depend on private CHIME packages.

I think you can mostly decode this by looking at the telescope `_beamclass` attribute. I think it's 0 for X pol and 1 for Y pol.

fandinomat:
@jrs65 thanks for the comments! I have fixed all of them.

One comment: When adding XX and YY to get Stokes I, I don't do inverse weighting. Just add them stright up. I think that's the correct thing to do but not sure.

fandinomat:
I have now added all the features I wanted to the beam-former.

draco/util/_fast_tools.pyx
```diff
@@ -152,3 +154,64 @@ def _calc_redundancy(float[:, ::1] input_flags, int16_t[:, ::1] prod_map, uint32
                 # Increment the redundancy counter for this unique baseline if both inputs good
                 redundancy[istack, jj] += input_flags[ia, jj] * input_flags[ib, jj]
 
+
+cdef extern from "complex.h" nogil:
+    double complex cexp(double complex)
+
+@cython.wraparound(False)
+@cython.boundscheck(False)
+cpdef beamform(float complex [:, :, ::1] vis,
```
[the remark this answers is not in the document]
fandinomat: Done.

draco/core/containers.py
```diff
@@ -1315,6 +1315,100 @@ class SpectroscopicCatalog(SourceCatalog):
...
+            'dtype': np.float64,
+            'initialise': True,
+            'distributed': False
+        }
+
+    }
+
+    @property
+    def ha(self):
+        return self['hour_angle']
```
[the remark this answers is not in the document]
fandinomat: Done.

draco/analysis/beamform.py
```diff
@@ -0,0 +1,651 @@
...
+            ra_index_range = ra_index_range[ha_mask]
+            # Hour angle array (convert to radians)
+            ha_array = np.deg2rad(ra[ra_index_range] - source_ra)
+            # For later convenience it is better if `ha_array` is
+            # in the range -pi to pi instead of 0 to 2pi.
+            ha_array = (ha_array + np.pi) % (2.*np.pi) - np.pi
+
+        return ha_array, ra_index_range, ha_mask
+
+    def _beamfunc(self, ha, pol, freq, dec, zenith=0.70999994):
```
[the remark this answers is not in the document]
fandinomat: Ok. Added TODO.

PR #34, opened 2019-07-26
Add option to flagg intracilinder visibilities.

fandinomat:
Also allow ApplyGain to opperate with containers.SiderealGainData.


jrs65:
I think this already exists in ch_pipeline (https://bitbucket.org/chime/ch_pipeline/src/fe130073d6d784384a0481c683b1e31bdb35a1f5/ch_pipeline/analysis/flagging.py#lines-1368), but I guess it's good to have a generic version of it. To that end you might want to change it to be a `mask_short_ew` and have a threshold for the EW baseline length.

fandinomat:
Thanks Richard, this is a good idea. I implemented it.

PR #35, opened 2019-08-16
Python 3 support

jrs65:
This should be everything required to support Python 3.

### DIFF
--- a/draco/analysis/transform.py
+++ b/draco/analysis/transform.py
@@ -226,7 +226,7 @@ class CollateProducts(task.SingleTask):
         feedmask = pack_product_array(self.telescope.feedmask)
         bt_rev = np.fromiter(zip(
                  np.where(feedmask, pack_product_array(self.telescope.feedmap), self.telescope.npairs),
-                 np.where(feedmask, pack_product_array(self.telescope.feedconj), 0))),
+                 np.where(feedmask, pack_product_array(self.telescope.feedconj), 0)),
                  dtype=[('stack', '<u4'), ('conjugate', 'u1')])
 
         # Create output container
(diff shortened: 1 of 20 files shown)


draco/analysis/transform.py
```diff
@@ -217,9 +224,9 @@ def pack_product_array(arr):
         # Construct the equivalent stack reverse_map for the telescope instance.  Note
         # that we identify invalid products here using an index that is the size of the stack axis.
         feedmask = pack_product_array(self.telescope.feedmask)
-        bt_rev = np.array(zip(
+        bt_rev = np.array(list(zip(
```
ketiltrout: Instead of an ephemeral listification, just use the iterator directly:
```suggestion
        bt_rev = np.fromiter(zip(
```

jrs65: Good catch. This is one that my syntax tree heuristics did not recognise!

draco/analysis/transform.py
```diff
@@ -217,9 +224,9 @@ def pack_product_array(arr):
         # Construct the equivalent stack reverse_map for the telescope instance.  Note
         # that we identify invalid products here using an index that is the size of the stack axis.
         feedmask = pack_product_array(self.telescope.feedmask)
-        bt_rev = np.array(zip(
+        bt_rev = np.fromiter(zip(
                  np.where(feedmask, pack_product_array(self.telescope.feedmap), self.telescope.npairs),
-                 np.where(feedmask, pack_product_array(self.telescope.feedconj), 0)),
+                 np.where(feedmask, pack_product_array(self.telescope.feedconj), 0))),
```
ketiltrout: You need to also revert the added `)` here:
```suggestion
                 np.where(feedmask, pack_product_array(self.telescope.feedconj), 0)),
```

PR #37, opened 2019-08-23
Qstack dev

fandinomat:
This is the much simpler version of the quasar stacker that uses the product of the beamformer ( #33 ).

### DIFF
--- a/draco/analysis/qstack.py
+++ b/draco/analysis/qstack.py
@@ -118,27 +118,23 @@ class QuasarStack(task.SingleTask):
         for lvi, gvi in bvec_m.enumerate(axis=0):
 
             gpi = data.index_map['stack'][gvi][0]  # Global product index
-            conj = data.index_map['stack'][gvi][1]  # Product conjugation
             # Inputs that go into this product
             ipt0 = data.index_map['input']['chan_id'][
                                 data.index_map['prod'][gpi][0]]
             ipt1 = data.index_map['input']['chan_id'][
                                 data.index_map['prod'][gpi][1]]
 
-            # Get position and polarization of each input
-            pos0, pol0 = self._pos_pol(ipt0)
-            pos1, pol1 = self._pos_pol(ipt1)
-
-            is_copol = (((pol0 == 0) and (pol1 == 0)) or
-                       ((pol0 == 1) and (pol1 == 1)))
-            if is_copol:
+            if self._is_array_copol(ipt0, ipt1):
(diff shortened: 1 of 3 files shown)


draco/analysis/qstack.py
```diff
@@ -0,0 +1,148 @@
...
+        # Gather quasar stack for all ranks. Each contains the sum
+        # over a different subset of quasars.
+        quasar_stack_full = np.zeros(
+            mpiutil.size * self.nstack, dtype=quasar_stack.dtype
+        )
+        quasar_weight_full = np.zeros(
+            mpiutil.size * self.nstack, dtype=quasar_weight.dtype
+        )
+        # Gather all ranks
+        mpiutil.world.Allgather(quasar_stack, quasar_stack_full)
```
jrs65: Try to avoid using `mpiutil` when you can. In this case you can get the communicator out of the input data and just use `comm.Allgather`

jrs65: You may want to use `allgather`, which will involve creating the destination array, but will be faster and scale to large numbers of quasars.

draco/core/containers.py
```diff
@@ -1335,6 +1335,40 @@ def spectrum(self):
         return self.datasets["spectrum"]
 
 
+class FrequencyStack(ContainerBase):
```
jrs65: Can you expand this documentation more? It's a bit generic.

draco/analysis/qstack.py
```diff
@@ -0,0 +1,148 @@
...
+from ..util._fast_tools import beamform
+from ..core import task, containers, io
+
+# Constants
+NU21 = units.nu21
+C = units.c
+
+
+class QuasarStack(task.SingleTask):
+    """
```
jrs65: Add a docstring. Also, maybe call it `SourceStack` as we'll surely work with other types of sources.

draco/analysis/qstack.py
fandinomat: Hmm. I am not sure what you mean. I do create the destination arrays (`quasar_stack_full`) and I thought the upper-case versions were the buffer-object, fast ones.
Also, keep in mind that this gathering is done from an object of size `nstack` (number of frequencies in the quasar stack, so around 100) to an object of size `(nprocess * nstack)`, so it's very small...

fandinomat:
@jrs65 I expanded the documentation a bit and got rid of the mpiutil calls.
Also merged the most recent developments of master and ran black.

PR #41, opened 2019-08-30
feat(delay): Add baseline dependent delay cut

fandinomat:
BREAKING CHANGE: DelayFilter now has a `setup` method and `requires` a telescope object.
Also fix normalization bug.

### DIFF
--- a/draco/analysis/delay.py
+++ b/draco/analysis/delay.py
@@ -21,13 +21,6 @@ from ..core import containers, task, io
 class DelayFilter(task.SingleTask):
     """Remove delays less than a given threshold.
 
-    Note: The baseline-dependent delay threshold take in
-    consideration only the Y (NS) component of the baseline.
-    This is the correct thing to do for a cylindrical transit
-    telescope oriented in the NS direction. For dish type
-    reflectors, this might need to be changed to use the full
-    baseline length.
-
     Attributes
     ----------
     delay_cut : float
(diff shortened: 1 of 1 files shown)


fandinomat:
> Looks great! So the baseline cut only uses the NS baseline length? I think that's probably the right thing for us, but maybe not so generic. Possibly add a note about that somewhere if you can. Otherwise looks great.

Yes. I figured this was the contaminated region for us, given that we are only really sensitive in the NS stripe of the beam. Using the baseline full length would cut out regions that look `clean` in the delay spectrum. I'll add a note.

fandinomat:
I can add a parameter to pick between using NS, EW or full baseline length.
Something like:
`telescope_orientation = config.enum(['NS', 'EW', 'none'], default='NS')`

jrs65:
Yeah, that sounds like a great option. Thanks!

fandinomat:
Done.

draco/analysis/delay.py
```diff
@@ -54,10 +83,31 @@ def process(self, ss):
...
-            freq_weight = np.median(ssw[:, lbi], axis=1)
-            NF = null_delay_filter(freq, self.delay_cut, freq_weight)
+            # Select the baseline length to use
+            baseline = ubase[uinv[bi]]
+            if self.telescope_orientation == 'NS':
+                baseline = abs(baseline[1])  # Y baseline
+            elif self.telescope_orientation == 'EW':
+                baseline = abs(baseline[0])  # X baseline
+            else:
+                baseline = np.linalg.norm(baseline)  # Norm
```
jrs65: I think this collapses both axes and thus is flat with baselines. Try `np.hypot` instead, or specify the `axis` argument.

fandinomat: I am inside of a baseline loop here, so I think I selected a single baseline entry right before the `if` statement. The only axis remaining is the xy one.
I couldn't test though because I can't get any jobs running on cedar.

fandinomat: Testing now.

jrs65: Oh. Good point. My bad!

PR #42, opened 2019-08-31
feat(task): add task that deletes input to free memory.

ssiegelx:


jrs65:
Seems like a good idea to me. Does it work? I always struggle to follow how the pipeline manager hangs on to references to the input and output of each task.

jrs65:
Okay. My concern was that maybe the pipeline manager hangs on to references longer than the lifetime of the task in the product queue, but it looks like it doesn't do that, so I guess it works fine.

PR #48, opened 2019-09-18
Chunkify

tristpinsm:


tristpinsm:
One thing I just noticed : In `Truncate`, where it uses the weights to determine the precision, at the moment it makes a copy of the weights array to scale by the fractional variance limit. We could avoid this copy by adding the limit factor as an argument to the `bit_truncate_weight` function. Is that worth adding to this PR?

jrs65:
> One thing I just noticed : In `Truncate`, where it uses the weights to determine the precision, at the moment it makes a copy of the weights array to scale by the fractional variance limit. We could avoid this copy by adding the limit factor as an argument to the `bit_truncate_weight` function. Is that worth adding to this PR?

That might be worth adding if you are making other changes. But I wouldn't bother otherwise.

jrs65:
@tristpinsm I'm wondering what the status of this branch is. I guess maybe we should wait until @nritsche has finished his new zarr compression support?

nritsche:
> @tristpinsm I'm wondering what the status of this branch is. I guess maybe we should wait until @nritsche has finished his new zarr compression support?

If not specified, zarr will guess the chunk sizes, so I'm not sure if we still need this.
(https://zarr.readthedocs.io/en/stable/tutorial.html#chunk-optimizations)

jrs65:
I think we may always want to set the chunk size. Setting it appropriately is key to achieving a good compression ratio.

tristpinsm:
I haven't touched this since then. I guess it would need to be integrated with the changes to the truncation routines now.

draco/core/containers.py
```diff
@@ -265,6 +265,32 @@ def add_dataset(self, name):
...
+
+        For every dataset, check if chunks and compression are set, and
+        if not set them to dataset_spec values.
+        """
+        for dset in self.dataset_spec:
+            if "chunks" in self.dataset_spec[dset] and self[dset].chunks is None:
+                # ensure chunks aren't larger than dataset shape
+                chunks = ()
+                for i, l in enumerate(self[dset].shape):
+                    chunks += (min(self.dataset_spec[dset]["chunks"][i], l),)
```
nritsche, line 279: What is this doing?
Why would this be below 1 or why would you choose chunk size 1?

tristpinsm: did you misread `l` as `1`?

nritsche: Thanks

nritsche:
I'm re-using this in #130

tristpinsm:
> I'm re-using this in #130

hooray, Chunkify lives!

jrs65:
@nritsche shall we close this branch and wrap it up into yours?

jrs65:
> @nritsche shall we close this branch and wrap it up into yours?

Ping!

nritsche:
> @nritsche shall we close this branch and wrap it up into yours?

I was planning to close this once mine is merged, but feel free to do it now.

jrs65:
This is subsumed into Rick's zarr branch that needs to be merged.

PR #49, opened 2019-09-23
fix(transform): use `time` property instead of `index_map['time']`

ssiegelx:
`CollateProducts` now creates the time axis of the output `TimeStream`
container from the `time` property of the input container, so that
the time axis is defined with respect to the integration centre
instead of the integration start.

### DIFF
--- a/draco/analysis/transform.py
+++ b/draco/analysis/transform.py
@@ -258,8 +258,10 @@ class CollateProducts(task.SingleTask):
         # Create output container
         if isinstance(ss, containers.SiderealStream):
             OutputContainer = containers.SiderealStream
+            output_kwargs = {"ra": ss.ra[:]}
         else:
             OutputContainer = containers.TimeStream
+            output_kwargs = {"time": ss.time[:]}
 
         sp = OutputContainer(
             freq=bt_freq,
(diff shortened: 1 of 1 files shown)


ssiegelx:
The current pipeline switches from a `CorrData` container to a `TimeStream` container at the `CollateProducts` task.  The time axis for the `TimeStream` container is taken from `CorrData.index_map['time']`.  However, unlike the `CorrData` container, the `TimeStream` container does not shift the time axis to convert from integration start to integration centre.  As a result, the RA axis on all of the pipeline products are referenced with respect to the integration start time.  This results in incorrect phases for fringestopping and was responsible for the large residuals in the point source subtraction.

This pull request creates the time axis from `CorrData.time` (and thus converts to integration centre within `CollateProducts`).

Another option would be to modify `TODContainer` so that the `time` property shifts `index_map['time]` if it has a `'ctime'` field (line 475 of `draco.core.containers`).  This would yield consistent behavior between `CorrData` and `TimeStream`.  Perhaps we should make both of these changes?

draco/analysis/transform.py
```diff
@@ -258,8 +258,10 @@ def pack_product_array(arr):
         # Create output container
         if isinstance(ss, containers.SiderealStream):
             OutputContainer = containers.SiderealStream
+            output_kwargs = {"ra": ss.ra[:]}
```
jrs65: This one presumably isn't strictly necessary? I think the RA axis is always pixel centered.

jrs65: Probably safe to leave it in there though.

PR #50, opened 2019-09-24
fix(transform): try to redefine stack axis using only unmasked products

ssiegelx:
If the timestream that is input to CollateProducts is stacked, then
attempt to redefine its stack axis so that it only contains products
consisting of inputs that are not masked in the telescope instance.

Also moves determination of `bt_rev` from `process` to `setup` since
it only depends on the telescope instance.

### DIFF
--- a/draco/analysis/transform.py
+++ b/draco/analysis/transform.py
@@ -158,6 +158,36 @@ class CollateProducts(task.SingleTask):
             .reshape(-1)
         )
 
+        # Construct the equivalent reverse_map stack for the telescope instance.
+        # Note that we identify invalid products here using an index that is the
+        # size of the stack axis.
+        def pack_product_array(arr):
+
+            nfeed = arr.shape[0]
+            nprod = (nfeed * (nfeed + 1)) // 2
+
+            ret = np.zeros(nprod, dtype=arr.dtype)
+            iout = 0
+
+            for i in range(nfeed):
+                ret[iout : (iout + nfeed - i)] = arr[i, i:]
+                iout += nfeed - i
+
+            return ret
+
+        feedmask = pack_product_array(self.telescope.feedmask)
+        self.bt_rev = np.fromiter(
+            zip(
+                np.where(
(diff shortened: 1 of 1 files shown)


ssiegelx:
There are about 10 unique baselines that we lose when we `CollateProducts` because the stack axis happens to reference inputs that are not connected to CHIME antennas.  This pull request will recover those baselines.

jrs65:
Hmmmm. This sounds very familiar. I think I've actually fixed this bug in *kotekan*, but not in the extra stage of stacking here.

ssiegelx:
Recent acquisitions still suffer from this problem.  The `20191012T005524Z_chimestack_corr` acquisition has 128 stacked baselines that reference a correlator input that is not connected to a CHIME antenna (e.g., noise source, 26m, or just disconnected).

draco/analysis/transform.py
```diff
@@ -158,6 +158,36 @@ def setup(self, tel):
...
+            iout = 0
+
+            for i in range(nfeed):
+                ret[iout : (iout + nfeed - i)] = arr[i, i:]
+                iout += nfeed - i
+
+            return ret
+
+        feedmask = pack_product_array(self.telescope.feedmask)
+        self.bt_rev = np.fromiter(
```
jrs65: I'm not sure about the efficiency of rebuilding from a zipped iterator here. You could reverse the order, and build the combined feedmap, feedconj array, and then do the where on top of that.

draco/analysis/transform.py
```diff
@@ -158,6 +158,36 @@ def setup(self, tel):
             .reshape(-1)
         )
 
+        # Construct the equivalent reverse_map stack for the telescope instance.
+        # Note that we identify invalid products here using an index that is the
+        # size of the stack axis.
+        def pack_product_array(arr):
```
jrs65: How about just doing `arr[np.triu_indices(arr.shape[0])]`?

PR #54, opened 2019-10-21
RFI dev

fandinomat:
Implementation of an RFI mask derived from Sensitivity data.

### DIFF
--- a/draco/analysis/flagging.py
+++ b/draco/analysis/flagging.py
@@ -16,6 +16,7 @@ Tasks
     SmoothVisWeight
     ThresholdVisWeight
     RFIMask
+    RFISensitivityMask
 """
 # === Start Python 2/3 compatibility
 from __future__ import absolute_import, division, print_function, unicode_literals
@@ -25,12 +26,13 @@ from future.builtins.disabled import *  # noqa  pylint: disable=W0401, W0614
 # === End Python 2/3 compatibility
 
 import numpy as np
-from scipy.ndimage import median_filter, uniform_filter
+from scipy.ndimage import median_filter
 
-from caput import config, weighted_median
+from caput import config, weighted_median, mpiarray
 
 from ..core import task, containers, io
 from ..util import tools
+from ..util import rfi
 
 
 class DayMask(task.SingleTask):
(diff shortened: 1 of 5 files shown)


fandinomat:
This branch is not ready to merge. Pull request created for discussion.
On my side, the RFI mask task is ready. It might require small changes in case we decide to change the naming of the Sensitivity container attributes.

fandinomat:
TODO: The code to avoid source transits is not extensively tested because I don't have a reliable time axis on the only sensitivity data I have.

Once the sensitivity data generating task is created it should be used to create more datasets with proper time axes and the RFI masking task should be tested against those.

draco/analysis/flagging.py
```diff
@@ -402,6 +406,205 @@ def process(self, timestream):
...
+            finalmask = stmask
+            finalmask[:, madtimes] = madmask[:, madtimes]
+
+        # Apply scale invariant rank (SIR) operator, if asked for.
+        if self.sir:
+            finalmask = self.add_sir(finalmask, static_flag)
+
+        # Create container to hold mask
+        rfimask = containers.RFIMask(axes_from=sensitivity)
+        rfimask.mask[:] = finalmask
```
anjakefala: Can properties be modified like this if you lack a `setter`?

fandinomat: I am pretty sure they can. I think the way containers are defined you can directly assign to datasets in the root of the hdf5 file. Datasets inside groups are read only.
The slice operator `[:]` returns a view of the dataset array.

anjakefala: Oh totally! This is what I meant: https://stackoverflow.com/questions/36029389/python-modifying-property-value

When you define a `@property`, I think it needs a corresponding `@setter` to be modifiable. But it is true that I am unsure how this plays out with NumPy views.

fandinomat: I think you are right. I think the array view gets around that limitation because we don't assign to the property directly, but to a view of the array contained in the dataset?

fandinomat: I think if you try rfimask.mask = some_array it would fail.

anjakefala: Understood!

draco/analysis/flagging.py
```diff
@@ -28,9 +29,12 @@
 from scipy.ndimage import median_filter, uniform_filter
 
 from caput import config, weighted_median
+from ch_util import rfi, ephemeris
```
jrs65: Remember, you can't depend on `ch_util` in here. For the ephemeris calls, you can probably get away with converting times to LSA's (using an `Observer` object), and then just masking based on that and the source RA's.

For the calls into `ch_util.rfi`, I think you can just copy/move the SIR implementation into `draco.util.rfi` and probably the same with the frequency mask. @ssiegelx might have some ideas in here.

draco/analysis/flagging.py
```diff
@@ -28,9 +29,12 @@
...
+from ch_util.tools import invert_no_zero
```
jrs65: I'm pretty sure `invert_no_zero` is available from `draco.util.tools`

draco/analysis/flagging.py
```diff
@@ -402,6 +406,205 @@ def process(self, timestream):
...
+            sidereal_day = ephemeris.SIDEREAL_S * 24 * 3600
+            timestamp = sensitivity.time
+            ntm = len(timestamp)
+            start_csd = ephemeris.unix_to_csd(timestamp[0])
+            start_ra = (start_csd - int(start_csd)) * (2.0 * np.pi)
+            ra_axis = (start_ra + (2.0 * np.pi) / ntm * np.arange(ntm)) % (2.0 * np.pi)
+            # Select Sun transit times
+            suntt = ephemeris.solar_transit(timestamp[0])  # Sun transit time
+            sunidx = np.argmin(abs(timestamp - suntt))
+            beam_window = (
```
jrs65: These params should maybe be exposed as config.

draco/core/containers.py
```diff
@@ -629,6 +629,87 @@ def conjugate(self):
...
+class SystemSensitivity(TODContainer):
+    """A container for holding a sensitivity dataset in time.
+
+    This should be averaged/collapsed in the stack/prod axis
+    to provide an overall summary of the system sensitivity.
+    Two datasets are available: the measured noise and the
+    estimated noise from the visibility weights.
+    """
+
+    _axes = ('freq', 'pol', 'time')
```
jrs65: You can remove the `time` entry from `_axes`. It's inherited (by some magic) from `TODContainer`

draco/core/containers.py
```diff
@@ -629,6 +629,87 @@ def conjugate(self):
...
+    def pol(self):
+        return self.index_map['pol']
+
+
+class RFIMask(TODContainer):
+    """A container for holding RFI mask.
+
+    """
+
+    _axes = ('freq', 'time')
```
jrs65: Same here.

draco/util/rfi.py
```diff
@@ -0,0 +1,74 @@
...
+        for ii in range(flag_temp.shape[0]):
+            flag[ii:ii+m, :] += flag_temp[ii, :][np.newaxis, :]
+
+        m *= 2
+
+    return flag
+
+
+# This routine might be substituted by a faster one later
+sumthreshold = sumthreshold_py
```
jrs65: Did @nritsche finish his faster C++ version?

draco/analysis/flagging.py
```diff
@@ -402,6 +406,205 @@ def process(self, timestream):
...
+                2.0 * np.pi
+            ) / 90.0  # 4 deg in radians for the Sun. Each side.
+            # Mask the sun for the maximum aparent dec of ~24 deg
+            ra_window = beam_window / np.cos(np.deg2rad(24))
+            time_window = ra_window / (2 * np.pi) * sidereal_day
+            # Time spans where we apply the MAD filter.
+            madtimes = abs(timestamp - suntt) < time_window
+            # Select bright source transit times. Only CasA, CygA and TauA.
+            sources = [ephemeris.CasA, ephemeris.CygA, ephemeris.TauA]
+            # , ephemeris.VirA]
```
jrs65: Remove commented out line.

draco/analysis/flagging.py
```diff
@@ -402,6 +406,205 @@ def process(self, timestream):
...
+            beam_window = (
+                2.0 * np.pi
+            ) / 90.0  # 4 deg in radians for the Sun. Each side.
+            # Mask the sun for the maximum aparent dec of ~24 deg
+            ra_window = beam_window / np.cos(np.deg2rad(24))
+            time_window = ra_window / (2 * np.pi) * sidereal_day
+            # Time spans where we apply the MAD filter.
+            madtimes = abs(timestamp - suntt) < time_window
+            # Select bright source transit times. Only CasA, CygA and TauA.
+            sources = [ephemeris.CasA, ephemeris.CygA, ephemeris.TauA]
```
jrs65: Ideally we could provide this list as a config parameter, but at the moment we don't have any similar functionality in `draco`. For the moment you might just need to replace this with a list of RA and DECs.

draco/analysis/flagging.py
```diff
@@ -402,6 +406,205 @@ def process(self, timestream):
...
+        return flagsir
+
+    def mad_tv_mask(
+        self,
+        data,
+        start_flag,
+        freq,
+        sigma=5.0,
+        tv_fraction=0.5,
+        base_size=(11, 3),
```
jrs65: These should probably get exposed as parameters.

jrs65:
@fandinomat I think this is now finished, it took a bit of work to get it meshed with @saurabh-astro's code, and get the parallelisation to work. I've pulled some of the CHIME specific functionality into some hooks that will get overloaded in `ch_pipeline`.

PR #55, opened 2019-10-22
feat(gaussiannoise): add a gaussian noise dataset generator

### LINKED ISSUE #52: Generate a Gaussian noise dataset
This task should:
- [ ] Use the noise estimates of an existing dataset to replace the data section (use the `weight` property).
- [ ] Support multiple container types. I'm not sure duck typing will work, because the data section has different names, and sometimes is real and sometimes complex, so you might need to specify the properties for different types of data
- [ ] Support both draco and CHIME timestream types without explicitly importing CHIME packages (this you may need to duck type)
- [ ] Efficiently generate random numbers (use `RandomGen` which is already a dependency)

@tristpinsm feel free to add items to this one.

anjakefala:
Closes #52.

### DIFF
--- a/draco/synthesis/noise.py
+++ b/draco/synthesis/noise.py
@@ -34,8 +34,10 @@ import numpy as np
 from caput import config
 from caput.time import STELLAR_S
 
+from cora.util import nputil
+
+from ..core import task, containers, io
 from ..util import tools
-from ..core import task, containers
 
 
 class ReceiverTemperature(task.SingleTask):
--- a/draco/synthesis/noise.py
+++ b/draco/synthesis/noise.py
@@ -30,6 +30,7 @@ from caput import config
 from ..core import task, containers
 from ..util import tools
 from caput.time import STELLAR_S
+import randomgen
 
 
 class ReceiverTemperature(task.SingleTask):
@@ -57,6 +58,35 @@ class ReceiverTemperature(task.SingleTask):
             # Great an auto!
             if prod[0] == prod[1]:
                 data.vis[:, pi] += self.recv_temp
+
+        return data
+
+class GaussianInPlace(task.SingleTask):
+    '''Generates a Gaussian distributed noise dataset using the
+    the noise estimates of an existing dataset.
+
+    Attributes
+    ----------
+    Related to distinguishing different container types
+    We need to be able to support both draco and CHIME timestream types
+    '''
+
+    def process(self, data):
+        '''Generates a noise dataset, given the provided dataset's visibility weights'''
+        data.redistribute("freq")
+
+        vis = data.vis[:]
+        vis_weight = np.copy(data.weight)
+
+        # vis_weight elements are inverse variances
+        # we want the standard deviation
+        vis_weight_std = np.sqrt(1. / vis_weight)
+
+        with mpi_random_seed(self.seed):
+            noise_real = randomgen.generator.Generator.normal(scale=vis_weight_std)
+
+        for pi, prod in enumerate(data.index_map["prod"]):
+            vis[:, pi] = noise_real[:, pi]
 
         return data
 
--- a/draco/synthesis/noise.py
+++ b/draco/synthesis/noise.py
@@ -97,13 +97,11 @@ class GaussianNoiseDataset(task.SingleTask):
 
         """
 
-        data.redistribute("freq")
-
         vis = data.vis[:]
 
         # data.weight elements are inverse variances
         # we want the standard deviation
-        vis_weight_std = np.sqrt(1.0 / np.copy(data.weight))
+        vis_weight_std = np.sqrt(tools.invert_no_zero(data.weight))
 
         # create a random generator, and create a local seed state
         rg = randomgen.generator.RandomGenerator()
--- a/draco/synthesis/noise.py
+++ b/draco/synthesis/noise.py
@@ -81,9 +81,10 @@ class GaussianInPlace(task.SingleTask):
         # vis_weight elements are inverse variances
         # we want the standard deviation
         vis_weight_std = np.sqrt(1. / vis_weight)
+        rg = randomgen.generator.RandomGenerator()
 
-        with mpi_random_seed(self.seed):
-            noise_real = randomgen.generator.Generator.normal(scale=vis_weight_std)
+        with randomgen_mpi_random_seed(self.seed, rg) as rg:
+            noise_real = rg.normal(scale=vis_weight_std)
 
         for pi, prod in enumerate(data.index_map["prod"]):
             vis[:, pi] = noise_real[:, pi]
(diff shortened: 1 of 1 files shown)


anjakefala:
This pull request is not ready.
Notable things that must be done:
* add/update documentation
* incorporate seeding for randomgen

The PR was opened so that the logic can be reviewed while these things are worked on.

jrs65:
@anjakefala do you want me to add some comments?

anjakefala:
@jrs65 I think you are supposed to be on vacation. >=|

But they would be very welcome!

jrs65:
I'm not on vacation until next week really, so I'll have a look over it now.

jrs65:
Okay. It's looking good. Seems like you're grokking the pipeline framework. A bunch of comments but shouldn't be hard to address them.

anjakefala:
> Okay. It's looking good. Seems like you're grokking the pipeline framework. A bunch of comments but it shouldn't be hard to address them.

Thanks!

I do not see the comments, though. I think you need to submit a review?

jrs65:
Fail. I have now clicked submit.

draco/synthesis/noise.py
```diff
@@ -60,6 +61,35 @@ def process(self, data):
...
+    Related to distinguishing different container types
+    We need to be able to support both draco and CHIME timestream types
+    '''
+
+    def process(self, data):
+        '''Generates a noise dataset, given the provided dataset's visibility weights'''
+        data.redistribute("freq")
+
+        vis = data.vis[:]
+        vis_weight = np.copy(data.weight)
```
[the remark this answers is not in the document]
anjakefala: Nope! I misunderstood one aspect of implicit copying and viewing.

draco/synthesis/noise.py
```diff
@@ -60,6 +61,35 @@ def process(self, data):
...
+
+        # vis_weight elements are inverse variances
+        # we want the standard deviation
+        vis_weight_std = np.sqrt(1. / vis_weight)
```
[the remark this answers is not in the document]
anjakefala: It seems like `invert_no_zero` returns a "zero" in the cases where it receives a "zero". So that would mean a variance of 0, and the random number generator will return a 0. 

I think that is reasonable behaviour, though if we had some concept of "`None`" that would probably be better.

draco/synthesis/noise.py
```diff
@@ -60,6 +61,35 @@ def process(self, data):
...
+
+    Attributes
+    ----------
+    Related to distinguishing different container types
+    We need to be able to support both draco and CHIME timestream types
+    '''
+
+    def process(self, data):
+        '''Generates a noise dataset, given the provided dataset's visibility weights'''
+        data.redistribute("freq")
```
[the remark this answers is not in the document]
anjakefala: I am excited to understand the MPI aspect of this better...

draco/synthesis/noise.py
```diff
@@ -60,6 +61,35 @@ def process(self, data):
 
         return data
 
+class GaussianInPlace(task.SingleTask):
```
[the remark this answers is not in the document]
anjakefala: Why is naming the hardest part!

draco/synthesis/noise.py
```diff
@@ -61,6 +67,59 @@ def process(self, data):
...
+
+        Returns
+        -------
+        data_noise : same as :param:`data`
+            The previous dataset with the visibility replaced with
+            a Gaussian distributed noise realisation.
+
+        """
+        # Distribute in something other than `stack`
+        data.redistribute("freq")
```
fandinomat: Because now we need to loop over stacks to find the auto-correlations, I thought it simpler to ensure we are not distributed in `stack` so I put this `redistribute` call back here.

draco/synthesis/noise.py
```diff
@@ -61,6 +67,59 @@ def process(self, data):
...
+        # Visibility to be replaced by noise
+        vis = data.vis[:]
+
+        # create a random generator, and create a local seed state
+        rg = randomgen.generator.RandomGenerator()
+
+        with randomgen_mpi_random_seed(rg, self.seed) as rg:
+            noise = rg.normal(
+                    scale = np.sqrt(tools.invert_no_zero(data.weight[:])/2),
+                    size = (2,) + data.weight[:].shape)
```
fandinomat: By passing a `size` parameter to `rg.normal`, I we bypass the use of the variable for the standard deviation, which saves some memory.

draco/synthesis/noise.py
```diff
@@ -61,6 +67,59 @@ def process(self, data):
...
+            vis[:] = noise[0] + 1j * noise[1]
+
+        for si, prod in enumerate(data.index_map["stack"]["prod"]):
```
fandinomat: @anjakefala Here we need to use `index_map["stack"]` for the loop.
This is the same size as the `stack` axis of the visibility and gives the index corresponding to an example product in `index_map["prod"] that goes into the stack.

draco/synthesis/noise.py
```diff
@@ -61,6 +67,59 @@ def process(self, data):
...
+            prod_inputs = data.index_map["prod"][prod]
+            if (prod_inputs[0]==prod_inputs[1]):
+                # This is an auto-correlation
+                vis[:, si].real *= 2**0.5
```
fandinomat: The standard deviation of the auto-correlations is twice that of the real/imag parts of a cross-correlation.

draco/synthesis/noise.py
```diff
@@ -96,23 +96,26 @@ def process(self, data):
             a Gaussian distributed noise realisation.
 
         """
-
+        # Distribute in something other than `stack`
+        data.redistribute("freq")
+        # Visibility to be replaced by noise
```
anjakefala: Good comment!

anjakefala:
It looks good to me @fandinomat! Thanks for walking me through all of your changes. ^^

anjakefala:
I will resolve the conflicts.

anjakefala:
Note: this is probably also relevant: https://github.com/radiocosmology/draco/pull/60

anjakefala:
In this hot mess of an accidental `git commit --amend`, I modified the code to take advantage of @cahofer's `VisContainer` work.

@jrs65 and @fandinomat, please review.

draco/synthesis/noise.py
```diff
@@ -62,6 +68,59 @@ def process(self, data):
...
+        vis = data.vis[:]
+
+        # create a random generator, and create a local seed state
+        rg = randomgen.generator.RandomGenerator()
+
+        with randomgen_mpi_random_seed(rg, self.seed) as rg:
+            noise = rg.normal(
```
jrs65: This looks to me like it will fail `black`. Has it been run through it?

draco/synthesis/noise.py
```diff
@@ -331,7 +390,7 @@ def draw_complex_wishart(C, n):
 
 @contextlib.contextmanager
 def mpi_random_seed(seed, extra=0):
```
jrs65: You could maybe just merge these two, and set/reset both seeds in the same context manager. You could also add some parameters to control it, i.e. `def mpi_random_seed(seed, extra=0, set_numpy=True, set_randomgen=True)`

draco/synthesis/noise.py
anjakefala: It has now!

draco/synthesis/noise.py
anjakefala: mpi_randomgen needs to be passed the generator, as well as the seed, but I can check for the presence of a generator. If present, we can set/reset its seed.

anjakefala: Also, a bit complicated by the yield statement. randomgen needs to yield its generator back.

So I can merge them, but the decision of which they set/reset will likely be exclusive.

anjakefala:
Added a parameter to `mpi_random_seed`. It now accepts an option RandomGen generator object. If passed, it will set and yield that generator, instead of setting `numpy.random`.

anjakefala:
Many thanks to @saurabh-astro for giving the push for us to review this! hehehe

PR #58, opened 2019-11-07
feat(GaussianNoise): Consider redundancy of baselines

cahofer:
- if an optional manager object is passed to this task, it calculates the redundant baselines from the telescope object and simulates gaussian noise accordingly.

### DIFF
--- a/draco/synthesis/noise.py
+++ b/draco/synthesis/noise.py
@@ -27,7 +27,7 @@ import numpy as np
 
 from caput import config
 
-from ..core import task, containers
+from ..core import task, containers, io
 from ..util import tools
 from caput.time import STELLAR_S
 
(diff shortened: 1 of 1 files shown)


draco/synthesis/noise.py
```diff
@@ -115,16 +132,28 @@ def process(self, data):
...
-        # Calculate the number of samples
-        nsamp = int(self.ndays * dt * df)
+        # Calculate the number of samples, this is a 1D array for the prod axis.
+        nsamp = int(self.ndays * dt * df) * redundancy
         std = self.recv_temp / np.sqrt(2 * nsamp)
 
         with mpi_random_seed(self.seed):
-            noise_real = std * np.random.standard_normal((nfreq, nprod, ntime))
-            noise_imag = std * np.random.standard_normal((nfreq, nprod, ntime))
+            noise_real = std[np.newaxis, :, np.newaxis] * np.random.standard_normal(
```
jrs65: In here you could just use `cora.util.nputil.complex_std_normal` to save a couple of lines. You'd need to remove the factor of 2 in `std` and when you set the weights though.

cahofer: Ok this should be done.

cahofer:
Updated the PR like suggested!

PR #60, opened 2019-12-11
Ch/containers prodstack

cahofer:
A restructure of the container base classes in `draco.core.containers`
 

- feat(containers): Make a base class `VisContainer` for all child classes handling visibilities
- fix(MaskData): bug fixes

### DIFF
--- a/draco/core/containers.py
+++ b/draco/core/containers.py
@@ -22,6 +22,7 @@ Container Base Classes
 
     ContainerBase
     TODContainer
+    VisContainer
 
 Helper Routines
 ---------------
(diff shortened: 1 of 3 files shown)


draco/core/containers.py
```diff
@@ -479,6 +461,118 @@ def time(self):
...
+    attrs_from : `memh5.BasicCont`, optional
+        Another container to copy attributes from. Must be supplied as keyword
+        argument. This applies to attributes in default datasets too.
+    kwargs : dict
+        Should contain entries for all other axes.
+    """
+
+    def __init__(self, *args, **kwargs):
+        # Call initializer from `ContainerBase`
+        super(VisContainer, self).__init__(*args, **kwargs)
```
jrs65: I think this needs to go to the bottom of this constructor.

anjakefala:
Do we want to hold on merging https://github.com/radiocosmology/draco/pull/55 until this work is merged in? Then we can adapt the code to work with `VisContainer`s.

jrs65:
> Do we want to hold on merging #55 until this work is merged in? Then we can adapt the code to work with `VisContainer`s.

Good point. Looks like we could switch that code to using `prodstack`.

PR #63, opened 2019-12-13
feat(containers): Add input axis to weight dataset of StaticGainData.

ssiegelx:
In the pathfinder-era the weight dataset of StaticGainData
was being populated with the dynamic range of the eigenvalue
decomposition.  This is a single number per frequency.
Today we estimate the uncertainty on the gain as 1 / sigma^2
for each frequency AND input.

### DIFF
--- a/draco/core/misc.py
+++ b/draco/core/misc.py
@@ -40,16 +40,16 @@ class ApplyGain(task.SingleTask):
 
     def process(self, tstream, gain):
         """Apply gains to the given timestream.
-        
+
         Smoothing the gains is not supported for SiderealStreams.
-        
+
         Parameters
         ----------
         tstream : TimeStream like or SiderealStream
             Time stream to apply gains to. The gains are applied in place.
         gain : StaticGainData, GainData or SiderealGainData
             Gains to apply.
-            
+
         Returns
         -------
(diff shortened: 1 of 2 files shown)


draco/core/misc.py
```diff
@@ -65,7 +65,7 @@ def process(self, tstream, gain):
 
             # Get the weight array if it's there
             weight_arr = (
-                gain.weight[:][..., np.newaxis] if gain.weight is not None else None
+                gain.weight[:] if gain.weight is not None else None
```
jrs65: I think this needs the same addition of a `time` axis in here.

draco/core/misc.py
```diff
@@ -167,7 +167,7 @@ def process(self, tstream, gain):
         if weight_arr is not None:
 
             # Convert dynamic range to a binary weight and apply to data
-            gain_weight = (weight_arr[:] > 2.0).astype(np.float64)
+            gain_weight = (weight_arr[:] > 0.0).astype(np.float64)
             tstream.weight[:] *= gain_weight[:, np.newaxis, :]
```
jrs65: I think this needs the input axis to be cast into a product axis and then the mask applied to the timestream.

PR #65, opened 2019-12-13
Hybrid map maker

jrs65:
This is now ready for review. There's a few broader fixes in here, but the biggest change is the introduction of the componentised ring map maker.

I have tested this against the current ring map maker and it gives nearly identical maps, that differ at the 10^-3 level. Here is a comparison of the maps using uniform weighting:
Unknown-33
The top panel is the map from the new map maker (from -10 Jy to 10 Jy), the middle panel is the difference between the old and new maps with the colour scale multiplied by 1000x (i.e. it's from +/- 10 mJy). The third is the difference if I apply a ~10^-3 level scale to the new map to get them to match better.

This is the same comparison but for natural weighting
Unknown-34

One obvious question is: why is there any difference at all? Essentially it all comes down to how the different maps are renormalised for missing baselines. The original map maker constructs the weights for all baselines initially, and normalises by the sum of all the weights to guarantee that the output units are faithfully Jy / beam (i.e. if you take the pixel containing CygA it will have the value of the flux of CygA). The new map maker first constructs a grid of NS beams, and each column of EW visibilities is normalised independently. This guarantees that the hybrid visibilities (i.e. localised NS beam, fourier mode EW) which come out of the BeamformNS stage are in units of Jy. However, because the EW transform is done separately it's tough (and not clearly desirable) for this to give each cylinder separation the same relative weighting that they are given in the old map maker.

If the missing NS baselines were identical for each cylinder separation the map makers would give the same result. However, when the missing NS baselines are different the map makers will give different results. The biggest thing which breaks this is the masking of the autocorrelations, which is why the differences are at the ~1/1024 level.

Given that we never really sorted out the most theoretically desirable scheme for the old map maker, I don't think this difference is very problematic, but it would be good to hear if any one else (e.g. @ashill and @ssiegelx who aren't tagged for review) can think of any objections.

### DIFF
--- a/draco/analysis/ringmapmaker.py
+++ b/draco/analysis/ringmapmaker.py
@@ -326,8 +326,8 @@ class BeamformEW(task.SingleTask):
                 # Only need the 0th term of the irfft, equivalent to summing in
                 # then EW direction
                 m[:, 1:] *= 2.0  # Factor to include negative elements in sum below
-                bfm = np.sum(v * m, axis=1).real[:, np.newaxis, ...]
-                sb = np.sum(w * m, axis=1).real[:, np.newaxis, ...]
+                bfm = np.asarray(np.sum(v * m, axis=1)).real[:, np.newaxis, ...]
+                sb = np.asarray(np.sum(w * m, axis=1)).real[:, np.newaxis, ...]
             else:
                 bfm = np.fft.irfft(v * m, nbeam, axis=1) * nbeam
                 sb = np.fft.irfft(w * m, nbeam, axis=1) * nbeam
--- a/draco/analysis/ringmapmaker.py
+++ b/draco/analysis/ringmapmaker.py
@@ -16,13 +16,6 @@ Tasks
     MakeVisGrid
     BeamformNS
 """
-# === Start Python 2/3 compatibility
-from __future__ import absolute_import, division, print_function, unicode_literals
-from future.builtins import *  # noqa  pylint: disable=W0401, W0614
-from future.builtins.disabled import *  # noqa  pylint: disable=W0401, W0614
-
-# === End Python 2/3 compatibility
-
 import numpy as np
 import scipy.constants
 
--- a/draco/analysis/ringmapmaker.py
+++ b/draco/analysis/ringmapmaker.py
@@ -198,10 +198,6 @@ class BeamformNS(task.SingleTask):
         gsv = gstream.vis[:]
         gsw = gstream.weight[:].copy()
 
-        # Remove auto-correlations
-        if not self.include_auto:
-            gsw[..., 0, 0] = 0.0
-
         # Construct phase array
         el = self.span * np.linspace(-1.0, 1.0, self.npix)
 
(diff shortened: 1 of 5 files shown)


wulfda02 (approved):
Looks good to me!

draco/analysis/ringmapmaker.py
```diff
@@ -0,0 +1,449 @@
...
+            sstream.prodstack.view(np.uint16).reshape(-1, 2)
+            != self.telescope.uniquepairs
+        ):
+            raise ValueError(
+                "Products in sstream do not match those in the beam transfers."
+            )
+
+        # Calculation the set of polarisations in the data, and which polarisation
+        # index every entry corresponds to
+        polpair = self.telescope.polarisation[self.telescope.uniquepairs].view("U2")
```
tristpinsm, line 69: very clean solution!

draco/core/containers.py
```diff
@@ -1285,10 +1305,207 @@ class KLModes(SVDModes):
...
+    def weight(self):
+        return self.datasets["vis_weight"]
+
+    @property
+    def dirty_beam(self):
+        """This isn't useful at this stage, but it's needed to propagate onward."""
+        return self.datasets["dirty_beam"]
+
+
+class HybridVisMModes(FreqContainer):
```
tristpinsm, line 1399: this should also be an `MContainer` right?

draco/core/containers.py
```diff
@@ -1285,10 +1305,207 @@ class KLModes(SVDModes):
...
+    """Visibilities beamformed in the NS direction and m-mode transformed in RA.
+
+    This container has visibilities beam formed only in the NS direction to give a
+    grid in elevation.
+    """
+
+    _axes = ("pol", "freq", "ew", "el")
```
tristpinsm, line 1406: can remove `"freq"` here

draco/core/containers.py
```diff
@@ -717,7 +719,95 @@ def is_stacked(self):
...
+        """The physical frequency associated with each entry of the time axis.
+
+        By convention this property should return the frequency in MHz at the centre
+        of each of frequency channel.
+        """
+        try:
+            return self.index_map["freq"][:]["centre"]
+        # Need to check for both types as different numpy versions return
+        # different exceptions.
+        except (IndexError, ValueError):
```
tristpinsm, line 742: Nice. can't count the number of times a missing `"centre"` has crashed my code...

tristpinsm (changes requested):
Looks great! I found a few little things that might need attention.

draco/analysis/ringmapmaker.py
```diff
@@ -0,0 +1,449 @@
...
+                vmax = nsmax / wv
+
+            if self.weight == "inverse_variance":
+                gw = gsw[:, lfi].copy()
+            elif self.weight == "natural":
+                gw = gsr[:].astype(np.float32)
+            else:
+                x = 0.5 * (vpos / vmax + 1)
+                ns_weight = tools.window_generalised(x, window=self.weight)
+                gw = (gsw[:, lfi] > 0) * ns_weight[
```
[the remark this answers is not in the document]
jrs65: It's mostly in there to make sure it generates an array of the right shape.

PR #66, opened 2020-01-30
A few fixes to NS beamforming

tristpinsm:
- Fix bug preventing excluding autos
- Use the weight type parameter
- Normalise the beamformed output by the sum of the weights


tristpinsm:
@fandinomat these are the few fixes I made

PR #67, opened 2020-02-07
[WIP] Radiometer metric

saurabh-astro:
Pipeline task to create datasets for weights, and autocorrelation derived noise estimates.

### DIFF
--- a/draco/analysis/sensitivity.py
+++ b/draco/analysis/sensitivity.py
@@ -1,0 +1,252 @@
+# === Start Python 2/3 compatibility
+from __future__ import absolute_import, division, print_function, unicode_literals
+from future.builtins import *  # noqa  pylint: disable=W0401, W0614
+from future.builtins.disabled import *  # noqa  pylint: disable=W0401, W0614
+
+# === End Python 2/3 compatibility
+
+import numpy as np
+
+from caput import config
+
+from ..core import task, io, containers
+from ..util import tools
+
+
+class ComputeSystemSensitivity(task.SingleTask):
+    """Compute the sensitivity of beamformed visibilities.
+
+    Parameters
+    ----------
+    exclude_intracyl : bool
+        Exclude the intracylinder baselines in the sensitivity estimate.
+        Default is to use all baselines.  Note that a RuntimeError
+        will be raised if exclude_intracyl is True and the visibilities
+        have already been stacked over cylinder.
+    """
+
+    exclude_intracyl = config.Property(proptype=bool, default=False)
+
+    def setup(self, telescope):
+        """Save the telescope model.
+
+        Parameters
+        ----------
+        telescope : TransitTelescope
+        """
+        self.telescope = io.get_telescope(telescope)
+
+    def process(self, data):
+        """Estimate the sensitivity of the input data.
+
+        Parameters
+        ----------
+        data : TODContainer
+            Must have a weight property that contains an
+            estimate of the inverse variance of the noise
+            in each visibility.  The visibilities can be
+            stacked to any level of redundancy.
+
+        Returns
+        -------
+        metrics : SystemSensitivity
+            Contains the measured and radiometric estimates of
+            the noise in the beamformed visibilities.
+        """
+        # Ensure we are distributed over frequency. Get shape of visibilities.
+        data.redistribute("freq")
+
+        nfreq, nstack, ntime = data.vis.local_shape
+
+        # Calculate redundancy
+        cnt = tools.calculate_redundancy(
+            data.input_flags[:],
+            data.prod,
+            data.reverse_map["stack"]["stack"],
+            data.stack.size,
+        )
+
+        # Determine stack axis
+        stack_new, stack_flag = tools.redefine_stack_index_map(
+            self.telescope, data.input, data.prod, data.stack, data.reverse_map["stack"]
+        )
+
+        if not np.all(stack_flag):
+            self.log.warning(
+                "There are %d stacked baselines that are masked "
+                "in the telescope instance." % np.sum(~stack_flag)
+            )
+
+        ps = data.prod[stack_new["prod"]]
+        conj = stack_new["conjugate"]
+
+        prodstack = ps.copy()
+        prodstack["input_a"] = np.where(conj, ps["input_b"], ps["input_a"])
+        prodstack["input_b"] = np.where(conj, ps["input_a"], ps["input_b"])
+
+        # Figure out mapping between inputs in data file and inputs in telescope
+        try:
+            bt_keys = self.telescope.input_index
+        except AttributeError:
+            bt_keys = np.array(
+                np.arange(self.telescope.nfeed), dtype=[("chan_id", "u2")]
+            )
+            field_to_match = "chan_id"
+        else:
+            if "correlator_input" in data.input.dtype.fields:
+                field_to_match = "correlator_input"
+            else:
+                field_to_match = "chan_id"
+
+        tel_index = tools.find_keys(
+            bt_keys[field_to_match], data.input[field_to_match], require_match=False
+        )
+
+        # Use the mapping to extract polarisation and EW position of each input
+        input_pol = np.array(
+            [
+                self.telescope.polarisation[ti] if ti is not None else "N"
+                for ti in tel_index
+            ]
+        )
+
+        ew_position = np.array(
+            [
+                self.telescope.feedpositions[ti, 0] if ti is not None else 0.0
+                for ti in tel_index
+            ]
+        )
+
+        # Next we determine indices into the stack axis for each polarisation product
+        # The next three lines result in XY and YX being
+        # combined into a single polarisation product
+        pa, pb = input_pol[prodstack["input_a"]], input_pol[prodstack["input_b"]]
+        pol_a = np.where(pa <= pb, pa, pb)
+        pol_b = np.where(pa <= pb, pb, pa)
+
+        baseline_pol = np.core.defchararray.add(pol_a, pol_b)
+
+        if self.exclude_intracyl:
+            baseline_flag = (
+                ew_position[prodstack["input_a"]] != ew_position[prodstack["input_b"]]
+            )
+        else:
+            baseline_flag = np.ones(prodstack.size, dtype=np.bool)
+
+        pol_uniq = [bp for bp in np.unique(baseline_pol) if "N" not in bp]
+        pol_index = [
+            np.flatnonzero((baseline_pol == up) & baseline_flag) for up in pol_uniq
+        ]
+        npol = len(pol_uniq)
+
+        auto_flag = (prodstack["input_a"] == prodstack["input_b"]).astype(np.float32)
+
+        if self.exclude_intracyl and (np.sum(auto_flag) == npol):
+            raise ValueError(
+                "You have requested the exclusion of "
+                "intracylinder baselines,  however it appears "
+                "that the visibilities have already been stacked "
+                "over cylinder, preventing calculation of the "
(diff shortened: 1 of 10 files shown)


jrs65:
Hey Saurabh. We should figure things out here, I've done a bunch of work on this PR over the break, and I'm very close to having it finished. Let me push things to another branch.

draco/analysis/radiometer_test.py
```diff
@@ -0,0 +1,238 @@
...
+	    for j in ind_YY:
+		if i >= j: 
+		    continue 
+		
+                n_feeds_cyl_YY = live_feeds(chr(i+ord('A')-len(ind_XX)),'YY',binp)*live_feeds(chr(j+ord('A')-len(ind_XX)),'YY',binp)
+		vis_avg_YY +=  bvis[:,i,:] * bvis[:,j,:] * n_feeds_cyl_YY/feed_qnt**2
+		
+                Nbl_YY += n_feeds_cyl_YY
+		N_YY += n_feeds_cyl_YY/feed_qnt**2
+
```
ssiegelx: I think lines 185 - 220 could be replaced with something like this:

```
radiometer = np.zeros((nfreq, npol, ntime), dtype=np.float32)
radiometer_counter = np.zeros((nfreq, npol, ntime), dtype=np.float32)

for ii, (ai, pi) in enumerate(zip(auto_input, auto_pol)):
    
    for jj, (aj, pj) in enumerate(zip(auto_input, auto_pol)):
        
        if pi == pj and ii > jj:
            # Avoid double counting
            continue
            
        if not self.include_intracyl and (self.telescope.feedpositions[ai, 0] == self.telescope.feedpositions[aj, 0]):
            # Exclude intracylinder
            continue
            
        if pi != pj and (ai > aj):
            # Deals with oddities in cross-pol intracylinder baselines
            if self.include_intracyl and (self.telescope.feedpositions[ai, 0] == self.telescope.feedpositions[aj, 0]):
                pass
            else:
                continue
            
        pp = upol.index(auto_pol[ii] + auto_pol[jj])
        
        n2 = num_live_feed[np.newaxis, ii, :] * bflag[:, ii, :] * num_live_feed[np.newaxis, jj, :] * bflag[:, jj, :]
        
        radiometer_counter[:, pp, :] += n2
        radiometer[:, pp, :] += n2 * bvis[:, ii, :] * bvis[:, jj, :]```

File: draco/analysis/radiometer_test.py [hunk 2]
```diff
@@ -0,0 +1,238 @@
...
+	#Get the quantities for normalization: 
+	#integration time, number of baselines, fraction of packets lost and frequency resolution
+	tint = np.median(np.diff(time_auto)) #there is a slight different in integrations at micro seconds, ignoring it
+        spw  = np.median([i[1] for i in auto_data.index_map['freq']]) #assuming same spectral bandwidth across the frequency
```
ssiegelx: I think you have to scale from MHz to Hz here?
`spw = np.median(data.index_map['freq']['width']) * 1e6`

draco/analysis/radiometer_test.py
```diff
@@ -0,0 +1,238 @@
...
+
+	n_b_tau_XX = spw * tint * Nbl_XX * (1-bfrac)
+	n_b_tau_YY = spw * tint * Nbl_YY * (1-bfrac)
+
```
ssiegelx: If you did as suggested above, then this would become:
```
nint = spw * tint * (1.0 - bfrac[:, np.newaxis, :])
radiometer = np.sqrt(radiometer * tools.invert_no_zero(nint * radiometer_counter**2 ))
```
which could be saved directly to the container.

draco/analysis/radiometer_test.py
ssiegelx: so that you don't have to repeat the code and also so that we can handle both co- and cross-polar baselines.

jrs65:
Okay @saurabh-astro . It looks like @ssiegelx  has suggested most of the changes I have implemented, so I'm happy to just leave this to you both. That'll give me time to fix up Mateus's part of the RFI code which also needs work.

When do you think you can have this finished? I'm aiming to get the new data processing mid next week.

saurabh-astro:
> Okay @saurabh-astro . It looks like @ssiegelx has suggested most of the changes I have implemented, so I'm happy to just leave this to you both. That'll give me time to fix up Mateus's part of the RFI code which also needs work.
> 
> When do you think you can have this finished? I'm aiming to get the new data processing mid next week.

I think end of Wednesday is a good time for wrapping it up. Does that work?

ssiegelx:
@saurabh-astro I went ahead and implemented the changes suggested above.

There are some subtleties in the calculation of the radiometric estimate of the noise for the case where the data has been stacked and you want to include intra-cylinder baselines (specifically the cross-polar intra-cylinder baselines) that I am still trying to figure out.

I hope to have this finished in an hour or so, at which point I'll try to run the code on some recent chimestack files.  Then we can work on validation tomorrow by comparing to the output of the Dias analyzer (Zarif may be able to help with this).  Hopefully we can have this task ready by end of the day tomorrow.

ssiegelx:
@jrs65 @saurabh-astro This is ready to go.

Note that all past sensitivity estimates have averaged over inter-cylinder baselines only.  If we wanted to use just inter-cylinder baselines in the pipeline, then the `ComputeSystemSensitivity` task would have to come before `CollateProducts`, which would either require RFI masking on individual files or sidereal grouping the sensitivity containers separately.  The former may result in problems at the file edges.

The default behavior is to average over all baselines.  Saurabh and I have compared this to the inter-cylinder average and find only small differences.  The inter-cylinder baseline average appears to enhance RFI coming from overhead, like the scattered TV stations, whereas the all baseline average enhances some of the static 6 MHz wide bands that are not being caught by the real-time excision.  In both cases the enhancement is at the 10% level, so I would probably just use the all baseline average, since it will be much simpler to fit in the daily processing flow.

We should save the output of this task to disk for each sidereal day so that we can use it for classifying CSDs as good or bad.

saurabh-astro:
> @jrs65 @saurabh-astro This is ready to go.
> 
> Note that all past sensitivity estimates have averaged over inter-cylinder baselines only. If we wanted to use just inter-cylinder baselines in the pipeline, then the `ComputeSystemSensitivity` task would have to come before `CollateProducts`, which would either require RFI masking on individual files or sidereal grouping the sensitivity containers separately. The former may result in problems at the file edges.
> 
> The default behavior is to average over all baselines. Saurabh and I have compared this to the inter-cylinder average and find only small differences. The inter-cylinder baseline average appears to enhance RFI coming from overhead, like the scattered TV stations, whereas the all baseline average enhances some of the static 6 MHz wide bands that are not being caught by the real-time excision. In both cases the enhancement is at the 10% level, so I would probably just use the all baseline average, since it will be much simpler to fit in the daily processing flow.
> 
> We should save the output of this task to disk for each sidereal day so that we can use it for classifying CSDs as good or bad.

I completely agree with using all the baselines. I am including the sample dataset ratio metric computed with inter-cylinder and all baselines respectively. All-baseline case has slightly lower RMS (as expected) and the difference between the two cases are at sub-percent levels.

TT_int_cyl

TT_all_cyl

jrs65:
This is looking great. I'll have a look over it tomorrow to see if I have any final suggestions and then we can merge it. Thanks for getting it all together @ssiegelx and @saurabh-astro !

jrs65:
> Note that all past sensitivity estimates have averaged over inter-cylinder baselines only. If we wanted to use just inter-cylinder baselines in the pipeline, then the `ComputeSystemSensitivity` task would have to come before `CollateProducts`, which would either require RFI masking on individual files or sidereal grouping the sensitivity containers separately. The former may result in problems at the file edges.

@ssiegelx can you explain this one in more detail? From your comment here, and looking through the code, I don't think I really understand how the fact that the collation has already happened (such that there are no longer autos for each cylinder), prevents estimate of the  noise by the radiometer equation. It seems like you should be able to do it fine as the autos are all pretty similar anyway.

ssiegelx:
The code ensures that each variance `\sigma_{ij}^2` that is summed over in the numerator of the radiometer test has a corresponding radiometer estimate `V_{ii} * V_{jj}` in the denominator.

Since we only have the sum of the autocorrelations to work with, it is not possible to exclude intra-cylinder baselines from the denominator if you have already collated across cylinders.

If you assume that all autocorrelations are the same, then yes you could do it with a simple rescaling of the average autocorrelation by the number of baselines in the numerator.  Note that since the data has been calibrated to sky units, the autocorrelations will also contain feed-to-feed variations in the beam at the location of Cyg A.  My impression from working on this at the hack-a-thon was that if you did not use the same set of products, it resulted in artifacts in the radiometer test that dominated over the RFI.

draco/analysis/transform.py
```diff
@@ -178,94 +178,60 @@ def process(self, ss):
...
-            except ValueError:
-                return None
-
-        match_sn = True
-        if "stack" in ss.index_map:
-            match_sn = ss.index_map["stack"].size == ss.index_map["prod"].size
-
         # For each input in the file, find the corresponding index in the telescope instance
-        ss_keys = ss.index_map["input"][:]
+        ss_keys = ss.input
```
jrs65: @ssiegelx I think this might be a little better if I just add the `input_index` property onto `draco.core.telescope.TransitTelescope` to guarantee that it's present. That would allow removal of the `try... except` block. Thoughts?

It might also be worth implementing a `tools.find_inputs` function that will also figure out what field to match on. Let me know what you think. I should be able to do the implementation.

draco/analysis/sensitivity.py
```diff
@@ -0,0 +1,252 @@
...
+        pol_uniq = [bp for bp in np.unique(baseline_pol) if "N" not in bp]
+        pol_index = [
+            np.flatnonzero((baseline_pol == up) & baseline_flag) for up in pol_uniq
+        ]
+        npol = len(pol_uniq)
+
+        auto_flag = (prodstack["input_a"] == prodstack["input_b"]).astype(np.float32)
+
+        if self.exclude_intracyl and (np.sum(auto_flag) == npol):
+            raise RuntimeError(
```
jrs65: Probably should be `ValueError`.

draco/analysis/sensitivity.py
```diff
@@ -0,0 +1,252 @@
...
+    Parameters
+    ----------
+    exclude_intracyl : bool
+        Exclude the intracylinder baselines in the sensitivity estimate.
+        Default is to use all baselines.  Note that a RuntimeError
+        will be raised if exclude_intracyl is True and the visibilities
+        have already been stacked over cylinder.
+    """
+
+    exclude_intracyl = config.Property(proptype=bool, default=False)
```
jrs65: This doesn't completely make sense in a generic context, but we can figure out what (if anything) to do about it later.

draco/analysis/transform.py
ssiegelx: Yeah this block of code for matching inputs is repeated in at least three places: here, `CollateProducts`, and the new method `tools.redefine_stack_index_map`.  Moving it to a single `tools.find_inputs` method is probably a good idea.

draco/analysis/sensitivity.py
ssiegelx: Yeah, we originally had a cut on baseline distance for that reason.  But we don't have the freedom to make arbitrary baseline cuts on the stacked data and still do the exact radiometer estimate, so it was changed to this.

jrs65:
@ssiegelx I have (with lots of git rebase-fu) reshaped the commits into a coherent set. This all looks good now so I'm going to merge.

PR #72, opened 2020-02-19
Better output naming

jrs65:


draco/analysis/flagging.py
```diff
@@ -581,10 +581,7 @@ def process(self, sensitivity):
             finalmask[:, madtimes] = madmask[:, madtimes]
 
         # Collect all parts of the mask onto rank 1 and then broadcast to all ranks
-        finalmask = mpiarray.MPIArray.wrap(finalmask[np.newaxis, ...], 1).redistribute(
-            axis=0
-        )
-        finalmask = sensitivity.comm.bcast(finalmask[0], root=0)
+        finalmask = mpiarray.MPIArray.wrap(finalmask, 0).allgather()
```
tristpinsm: It took me a while to make sense of the old snippet here even knowing what behaviour the new line is recreating...

draco/core/task.py
```diff
@@ -381,7 +401,13 @@ def _save_output(self, output):
...
 
             # Construct the filename
-            outfile = self.output_root + str(tag) + ".h5"
+            name_parts = {
+                "tag": tag,
+                "count": self._count,
+                "task": self.__class__.__name__,
+                "key": self._out_keys[0],
+            }
+            outfile = self.output_name.format(**name_parts)
```
tristpinsm: `output_name` currently defaults to `None` which I think would raise an error here if left unset. Should it instead default to `""` as `output_root` did?

PR #73, opened 2020-02-19
feat(task): write metadata to file

### LINKED ISSUE #71: add version strings to output.attrs written by SingleTask._save_output
version strings should be in ~SingleTask._global_config~ `SingleTask._metadata`

nritsche:
In SingleTask._save_output, also write _metadata to disk (if it was set
by e.g. the manager that created the task.

Depends on https://github.com/radiocosmology/caput/issues/104

TODO:
- [x] remove branch from caput in requirements

Closes #71


draco/core/task.py
```diff
@@ -251,6 +251,8 @@ class SingleTask(MPILoggedTask, pipeline.BasicContMixin):
     output_root : string
         Pipeline settable parameter giving the first part of the output path.
         If set to 'None' no output is written.
+    metadata : dict
```
jrs65: This doesn't actually seem to be a controllable parameter.

draco/core/task.py
```diff
@@ -374,9 +383,20 @@ def finish(self):
 
     def _save_output(self, output):
         # Routine to write output if needed.
-
         if self.save and output is not None:
 
+            # add metadata to output
+            metadata = {
```
jrs65: Change me!

PR #76, opened 2020-02-21
Mmode inverse dev

fandinomat:
Add mmode inverse task.
In case the mmodes were padded with zeroes in the forward mmode transform, this will include the zeroes in the inverse transform.


draco/analysis/transform.py
```diff
@@ -552,6 +552,59 @@ def _pack_marray(mmodes, mmax=None):
     return marray
 
 
+class MModeInverseTransform(task.SingleTask):
```
jrs65: Doc string for the class.

draco/analysis/transform.py
```diff
@@ -552,6 +552,59 @@ def _pack_marray(mmodes, mmax=None):
...
+    def process(self, ma):
```
jrs65: Doc string for the method.

draco/analysis/transform.py
jrs65: Param for producing a sidereal stream with a specified number of samples

draco/analysis/transform.py
```diff
@@ -552,6 +552,59 @@ def _pack_marray(mmodes, mmax=None):
...
+
+        ma.redistribute("freq")
+        ntime = ma.vis.shape[0] * 2 - 1
+
+        # Re-construct array of S-streams
+        ssarray = _make_ssarray(ma.vis[:])  # , mmax)
```
jrs65: Remove commented out code.

draco/analysis/transform.py
```diff
@@ -552,6 +552,59 @@ def _pack_marray(mmodes, mmax=None):
...
+        sstream.weight[:] = ma.weight[0, 0, :, :][:, :, np.newaxis] / float(ntime)
+
+        return sstream
+
+
+def _make_ssarray(ma):  # , mmax):
+    # Construct an array of sidereal time streams from m-modes
+    #
+    marray = _unpack_marray(ma)
+    ssarray = np.fft.ifft(marray * marray.shape[-1], axis=-1)
```
tristpinsm: My implementation was basically identical, except that here I multiply by the equivalent of `marray.shape[-1]`. This is to undo the normalisation in `_make_marray`, which appears to enforce the opposite convention than is used in `numpy`. If the number of samples is different in the inverse transform though I'm not sure exactly how this will work out... Maybe @jrs65 should weigh in.

tristpinsm: Sorry I got confused in my previous comment. But what I'm wondering about is do we need to alter the normalisation if the number of samples is different in the inverse transform? (or if some modes were cut)

fandinomat: I have tested the new code and the normalization works fine now.

fandinomat:
I have implemented the changes.
I have tested the new code and the normalization works fine even with a larger number of points.

PR #77, opened 2020-02-28
feat(containerBase): support for sel_* params

nritsche:
caput.core.memh5 introduces sel_* parameters for from_file functions in
https://github.com/radiocosmology/caput/commit/54bc2ae58e383e3518549ed14efd8d61c2c20f07

(also see https://github.com/radiocosmology/caput/pull/111 )

This adds support for mapping axis selections to the axes of
ContainerBase.

TODO before merge:
- [x] Merge the corresponding PR in caput first and then remove the caput branch name from requirements.txt

### DIFF
--- a/draco/__init__.py
+++ b/draco/__init__.py
@@ -6,6 +6,7 @@ from future.builtins.disabled import *  # noqa  pylint: disable=W0401, W0614
 # === End Python 2/3 compatibility
 
 from ._version import get_versions
+from .core.containers import ContainerBase
 
 __version__ = get_versions()["version"]
 del get_versions
--- a/draco/__init__.py
+++ b/draco/__init__.py
@@ -6,7 +6,6 @@ from future.builtins.disabled import *  # noqa  pylint: disable=W0401, W0614
 # === End Python 2/3 compatibility
 
 from ._version import get_versions
-from .core.containers import ContainerBase
 
 __version__ = get_versions()["version"]
 del get_versions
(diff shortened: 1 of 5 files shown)


requirements.txt
```diff
@@ -1,5 +1,5 @@
 driftscan @ git+https://github.com/radiocosmology/driftscan.git
-caput @ git+https://github.com/radiocosmology/caput.git
+caput @ git+https://github.com/radiocosmology/caput.git@rn/h5sel
```
nritsche: Merge the corresponding PR in caput first and then remove the branch from here

draco/__init__.py
```diff
@@ -6,6 +6,7 @@
 # === End Python 2/3 compatibility
 
 from ._version import get_versions
+from .core.containers import ContainerBase
```
jrs65: Why?

draco/core/containers.py
```diff
@@ -316,6 +316,47 @@ def axes(self):
...
+            the selections for those.
+        """
+        # Check if all those axes exist
+        for axis in sel_args.keys():
+            if axis not in self._axes:
+                raise RuntimeError("No '{}' axis found to select from.".format(axis))
+
+        # Build selections dict
+        selections = {}
+        for name, dataset in self._dataset_spec.items():
```
jrs65: This will insert dataset selection entries even when they have no relevant axes. Should you just leave them missing?

PR #78, opened 2020-02-28
fix(ContainerBase): convert strings in containers by default

jrs65:


jrs65:
@tristpinsm I hadn't pushed it apparently, so here it is for you to review!

tristpinsm:
> This fixes the problem I was having writing out files with unicode attributes. Thanks!
> 
> However I'm still having an issue running `ch_util.reorder_correlator_inputs` with the inputs from the index map of a `CorrData` file because they are read in as bytestrings. This should be as simple as adding similar parameters to `CorrData` in `ch_util`, right?

Actually seems like that change was already made... The `input` index map still comes out with serials as bytestrings. Do the changes to `memh5` work with compound data types too?

tristpinsm:
> > This fixes the problem I was having writing out files with unicode attributes. Thanks!
> > However I'm still having an issue running `ch_util.reorder_correlator_inputs` with the inputs from the index map of a `CorrData` file because they are read in as bytestrings. This should be as simple as adding similar parameters to `CorrData` in `ch_util`, right?
> 
> Actually seems like that change was already made... The `input` index map still comes out with serials as bytestrings. Do the changes to `memh5` work with compound data types too?

It seems like all this should be working, but when I read a `26mgated` file with `CorrData.from_acq_h5` I get bytestrings in the input index map, but everything else I can check seems to be unicode. I haven't been able to figure out why yet...

jrs65:
Yeah sorry. I’ve fixed that one too. I opened a PR in caput that fixes it,
but Rick hasn’t reviewed it yet. If you can have a quick look, then I’ll
merge it.

tristpinsm:
I don't see any open PRs in `caput` or `ch_util` related to this. If you point me to it I'll be glad to take a look.

jrs65:
Hmmm. I’m pretty sure I pushed the branch. Maybe I didn’t open a PR. Let me check.

PR #80, opened 2020-03-03
fix(SVDModeProject): don't take attrs from MModes

sjforeman:
In `analysis.fgfilter.SVDModeProject`, when the new `SVDModes` container is initialized, we don't want to inherit attributes from the parent `MModes` container: if we do, then the `vis` dataset in `SVDModes` will have `axes` equal to `["m", "msign", "freq", "stack"]`, when it should have `["m", "mode"]`. The same applies when we do a backwards projection from SVD modes into m-modes.

### DIFF
--- a/draco/analysis/fgfilter.py
+++ b/draco/analysis/fgfilter.py
@@ -80,7 +80,7 @@ class SVDModeProject(_ProjectFilterBase):
         tel = bt.telescope
 
         svdmodes = containers.SVDModes(
-            mode=bt.ndofmax, axes_from=mmodes, attrs_from=mmodes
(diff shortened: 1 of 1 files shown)


jrs65:
This is a known problem, but this is not quite the correct fix. What we need to do is make sure that generally `attrs_from` does not overwrite the `axes` attributes on each dataset. I'm working on a fix for that now. Will tag you when it's done.

sjforeman:
Ah, okay! No urgency at all at my end. Let me know if you want me to do anything.

jrs65:
Superseded by #84

PR #83, opened 2020-03-08
Sens frac lost

jrs65:


draco/analysis/sensitivity.py
```diff
@@ -225,19 +225,19 @@ def process(self, data):
         dnu = np.median(data.index_map["freq"]["width"]) * 1e6
 
         if ("flags" in data) and ("frac_lost" in data["flags"]):
-            frac_lost = data["flags"]["frac_lost"][:][:, np.newaxis, :]
+            frac_lost = data["flags"]["frac_lost"][:]
         else:
-            frac_lost = 0.0
+            frac_lost = np.zeros((1, 1, 1))
```
ssiegelx: This array should be two dimensions (freq, time):
```
frac_lost = np.zeros((1, 1))
```

PR #85, opened 2020-03-11
Fix crash print

jrs65:
Fix a crash when the SiderealGrouper runs with no inputs and remove some useless print statements.


jrs65:
Thanks Rick!

PR #89, opened 2020-04-26
fix(beamform): convert sources to CIRS coordinates before beamforming

### LINKED ISSUE #87: Convert ICRS to CIRS coordinates while beam forming

jrs65:
The coordinates of the sources used in beamforming must be converted to
CIRS positions before beamforming or they'll be offset by the precessing
of the polar axis since J2000.

Fixes #87

### DIFF
--- a/draco/analysis/beamform.py
+++ b/draco/analysis/beamform.py
@@ -6,9 +6,11 @@ from future.builtins.disabled import *  # noqa  pylint: disable=W0401, W0614
 # === End Python 2/3 compatibility
 
 import numpy as np
-from mpi4py import MPI
+from skyfield.api import Star, Angle
 
-from caput import config, mpiutil, mpiarray
+from caput import config
+from caput import time as ctime
+
 from cora.util import units
 
 from ..core import task, containers, io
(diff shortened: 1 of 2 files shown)


draco/analysis/beamform.py
```diff
@@ -495,10 +497,20 @@ def _process_data(self, data):
...
             self.is_sstream = True
             self.ra = data.index_map["ra"]
+
+            # Calculate the epoch for the data so we can calculate the correct
+            # CIRS coordinates
+            if "lsd" not in data.attrs:
+                raise ValueError(
+                    "SiderealStream must have an LSD attribute to calculate the epoch."
+                )
+            self.epoch = self.telescope.lsd_to_unix(int(data.attrs["lsd"]))
```
fandinomat, line 507: I think we need a provision here for when `data.attrs["lsd"]` is a list (sidereal stacks).
Alternatively, it could go in `icrs_to_cirs`.

fandinomat:
I'll run this against mine to check we get the same result.

draco/analysis/beamform.py
jrs65: Thanks Mateus. That's an excellent point.

PR #92, opened 2020-04-29
Container copy

jrs65:
Beyond the few small ancillary commits (changing CI to Github; a few bug fixes), the main point of this PR is to add the ability to easily copy containers while sharing select datasets with their source.

A key area this is useful for is for masking tasks which take a container and modify the weights dataset. At the moment they primarily do this *in place*, which adds some awkward constraints in the pipeline. They could do a full copy, but the memory usage may be prohibitive. This allows them to create an independent copy of the weights dataset, which gets modified, but share the vis dataset, saving a bunch of memory.


draco/core/containers.py
```diff
@@ -366,6 +379,51 @@ def _make_selections(self, sel_args):
...
+            The copied container.
+        """
+        new_cont = self.__class__(attrs_from=self, axes_from=self, skip_datasets=True)
+
+        # Loop over datasets that exist in the source and either add a view of
+        # the source dataset, or perform a full copy
+        for name, data in self.datasets.items():
+
+            if shared and name in shared:
+                # TODO: find a way to do this that doesn't depend on the
```
tristpinsm, line 411: maybe you could use `MemDataset.view` on the right side, but I don't know how to access the left-hand side another way...

test/test_containers.py
```diff
@@ -24,9 +25,51 @@ def test_attrs_from(ss_container):
...
+    assert len(ts.weight.attrs) == 2
     assert tuple(ts.vis.attrs["axis"]) == ("freq", "stack", "time")
     assert tuple(ts.weight.attrs["axis"]) == ("freq", "stack", "time")
+
+
+def test_copy(ss_container):
+
+    ss_container.vis[:] = np.arange(16)
+    ss_container.weight[:] = np.arange(16)
+    ss_container.weight.attrs
```
tristpinsm, line 41: is this a forgotten line?

test/test_containers.py
```diff
@@ -24,9 +25,51 @@ def test_attrs_from(ss_container):
...
+    assert ss_copy.vis.attrs["test_attr4"] == "hello4"
+    assert "test_attr5" not in ss_copy.weight.attrs
+
+    # These tests only make sense when running across multiple MPI processes
+    assert ss_container.vis.distributed_axis == 0
+    assert ss_copy.vis.distributed_axis == 0
+
+    current_shape = ss_copy.weight.local_shape
+
+    ss_container.redistribute("ra")
```
tristpinsm, line 72: I hadn't realised on first pass that calling `redistribute` on either of the shared copies will lead to the other one having different distribution schemes between its datasets. This is tricky and could easily cause some funny behaviour if the user is not careful. It might be good to explicitly document this behaviour in `ContainerBase.copy` and maybe even in the tasks that make use of a `shared` config property.

tristpinsm:
Oh, and I also noticed that the MPI CI tests have been running for over an hour. I wonder what is going on...

test/test_containers.py
jrs65: Errr yes.

draco/core/containers.py
jrs65: That's what I had originally in here. However you don't want to do that because although initially it shares a copy of the data, as soon as you redistribute it doesn't. That's because it they only share the internal `._data` in the MemDataset which gets replaced during redistribution. It's subtle. Probably should comment on that explicitly.

test/test_containers.py
jrs65: Yah. It's a little tricky I agree. You're right that we need to make sure that any class that does this clearly documents it, I tried that on the one task I've modified to do it, but it should maybe be even clearer.

jrs65:
The idea of an `.is_shared` property is really good idea. I'm just not exactly sure how to implement it at the moment. To do it properly you'd need some kind of reference counting. An easier version would be to mark the dataset as "shared" as soon as you create a shared copy, but never unmark them, as it's very hard to track when you can actually do that without reference counting.

Part of the context here is that Rick is probably going to do a big rewrite of memh5, and I don't want to add in significant complexity before that.

tristpinsm:
> The idea of an `.is_shared` property is really good idea. I'm just not exactly sure how to implement it at the moment. To do it properly you'd need some kind of reference counting. An easier version would be to mark the dataset as "shared" as soon as you create a shared copy, but never unmark them, as it's very hard to track when you can actually do that without reference counting.
> 
> Part of the context here is that Rick is probably going to do a big rewrite of memh5, and I don't want to add in significant complexity before that.

The latter option is what I had in mind. That would be easy, but I guess it's not unusual for a copy to get unallocated while the other one is still in use and that wouldn't be tracked...

If Rick is going to reimplement everything, then maybe it's not worth putting effort into making this idiot-proof. It likely won't be used a ton before then (although this memh5 stuff tends to drag on...). But we could make sure memory sharing between groups is on the tasklist for the rewrite.

tristpinsm:
and it looks like the MPI tests failed after less than a second but then didn't exit for nearly another 3h.

jrs65:
> > The idea of an `.is_shared` property is really good idea. I'm just not exactly sure how to implement it at the moment. To do it properly you'd need some kind of reference counting. An easier version would be to mark the dataset as "shared" as soon as you create a shared copy, but never unmark them, as it's very hard to track when you can actually do that without reference counting.
> > Part of the context here is that Rick is probably going to do a big rewrite of memh5, and I don't want to add in significant complexity before that.
> 
> The latter option is what I had in mind. That would be easy, but I guess it's not unusual for a copy to get removed while the other one is still in use and that wouldn't be tracked...
> 
> If Rick is going to reimplement everything, then maybe it's not worth putting effort into making this idiot-proof. It likely won't be used a ton before then (although this memh5 stuff tends to drag on...). But we could make sure memory sharing between groups is on the tasklist for the rewrite.

I'll create an issue for this one.

PR #94, opened 2020-05-04
Apply relative threshold per baseline

jrs65:
Because of the large variation in weights across stacked baselines it's more appropriate to do the thresholding relative to the mean level of each baseline. Without this it's impossible to set appropriate thresholding for the delay spectrum estimator.

@ssiegelx I think you only use this task with the absolute threshold, right? Is this change going to break anything you do?

### DIFF
--- a/draco/analysis/flagging.py
+++ b/draco/analysis/flagging.py
@@ -384,7 +384,9 @@ class SmoothVisWeight(task.SingleTask):
 class ThresholdVisWeight(task.SingleTask):
     """Set any weight less than the user specified threshold equal to zero.
 
-    Threshold is determined as `maximum(absolute_threshold, relative_threshold * mean(weight))`.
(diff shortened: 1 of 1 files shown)


jrs65:
There's not as much variation as you'd think. This is an arbitrary baseline that I picked, where I've grayed out the points that are masked out image

jrs65:
At the moment I'm applying quite an aggressive thresholding cut (=0.5), which gives very good results for the delay spectrum estimator. I was a little worried that the frequency low frequency ripples may get cut accidentally but that doesn't seem to be happening. I think the fact that half the band is already weight~0, and the 30 MHz ripple reduces the weights substantially over a good part of the band means that cut is not as aggressive as you might think.

This plot is the weight threshold applied to the same baseline I showed above:
image

PR #96, opened 2020-05-06
Ch/psestimate memory

cahofer:
A quadratic estimator that is distributed over power spectrum bands.

feat(QuadraticPSEstimationXLarge): 
- a pipeline task to perform the m-mode quadratic powerspectrum estimation distributed over power spectrum bands. This solves the memory error problem when performing power spectrum estimation for large telescope classes. Depends on [driftscan PR #91](https://github.com/radiocosmology/driftscan/pull/91)


cahofer:
This PR goes along with the one in driftscan. Asking also @tristpinsm and @sjforeman additionally to @jrs65 for review. Thanks!

draco/analysis/powerspectrum.py
```diff
@@ -69,13 +71,110 @@ def process(self, klmodes):
...
+        elif self.pstype == "minimum_variance":
+            M = np.diag(fisher.sum(axis=1) ** -1)
+
+        ps.powerspectrum[:] = np.dot(M, q - bias).reshape(nperp, npar)
+        ps.C_inv[:] = fisher.reshape(nperp, npar, nperp, npar)
+
+        return ps
+
+
+class QuadraticPSEstimationXLarge(QuadraticPSEstimation):
```
jrs65, line 112: Why does this need to be a separate class at all? Is there any reason this can't just replace the existing implementation? Or if not, merge them and change the behaviour to a config option, or even better, something that gets automatically determined based on the input data size.

draco/analysis/powerspectrum.py
```diff
@@ -69,13 +71,110 @@ def process(self, klmodes):
...
         for mi, m in klmodes.vis[:].enumerate(axis=0):
             ps_single = pse.q_estimator(m, klmodes.vis[m, : klmodes.nmode[m]])
             q_list.append(ps_single)
 
         q = klmodes.comm.allgather(np.array(q_list).sum(axis=0))
         q = np.array(q).sum(axis=0)
 
+        et = time.time()
+        if mpiutil.rank0:
+            print("m, time needed for quadratic pse", m, et - st)
```
jrs65, line 85: This looks like left over debug statements. Either remove it, or at least improve the formatting.

draco/analysis/powerspectrum.py
cahofer: I mean, I could put them into one class and add a config option, but they two classes are fundamentally very different for the following reasons:
- the `QuadraticPSEstimationLarge` uses from `driftscan` the `PSMontelCarloLarge` class in which the clarray and the qestimator are MPIarrays parallelized over power spectrum bands. This is very different from the original code, in which clarray is gathered on one rank and qestimators are calculated locally. 
- This class respects the fact that clarray and qestimator are parallelized over nbands, therefore the sky data need to be pre-computed for all m first as each rank needs all m's. This is also very different from the original code, where the sky data were handled parallel in m. 

If we were to use a config option - then we would have one class but and if/else statement in the process method handling the two different calculations. 

I personally thought a separate class is cleaner, but after all - you are the boss ;)

draco/analysis/powerspectrum.py
sjforeman: Would be good to keep, but agree about improving formatting

sjforeman (changes requested):

draco/analysis/powerspectrum.py
```diff
@@ -62,13 +64,111 @@ def process(self, klmodes):
...
```
jrs65, line 78:
```suggestion
        et = time.time()
        self.log.debug(f"{m=}, time needed for quadratic pse {et - st}")
```

jrs65 (changes requested):

PR #100, opened 2020-08-12
Fix selections with inherited axes

jrs65:
This fixes the issues which was with the incorrect `._axes` attribute being checked because the correct `.axes` method can't be called from a class method. This sorts all of that out.

### DIFF
--- a/test/test_selections.py
+++ b/test/test_selections.py
@@ -1,4 +1,4 @@
-from draco.core.containers import SiderealGainData
+from draco.core.containers import GainData
 
 from caput import mpiarray, mpiutil
 
@@ -21,16 +21,17 @@ dset2 = dset2.reshape((len_axis, len_axis))
 
 freqs = np.arange(len_axis)
 inputs = np.arange(len_axis)
-ra = np.arange(len_axis)
+times = np.arange(len_axis)
 
 fsel = slice(5)
 isel = slice(1, 4)
+tsel = slice(1, 4)
 
 
 @pytest.fixture
 def container_on_disk():
     fname = "tmp_test_memh5_select.h5"
-    container = SiderealGainData(freq=freqs, input=inputs, ra=ra)
+    container = GainData(freq=freqs, input=inputs, time=times)
     container.create_dataset("gain", data=dset1.view())
     container.create_dataset("weight", data=dset2.view())
     container.save(fname)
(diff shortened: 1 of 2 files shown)


jrs65:
Requires radiocosmology/caput#123

test/test_selections.py
```diff
@@ -87,37 +88,44 @@ def container_on_disk_distributed():
...
-    assert np.all(m["gain"][:] == dset1[(fsel, isel, slice(None))])
-    assert np.all(m["weight"][:] == dset2[(fsel, slice(None))])
+    m = GainData.from_file(
+        container_on_disk, freq_sel=fsel, input_sel=isel, time_sel=tsel
+    )
+    assert np.all(m["gain"][:] == dset1[(fsel, isel, tsel)])
+    assert np.all(m["weight"][:] == dset2[(fsel, tsel)])
     assert np.all(m.index_map["freq"] == freqs[fsel])
     assert np.all(m.index_map["input"] == inputs[isel])
+    assert np.all(m.index_map["time"] == inputs[tsel])
```
ssiegelx, line 98:
```suggestion
    assert np.all(m.index_map["time"] == times[tsel])
```

test/test_selections.py
```diff
@@ -87,37 +88,44 @@ def container_on_disk_distributed():
...
+        time_sel=tsel,
+        distributed=True,
     )
-    assert np.all(m["gain"][:] == dset1[(fsel, isel, slice(None))])
-    assert np.all(m["weight"][:] == dset2[(fsel, slice(None))])
+    assert np.all(m["gain"][:] == dset1[(fsel, isel, tsel)])
+    assert np.all(m["weight"][:] == dset2[(fsel, tsel)])
     assert np.all(m.index_map["freq"] == freqs[fsel])
     assert np.all(m.index_map["input"] == inputs[isel])
+    assert np.all(m.index_map["time"] == inputs[tsel])
```
ssiegelx, line 115:
```suggestion
    assert np.all(m.index_map["time"] == times[tsel])
```

test/test_selections.py
```diff
@@ -87,37 +88,44 @@ def container_on_disk_distributed():
...
-        )
+        md = GainData.from_file(container_on_disk_distributed, distributed=False)
 
-        assert np.all(md["gain"][:] == dset1[(fsel, isel, slice(None))])
-        assert np.all(md["weight"][:] == dset2[(fsel, slice(None))])
+        assert np.all(md["gain"][:] == dset1[(fsel, isel, tsel)])
+        assert np.all(md["weight"][:] == dset2[(fsel, tsel)])
         assert np.all(md.index_map["freq"] == freqs[fsel])
         assert np.all(md.index_map["input"] == inputs[isel])
+        assert np.all(md.index_map["time"] == inputs[tsel])
```
ssiegelx, line 128:
```suggestion
        assert np.all(md.index_map["time"] == times[tsel])
```

test/test_selections.py
jrs65: Ooops. This is a fail! I'll post a new commit to fix all this up.

PR #103, opened 2020-08-31
Axis selections when loading containers

jrs65:


draco/core/io.py
```diff
@@ -266,13 +315,62 @@ def process(self):
...
+            if not isinstance(v, int):
+                raise ValueError(f"All elements of range spec must be ints. Got {x}")
+
+        return slice(*x)
+
+    def _parse_index(self, x):
+        # Parse and validate an index type selection
+
+        if not isinstance(x, (list, tuple)) or len(x) == 0:
+            raise ValueError(f"Index spec must a non-empty list or tuple. Got {x}.")
```
nritsche, line 359:
```suggestion
            raise ValueError(f"Index spec must be a non-empty list or tuple. Got {x}.")
```

draco/core/io.py
```diff
@@ -266,13 +315,62 @@ def process(self):
...
+
+            sel[f"{axis_name}_sel"] = sel_parsers[type_](self.selections[k])
+
+        return sel
+
+    def _parse_range(self, x):
+        # Parse and validate a range type selection
+
+        if not isinstance(x, (list, tuple)) or len(x) > 3 or len(x) < 2:
+            raise ValueError(f"Range spec must a length 2 or 3 list or tuple. Got {x}.")
```
nritsche, line 347:
```suggestion
            raise ValueError(f"Range spec must be a length 2 or 3 list or tuple. Got {x}.")
```

nritsche:
Depends on https://github.com/radiocosmology/caput/pull/139

Closes https://github.com/chime-experiment/Pipeline/issues/79

jrs65:
Fixed, and I'm going to merge.

I'm not going to put any tests in here, but I think we should probably come up with a plan for how to test the pipeline more thoroughly. At the moment we have pretty decent unit tests for library code, but we don't ever test the tasks. I'll create an issue for this one

jrs65:
Issue created in chime-experiment/Pipeline#80. However, I'm going to quickly try and write a test for this now. At least to get some idea of how these tests might work.

PR #105, opened 2020-09-10
feat(core.containers): HEALPixBeam container.

tristpinsm:
Used as an input to the driftscan beam transfer matrix calculation.

### DIFF
--- a/draco/core/containers.py
+++ b/draco/core/containers.py
@@ -1157,8 +1157,14 @@ class GridBeam(FreqContainer):
 
 
 class HEALPixBeam(FreqContainer, HealpixContainer):
-    """Generic container for representing the 2-d beam in spherical
-    coordinates on a HEALPix grid.
+    """Container for representing the spherical 2-d beam in a HEALPix grid.
+
+    Parameters
+    ----------
+    ordering : {"nested", "ring"}
+        The HEALPix ordering scheme used for the beam map.
+    coords : {"celestial", "galactic", "telescope"}
+        The coordinate system that the beam map is defined on.
     """
 
     _axes = ("pol", "input")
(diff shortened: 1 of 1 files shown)


draco/core/containers.py
```diff
@@ -781,7 +784,7 @@ class SiderealStream(VisContainer):
         The number of points to divide the RA axis up into.
     """
 
-    _axes = ("ra",)
+    _axes = ("ra", "freq", "stack", "input")
```
jrs65, line 787: Why did you change this? They're inherited from VisContainer.

draco/core/containers.py
```diff
@@ -1064,6 +1067,73 @@ def phi(self):
...
+    def pol(self):
+        return self.index_map["pol"]
+
+    @property
+    def input(self):
+        return self.index_map["input"]
+
+    @property
+    def pix(self):
+        return self.index_map["pix"]
```
jrs65, line 1130: I probably wouldn't bother exposing this one as a property. It's not very useful to get an array of consecutive numbers. The only interesting thing is its length.

draco/core/containers.py
```diff
@@ -1064,6 +1067,73 @@ def phi(self):
...
+            "distributed": True,
+            "distributed_axis": "freq",
+        },
+    }
+
+    def __init__(self, coords="unknown", ordering="unknown", *args, **kwargs):
+
+        super(HEALPixBeam, self).__init__(*args, **kwargs)
+        self.attrs["coords"] = coords
+        self.attrs["ordering"] = ordering
```
jrs65, line 1098: You should document (in the class docstring) what `coords` and `ordering` are for (listing possible values if that's reasonable).

jrs65 (changes requested):

draco/core/containers.py
tristpinsm: I don't know, this doesn't have anything to do with the PR, it must just have slipped in.

jrs65:
@tristpinsm I'm trying to get through the massive back log of changes to merge. This looks good to me, but maybe you could quickly change it to be a subclass of the new HealpixContainer?

tristpinsm:
ok, I did that, and it looks like the CI is happy now

tristpinsm:
@jrs65 I can't merge until you approve

jrs65 (approved):
Approved. I just added a quick commit to use proper numpydoc format, squash it if you like.

tristpinsm:
ok, thanks! black caught you leaving whitespace, so I'll fix that and squash everything into on commit.

PR #106, opened 2020-10-02
fix(noise): RandomGenerator rename and other GaussianNoiseDataset bugs

ssiegelx:
- randomgen 1.18 renamed RandomGenerator to Generator.
This commit renames Generator back to RandomGenerator on import
for synthesis.noise, in the same way as was done in analysis.delay.
- Fixes bug in how autocorrelations are being identified in GaussianNoiseDataset
- Fixes inconsistent variable name in mpi_random_seed
(both 'gen' and 'randomgen' are used to refer to the same variable)

### DIFF
--- a/draco/synthesis/noise.py
+++ b/draco/synthesis/noise.py
@@ -28,8 +28,13 @@ from future.builtins.disabled import *  # noqa  pylint: disable=W0401, W0614
 
 import contextlib
 
-import randomgen
 import numpy as np
+
+try:
+    # For backwards compatibility
+    from randomgen import RandomGenerator
+except ImportError:
+    from randomgen import Generator as RandomGenerator
 
 from caput import config
 from caput.time import STELLAR_S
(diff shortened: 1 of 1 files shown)


ssiegelx:
This applies the same fix to `synthesis.noise` as was applied to `analysis.delay`
regarding `randomgen` version 1.18 renaming `RandomGenerator` to `Generator`.

Note that currently `randomgen` is spamming the logs with this warning message:

   ```
707.6s [MPI 66/88] - WARNING  py.warnings: /home/ssiegel/ch_pipeline_rev01/venv/src/draco/draco/synthesis/noise.py:106: FutureWarning: Generator is deprecated and will be removed sometime after the release of
NumPy 1.21 (or 2 releases after 1.19 if there is a major release).

Unique features of Generator have been moved to
randomgen.generator.ExtendedGenerator.

Now is the time to start using numpy.random.Generator.

In the mean time Generator will only be updated for the most egregious bugs.

You can silence this warning using

import warnings
warnings.filterwarnings("ignore", "Generator", FutureWarning)

  rg = randomgen.generator.Generator()
```

So perhaps we should modify this PR to use `ExtendedGenerator` or `numpy`?
It looks like `ExtendedGenerator` adds several distributions not included
in `numpy`, specifically `complex_normal`, `multivariate_normal`, and `uintegers`.
I don't think these distributions are being used at the moment, although
I could imagine using something like `complex_normal` in the future.

jrs65:
I just started changing some of this in my PR as well, but let's merge what you have here, and I'll rebase on top.

It seems like we should just drop randomgen entirely in favour of numpy. We'd need to bump the minimum version up to 1.17, but I think that's fine. If you agree, I'll go that in my next PR.

anjakefala:
@jrs65 (for your next PR: https://github.com/radiocosmology/draco/issues/56)

ssiegelx:
@jrs65 Okay sounds good.  I'm fine with moving to numpy.

PR #108, opened 2020-10-15
ss/regrid

ssiegelx:
Still doing some validation, but issuing this PR now to get any feedback on the code.

### DIFF
--- a/draco/analysis/sidereal.py
+++ b/draco/analysis/sidereal.py
@@ -24,6 +24,7 @@ into  :class:`SiderealGrouper`, then feeding that into
 import numpy as np
 
 from caput import config, mpiutil, mpiarray, tod
+from cora.util import units
 
 from .transform import Regridder
 from ..core import task, containers, io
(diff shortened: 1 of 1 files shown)


draco/analysis/sidereal.py
```diff
@@ -182,8 +183,13 @@ class SiderealRegridder(Regridder):
...
         Ratio of signal covariance to noise covariance (used for Wiener filter).
+    down_mix: bool
+        Down mix the visibility prior to interpolation using the fringe rate
+        of a source at zenith.  This is un-done after the interpolation.
     """
 
+    down_mix = config.Property(proptype=bool, default=False)
+
     def setup(self, manager):
         """Set the local observers position.
```
jrs65, line 194: I think we want to test in here if the observer is a Telescope or not, and whether down_mix has also been set.

jrs65 (changes requested):
Really nice. Just the one change we discussed earlier.

jrs65:
Also you'll need to rebase onto master I think.

jrs65:
Actually, I'll just do that myself while merging things together.

draco/analysis/sidereal.py
jrs65: Actually, looking a bit more closely at the broader code, it's pretty clear in the SiderealRegridder code that the argument to setup should be a Telescope class (or ProductManager), and so I think it's fine. The place we need to change is really in ch_pipeline I think, so I'm changing this to approve.

jrs65 (approved):

jrs65:
@ssiegelx I've just rebased on master and force pushed the changes. Provided your validation is fine, I'm ready to merge this.

PR #109, opened 2020-10-15
SDSS stacking related changes

jrs65:
Here's a bundle of some stacking related changes:
- There's a decent speedup in the beamform code
- Most of the work is to be able to load the SDSS catalogs from the released FITS files and generating random catalogs by subsampling the official SDSS random catalogs.
- Some refactoring of the code which handles random number generation and seeding. I've deprecated the `mpi_random_seed` context manager in favour of a pipeline baseclass which does a better job of the same thing.

### DIFF
--- a/draco/core/io.py
+++ b/draco/core/io.py
@@ -45,6 +45,8 @@ from yaml import dump as yamldump
 from caput import pipeline
 from caput import config
 
+from cora.util import units
+
 from . import task
 from ..util.truncate import bit_truncate_weights, bit_truncate_fixed
 from .containers import SiderealStream, TimeStream, TrackBeam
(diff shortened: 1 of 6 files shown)


draco/core/io.py
```diff
@@ -204,6 +204,101 @@ def next(self):
...
+        if len(self.catalogs) == 0:
+            raise pipeline.PipelineStopIteration
+
+        group = self.catalogs.pop(0)
+
+        # Set the redshift selection
+        if self.freq_range:
+            zl = 1420.0 / self.freq_range[1] - 1
+            zh = 1420.0 / self.freq_range[0] - 1
+            self.z_range = (zl, zh)
```
ssiegelx, line 250: May want to replace 1420 with `cora.units.nu21` here for consistency with the beamform and stacking tasks.  This will shift range by ~1 chime frequency bin.

draco/synthesis/noise.py
```diff
@@ -98,19 +82,14 @@ def process(self, data):
...
-        with mpi_random_seed(self.seed, gen=rg) as rg:
-            noise = rg.normal(
-                scale=np.sqrt(tools.invert_no_zero(data.weight[:]) / 2),
-                size=(2,) + data.weight[:].shape,
-            )
-            vis[:] = noise[0] + 1j * noise[1]
+        # Replace visibilities with noise
+        vis = data.vis[:]
+        random.complex_normal(
+            scale=tools.invert_no_zero(data.weight[:]), out=vis, rng=self.rng
```
ssiegelx, line 89:
```suggestion
            scale=np.sqrt(tools.invert_no_zero(data.weight[:])), out=vis, rng=self.rng
```
I think there should be a `sqrt` here since `complex_normal` takes the standard deviation.

draco/util/random.py
```diff
@@ -0,0 +1,330 @@
...
+):
+    """Get a set of complex normal variables.
+
+        By default generate standard complex normal variables.
+
+        Parameters
+        ----------
+        size : tuple
+            Shape of the array of variables.
+        loc : float, optional
```
ssiegelx, line 57: May want to indicate in doc string that `loc` and `scale` can be any `np.ndarray` that broadcasts with an array of shape `size`, as is being done in the `GaussianNoise` and `GaussianNoiseDataset` task.

draco/util/random.py
```diff
@@ -0,0 +1,330 @@
...
+    # Fill the complex array by creating a real type view of it
+    rtype = _type_map[dtype]
+    rsize = size[:-1] + (size[-1] * 2,)
+    rng.standard_normal(rsize, out=out.view(rtype))
+
+    # Use inplace ops for scaling and adding to avoid intermediate arrays
+    rscale = scale / 2 ** 0.5
+    out *= rscale
+
+    if loc != 0.0:
```
ssiegelx, line 119: I think this will cause an error if `loc` is an `np.ndarray`.
```suggestion
    if np.any(loc):
```

ssiegelx (changes requested):

draco/core/io.py
jrs65: Fair point.

draco/synthesis/noise.py
jrs65: Oooh good catch. Mea cupla.

draco/util/random.py
```diff
@@ -0,0 +1,332 @@
...
+    # Fill the complex array by creating a real type view of it
+    rtype = _type_map[dtype]
+    rsize = size[:-1] + (size[-1] * 2,)
+    rng.standard_normal(rsize, out=out.view(rtype))
+
+    # Use inplace ops for scaling and adding to avoid intermediate arrays
+    rscale = scale / 2 ** 0.5
+    out *= rscale
+
+    if np.all(loc != 0.0):
```
ssiegelx, line 121: I think this will have unintended results in some case.  For example, when `loc` is an array with mostly non-zero values but a few zeros.

I think there is no harm in adding `loc` to `out` if it is zero, and the only reason to check this is to prevent an unnecessary addition?  I would suggest removing the if statement entirely or checking `if np.any(loc)`.

draco/core/io.py
```diff
@@ -204,6 +206,101 @@ def next(self):
...
+
+        # Exit this task if we have eaten all the file groups
+        if len(self.catalogs) == 0:
+            raise pipeline.PipelineStopIteration
+
+        group = self.catalogs.pop(0)
+
+        # Set the redshift selection
+        if self.freq_range:
+            zl = units.nu_21cm / self.freq_range[1] - 1
```
ssiegelx, line 250: I think the constant is named `nu21` not `nu_21cm`.

ssiegelx (approved):
I think there are issues with the fixes for two of my previous comments, but otherwise looks great.

draco/util/random.py
jrs65: Sorry. Should have been `.any()` in there. It is indeed to prevent unnecessary additions. I should stop trying to write code while on telecons apparently.

jrs65: Although that would be hard. I'm at 5 1/2 hours of calls today :-(

jrs65:
@fandinomat I'm going to merge this now anyway to make the `rev_02` prep easier. If you do have some time to look through the changes please do, I can always fix things up afterwards, but there's no hurry.

PR #110, opened 2020-10-15
Delay filter/estimator fixes and BlendStack task

jrs65:
These build ontop of the work that's in #109, so once that's merged we'll need to rebase this PR.

The biggest thing in here is to introduce a new task which requires a stack and then for each input stream it takes a weighted average with a small amount of the stack to fill in holes. This reduces the artifacts in the estimated delay spectrum quite by around an order of magnitude on the data I've used for testing.

### DIFF
--- a/draco/analysis/delay.py
+++ b/draco/analysis/delay.py
@@ -3,19 +3,12 @@
 
 import numpy as np
 import scipy.linalg as la
-import scipy.stats as st
-
-try:
-    # For backwards compatibility
-    from randomgen import RandomGenerator
-except ImportError:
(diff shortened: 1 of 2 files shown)


ssiegelx (approved):
Looks great.  I've suggested a few additional modifications to`null_delay_filter` that I had to make when comparing to the dayenu filter.

jrs65:
@ssiegelx I think this is basically done. But could you give the two new commits I added a once over.

ssiegelx (approved):
They look good, thanks!

PR #113, opened 2020-10-29
feat(MakeSiderealDayStream): a new task to simulate a specific LSD

jrs65:


draco/synthesis/stream.py
```diff
@@ -352,3 +352,73 @@ def _next_time_axis(self):
         self._cur_time += nsamp * int_time
 
         return time
+
+
+class MakeSiderealDayStream(task.SingleTask):
```
cahofer, line 357: just wondering if there is a better name for this task... Maybe just SiderealDayStream or SiderealStreamLSD. But that's up to you Richard.

cahofer (approved):
Did you test the task? I don't see anything obvious that needs change. The f-string syntax is new to me. Is that new python 3 standard?

draco/synthesis/stream.py
jrs65: Yeah. I couldn't think of a better one. This is vaguely in line with our conventions for other names. At somepoint we might go through all the task names and try and impose some order on them.

jrs65:
I have indeed tested it. Thanks for the review!!

PR #114, opened 2020-11-13
fix(beamform): incorrect normalization for collapse_ha

ssiegelx:
- The normalization of the beamformed visibility was missing
a factor of the primary beam when collapsing over hour angle.

- Simplify the calculation of the uncertainty on the
beamformed visibility when using the inverse variance
weighting scheme.

### DIFF
--- a/draco/analysis/beamform.py
+++ b/draco/analysis/beamform.py
@@ -246,7 +246,7 @@ class BeamFormBase(task.SingleTask):
                     # this_formed_beam was never normalized (this avoids
                     # re-work and makes code more efficient).
                     this_sumweight = np.sum(
-                        np.sum(sumweight_inrange, axis=-1) * primary_beam, axis=1
+                        np.sum(sumweight_inrange, axis=-1) * primary_beam ** 2, axis=1
                     )
 
                     formed_beam_full[pol] = np.sum(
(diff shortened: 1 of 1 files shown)


ssiegelx:
I was finding a 25% reduction in the amplitude of the signal when stacking on injected catalogs (prior to filtering), for example: [sourcestack_eBOSS_QSO_NGC_600_700MHz_intercyl_10min_invvar_data.pdf](https://github.com/radiocosmology/draco/files/5534542/sourcestack_eBOSS_QSO_NGC_600_700MHz_intercyl_10min_invvar_data.pdf)

I tracked this down to a missing factor of the primary beam in the normalization of the beamformed visibilities.  Here is the math describing it:  [beamformer_equation.pdf](https://github.com/radiocosmology/draco/files/5534556/beamformer_equation.pdf)

After fixing this bug, I recover the input amplitude of the injected catalogs.

This error effects both the signal and noise, so it will not change the result of the stacking analyses.

It will impact the absolute flux of the source spectra, but because we are only integrating over a very narrow region around the peak where the beam is close to 1.0, I suspect the magnitude of the effect is small.

fandinomat (approved):
Good point. Thanks for tracking this down.

jrs65 (approved):

PR #115, opened 2020-12-10
feat(io): raise ConfigError if file not found

nritsche:
Depends on https://github.com/radiocosmology/caput/pull/147

### DIFF
--- a/draco/core/io.py
+++ b/draco/core/io.py
@@ -182,6 +182,10 @@ def _list_of_filegroups(groups):
             files = group["files"]
         except KeyError:
             raise ConfigError("File group is missing key 'files'.")
+        except TypeError:
+            raise ConfigError(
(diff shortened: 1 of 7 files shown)


draco/core/io.py
```diff
@@ -75,40 +76,98 @@
...
+    Process a file group/groups
+
+    Parameters
+    ----------
+    groups : List[Dict] or Dict
+        Dicts should contain keys 'files': An iterable with file path or glob pattern strings, 'tag': the group tag str
+
+    Returns
+    -------
+    List[Dict]
```
jrs65, line 163: If you're going to give the types in this format, you might as well make them actual type annotations, and drop them from the docstring. Sphinx should be able to pull them out as expected.

jrs65 (changes requested):

jrs65 (approved):

PR #116, opened 2021-01-05
docs build / readthedocs

nritsche:
I wasn't able to make this work on readthedocs because it needs mpi4py to build the docs. So I put it on github pages instead: https://radiocosmology.github.io/draco

This repository now has a branch `gh-pages` that is used only to host the automated shinx build.

The commit history is a mess because I tried around first with rtd and then with gh-pages. I will squash it into:
```
docs(sphinx): fix docs build and host on github pages
```


nritsche:
By mistake I made these changes on top of rn/configtest. I think merging that first will resolve things

draco/analysis/mapmaker.py
```diff
@@ -194,7 +182,7 @@ class MaximumLikelihoodMapMaker(BaseMapMaker):
     The dirty map is produced by generating a set of :math:`a_{lm}` coefficients
     using
 
-    .. math:: \hat{\mathbf{a}} = \left( \mathbf{N}^{-1/2 }\mathbf{B} \right)^+ \mathbf{N}^{-1/2} \mathbf{v}
+    .. math:: \hat{\mathbf{a}} = ( \mathbf{N}^{-1/2 }\mathbf{B} ) ^+ \mathbf{N}^{-1/2} \mathbf{v}
```
nritsche, line 185: @jrs65 I couldn't find out what `\left` and `\right` did, but they made this break with an error like "Expected an empty line, unexpected unindent" or similar

nritsche: oh, I see: it ensures the brackets are big enough to encapsulate the entire block inside. Let's see what comes out of this. Apart from the warning, it definitely produces garbage *with* the `\right`,`\left`

jrs65: It could well be that the `\` are getting interpreted as escape codes in the string. Try changing the string to be a raw string, i.e. start it with `r"""`

nritsche: thanks, that worked! (and the brackets really are a tiny bit bigger now :facepalm: )

nritsche:
I wasn't able to make this work on readthedocs because it needs mpi4py to build the docs. So I put it on github pages instead: https://radiocosmology.github.io/draco

This repository now has a branch `gh-pages` that is used only to host the automated shinx build.

jrs65 (approved):

PR #118, opened 2021-01-12
docs(sphinx): add categorized lists again as simple references

nritsche:


jrs65 (approved):
Other than a general comment about whether you need the `:py:xxxx:` prefixes this looks good. Thanks for going back and sorting all of these out, I very much appreciate it.

jrs65 (approved):

PR #120, opened 2021-01-13
Deprecate caput.pipeline.PipelineConfigError

nritsche:
see https://github.com/radiocosmology/caput/pull/156

I don't think these should ever have been config errors, but please correct me if I'm wrong.

### DIFF
--- a/draco/core/io.py
+++ b/draco/core/io.py
@@ -739,9 +739,9 @@ class Truncate(task.SingleTask):
                 or self.fixed_precision is None
                 or self.variance_increase is None
             ):
-                raise pipeline.PipelineConfigError(
+                raise config.CaputConfigError(
                     "Container {} has no preset values. You must define all of 'dataset', "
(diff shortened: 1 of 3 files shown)


tristpinsm:
In the first case it was made a `ConfigError` because the error results from feeding a container to this task without specifying the config parameters `fixed_precision` or `variance_increase`. (if you encounter this error you fix it in the config)

In the second case, it's not clear to me either of those exceptions are really appropriate since it is raised due to a problem with the definition of the task method `process`.

draco/core/io.py
```diff
@@ -739,7 +739,7 @@ def _get_params(self, container):
                 or self.fixed_precision is None
                 or self.variance_increase is None
             ):
-                raise pipeline.PipelineConfigError(
+                raise pipeline.PipelineRuntimeError(
```
tristpinsm, line 742: This should remain a config error.

tristpinsm (changes requested):

tristpinsm (approved):

PR #122, opened 2021-02-03
Deconvolving ringmap

jrs65:


wulfda02:
Currently, you have the "natural" EW weights as [4,3,2,1], but I think it should be [2,3,2,1].  The 0-cylinder separation baselines don't include any independent negative v baselines, unlike the non-zero cylinder separations (thus reducing their weight by half). Right? 
```
else:  # self.weight_ew == "natural"
            weight_ew = n_ew - np.arange(n_ew)
```

wulfda02:
Also, when forming the wiener filter, you are summing over the cylinder axis, as well as the +/-m axis in the denominator
```
C_inv = self.inv_SN + (mB.conj() * weight_ew[:, np.newaxis] * mB).sum(
                    axis=(1, -2)
                )
```
But, shouldn't we keep positive and negative m separate until after filtering? I.e. only sum over axis=-2

jrs65:
> Currently, you have the "natural" EW weights as [4,3,2,1], but I think it should be [2,3,2,1]. The 0-cylinder separation baselines don't include any independent negative v baselines, unlike the non-zero cylinder separations (thus reducing their weight by half). Right?
> 
> ```
> else:  # self.weight_ew == "natural"
>             weight_ew = n_ew - np.arange(n_ew)
> ```

This is a good question. I am pretty sure it's correct, but it's not trivial to come up with a concrete explanation as to why.

Here are a few different stabs at it:

- With reference to how the standard ringmap maker works, the non-zero positive cylinder separations get implicitly conjugated into the negative separations and so by your argument you need to halve the weighting of all of them. This is relevant for the deconvolved ringmap maker because the standard ringmap maker is essentially the limit as you take the primary beam width to zero.
- Second, as you point out there aren't independent negative baselines, so in some sense you should get half the weighting, but because of that same property there zero cylinder separation beams are purely real and thus have no noise in an imaginary part, so they have half the variance (i.e. twice the weight). These basically cancel out. This is one of the subtleties of interferometry, for very short baselines the noise properties are very different for real and imaginary parts. Interestingly this slightly odd effect disappears when using m-modes (in the general sense, I'm not sure about this specific hybrid case atm).

jrs65:
> Also, when forming the wiener filter, you are summing over the cylinder axis, as well as the +/-m axis in the denominator
> 
> ```
> C_inv = self.inv_SN + (mB.conj() * weight_ew[:, np.newaxis] * mB).sum(
>                     axis=(1, -2)
>                 )
> ```
> 
> But, shouldn't we keep positive and negative m separate until after filtering? I.e. only sum over axis=-2

In order to guarantee the output sky is real, you need to treat the positive/negative m's as separate measurements of the same underlying mode and only infer the m >= 0 modes (as the negative ones on the sky are simple conjugates). This is done in the m-mode papers, although I have a similar derivation of it for the hybrid case that I can tidy up. Ultimately it means that the +/- modes are independent measurements at a similar level to the different cylinders and thus you need to sum over both in the filter.

Ultimately I think the fact that the +/- modes are mostly disjoint (coming from above/below the NCP), means that this doesn't really matter a huge amount in the practical output for CHIME. However, any instrument that has an EW baseline length > 0, but < aperture width (i.e. a responses that covers both +/-m in some asymmetric manner) would need to account for this properly to get the right S/N weighting. CHIME gets away with it because either your baseline has + or - m response at a particular declination (but not both), or it is the 0-cyl separation in which case the response is symmetric for +/- m and they should get weighted equally.

jrs65:
I'm having doubts about the relative cylinder weighting point. I think I'm going to need to write it all out to be sure. I'll circulate it when I've got something written up.

ssiegelx:
I'm finding that the maps generated with this task are scaled down by a factor of 4 relative to maps created with the standard ringmap maker.  Also the flux of point sources is off by a factor of four from their expected value. 

This is also true of the deconvolving map maker task that I wrote that uses an external beam model.

[example_normalization_problem_3c295.pdf](https://github.com/radiocosmology/draco/files/6172792/example_normalization_problem_3c295.pdf)

This plot shows an RA slice through the map at the (RA, Dec) of 3C 295.  I'm showing the standard ring map maker in blue (i.e., using the BeamformEW task on the same HybridVis container), this task in red, and my map maker in green.  The expected flux of the source is indicated by the black dashed line.

@wulfda02 Have you encountered this?

wulfda02:
As I mentioned on the stacking call, I think this can be explained by the dirty beam amplitude after going through the wiener filter.  It sounds like there's still some uncertainty as to _why_ this amplitude isn't 1, but in my work I find that re-normalizing by this amplitude produces the correct flux for point sources.

ssiegelx:
See PR #126 for an extension to this that allows for an external beam model.

jrs65:
@ssiegelx @wulfda02 as this doesn't effect any existing functionality (and you have your own updates to), I'd like to get this very quickly reviewed and merged into master, and then we can work on getting your changes merged on top.

draco/analysis/ringmapmaker.py
```diff
@@ -230,7 +231,7 @@ def process(self, gstream):
                 vmax = nsmax / wv
 
             if self.weight == "inverse_variance":
-                gw = gsw[:, lfi].copy()
+                gw = gsw[:, lfi].view(np.ndarray).copy()
```
anjakefala, line 234: Should the `view()` occur before the slice?

draco/analysis/transform.py
```diff
@@ -485,62 +484,71 @@ def process(self, sstream):
         if self.telescope is not None:
             mmax = self.telescope.mmax
         else:
-            mmax = None
-
-        # Construct the array of m-modes
-        marray = _make_marray(sstream.vis[:], mmax)
-        marray = mpiarray.MPIArray.wrap(marray[:], axis=freq_axis, comm=sstream.comm)
+            mmax = (sstream.vis.shape[-1] // 2) - 1
```
ssiegelx, line 487: I think this needs to be:
```suggestion
            mmax = sstream.vis.shape[-1] // 2
```

Otherwise the round-trip m-mode transform:
```
    - type: draco.analysis.transform.MModeTransform
      in: sstream
      out: mmode

    - type: draco.analysis.transform.MModeInverseTransform
      in: mmode
      out: rt_sstream
```
results in an RA axis with size N_{RA} -1 when N_{RA} is even and N_{RA} - 2 when N_{RA} is odd.

ssiegelx (approved):
@jrs65 I think I've identified one bug described in the comment above.

Beyond that, all of my suggestions are implemented in PR #126.

draco/analysis/transform.py
jrs65: Thanks! I'll fix that one.

draco/analysis/ringmapmaker.py
```diff
@@ -386,7 +387,166 @@ def process(self, hstream):
 class RingMapMaker(task.group_tasks(MakeVisGrid, BeamformNS, BeamformEW)):
     """Make a ringmap from the data."""
 
-    pass
+
+class DeconvolveHybridM(task.SingleTask):
+
```
anjakefala, line 392: Do we usually have class level documentation for Tasks?

draco/analysis/ringmapmaker.py
anjakefala: I believe that both `.local_array` and moving the view before the slice is old-code compatible!

draco/analysis/ringmapmaker.py
```diff
@@ -386,7 +387,166 @@ def process(self, hstream):
...
+            )
+
+            # TODO: the handling of the polarisations in here is probably very messed
+            # up.
+            for pi, (pa, pb) in enumerate(pol_map):
+
+                # Get the effective beamwidth for the polarisation combination
+                sig_a = beam_width[pa](freq, dec)
+                sig_b = beam_width[pb](freq, dec)
+                sig = sig_a * sig_b / (sig_a ** 2 + sig_b ** 2) ** 0.5
```
anjakefala, line 515: What is the order of operations that you expect here?

jrs65: ??

PR #123, opened 2021-02-08
perf(SingleTask): remove no-effect-__getitem__ call

nritsche:
I'm not sure if this will speed up `_nan_check_walk`, but I don't see how this line can have an effect anyways.

Percentages are of the total run time of the daily processing pipeline:
- 10% _nan_check_walk (draco.core.task)
  - 7% __getitem__ (memh5)

### DIFF
--- a/draco/core/task.py
+++ b/draco/core/task.py
@@ -435,7 +435,6 @@ class SingleTask(MPILoggedTask, pipeline.BasicContMixin):
                 try:
                     is_nan = np.isnan(arr)
                     is_inf = np.isinf(arr)
-                    arr = n[:]
                 except TypeError:
                     continue
 


jrs65 (changes requested):
This is slightly more subtle than this. `arr` is used on L438 and L456, but you can't use `n` directly there because those conditionals don't run on every rank, and thus it will hang because of the MPI call hidden in the `[:]`

jrs65:
It's an excellent point that we should remove that MPI call when there isn't an error though. I think you can probably pull out the array directly, by accessing `n._data` or similar which won't involve the MPI call.

nritsche:
> This is slightly more subtle than this. `arr` is used on L438 and L456, but you can't use `n` directly there because those conditionals don't run on every rank, and thus it will hang because of the MPI call hidden in the `[:]`

I think you missed L434

jrs65:
Ha. I did indeed. Thanks!!

jrs65 (approved):

nritsche:
> It's an excellent point that we should remove that MPI call when there isn't an error though. I think you can probably pull out the array directly, by accessing `n._data` or similar which won't involve the MPI call.

Does this still apply?

jrs65:
> > It's an excellent point that we should remove that MPI call when there isn't an error though. I think you can probably pull out the array directly, by accessing `n._data` or similar which won't involve the MPI call.
> 
> Does this still apply?

I was just reading the code of MPIArray to tell. I *think* it does. It looks like even `n[:]` generates an MPI call (via `MPIArray.wrap`) which isn't needed for the stuff below.

Ultimately I think there's an optimisation that should maybe go in caput, as a  `distributed_dataset[:]` call should be able to create a shallow clone of the underlying MPIArray without invoking any MPI calls.

PR #127, opened 2021-03-25
ss/common mode beam

ssiegelx:
feat(containers): create CommonModeGridBeam and CommonModeTrackBeam

Similar to the existing GridBeam and TrackBeam containers but without
an input axis.  The (freq, pol) axes also transposed to match the order of
the HybridVisStream container for easy application.

This commit makes GridBeam and TrackBeam subclasses of the
CommonMode- versions.

### DIFF
--- a/draco/core/containers.py
+++ b/draco/core/containers.py
@@ -1031,12 +1031,84 @@ class TimeStream(FreqContainer, VisContainer, TODContainer):
         return self.datasets["input_flags"]
 
 
-class GridBeam(FreqContainer):
-    """Generic container for representing the 2-d beam in spherical
-    coordinates on a rectangular grid.
-    """
+class CommonModeGridBeam(FreqContainer):
+    """Generic container for representing a 2D beam on a rectangular grid."""
 
-    _axes = ("pol", "input", "theta", "phi")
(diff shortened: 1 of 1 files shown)


tristpinsm (approved):
It's a little confusing that the `freq` and `pol` axis order is swapped between the common-mode and the other containers, but I guess this is to conform with some other convention, and it's not a big deal.

tristpinsm:
Oh, and actually LGTM caught a small mistake that should be fixed before merging

ssiegelx:
Thanks @tristpinsm.  I went back and forth on this and in the end decided to use the existing containers with a length 1 input axis to hold the baseline-averaged beams.  This will be a small nuisance for people accessing the beams, but makes it easier to write tasks that can be applied to both the baseline-averaged beams and the input-dependent beams, and also intermediate cases (for example, a different beam for each cylinder pair).

So now the only changes in this PR involve adding an optional `quality` dataset to the `GridBeam` container and updating some of the doc strings.

tristpinsm (approved):
OK, that sounds like a good decision to me. If the need for a quality dataset arises with other containers, we can add it then. This should be squash-merged into master to clean up the commit history.

PR #129, opened 2021-04-20
feat(sourcestack): separately stack each pol

ssiegelx:
Creates FrequencyStackByPol container, which is a
FrequencyStack container with a polarisation axis.
Updates SourceStack to use the FrequencyStackByPol
container and stack each polarisation separately if
the input  FormedBeam container has more than
one polarisation.

### DIFF
--- a/draco/analysis/sourcestack.py
+++ b/draco/analysis/sourcestack.py
@@ -58,6 +58,10 @@ class SourceStack(task.SingleTask):
         freq = formed_beam.freq
         nfreq = len(freq)
 
+        # Polarisation axis
+        pol = formed_beam.pol
+        npol = len(pol)
+
         # Frequency of quasars
         qso_freq = NU21 / (formed_beam["redshift"]["z"] + 1.0)  # MHz.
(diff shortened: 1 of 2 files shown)


ssiegelx:
@fandinomat I have found it helpful to retain the polarisation axis when stacking over sources.  This PR makes that possible by creating a new `FrequencyStackByPol` container that has a polarisation axis, and having `SourceStack` use it if the beamformer has not already averaged down to total intensity.  If the beamformer has already averaged down to total intensity, then `SourceStack` will use the old `FrequencyStack` container.

I think it would simplify things if, instead of what is done in this PR, we updated the `FrequencyStack` container so that it always has a polarisation axis (which would be length 1 for total intensity).  However I realize that this might cause issues with old data products.  If you think this would indeed cause you problems, then we can go ahead with the solution in this PR (assuming everything else looks good).

I'm also tagging @jrs65 in case he has an opinion on this.

jrs65 (approved):
Looks good. I just added a docstring for the new container.

PR #130, opened 2021-05-06
feat(task): zarr support

nritsche:
Depends on changes in https://github.com/radiocosmology/caput/pull/169

### DIFF
--- a/draco/core/containers.py
+++ b/draco/core/containers.py
@@ -142,7 +142,7 @@ class ContainerBase(memh5.BasicCont):
         skip_datasets = kwargs.pop("skip_datasets", False)
         dist = kwargs.pop("distributed", True)
         comm = kwargs.pop("comm", None)
-        self.allow_chunked = kwargs.pop("allow_chunked", False)
+        self.allow_chunked = kwargs.pop("allow_chunked", True)
 
         # Run base initialiser
         super().__init__(distributed=dist, comm=comm)
@@ -280,6 +280,36 @@ class ContainerBase(memh5.BasicCont):
         dset.attrs["axis"] = np.array(axes)
 
         return dset
+
+    def _ensure_chunked(self):
+        """Ensure datasets that have chunk/compression specs are chunked.
+
+        For every dataset, check if chunks and compression are set, and
+        if not set them to dataset_spec values.
+        """
+        for dset in self.dataset_spec:
+            if dset not in self:
+                continue
+            if "chunks" in self.dataset_spec[dset] and self[dset].chunks is None:
+                # ensure chunks aren't larger than dataset shape
+                chunks = ()
+                for i, l in enumerate(self[dset].shape):
+                    chunks += (min(self.dataset_spec[dset]["chunks"][i], l),)
+                self._data._storage_root[dset].chunks = chunks
+            if (
+                "compression" in self.dataset_spec[dset]
+                and self[dset].compression is None
+            ):
+                self._data._storage_root[dset].compression = self.dataset_spec[dset][
+                    "compression"
+                ]
+            if (
+                "compression_opts" in self.dataset_spec[dset]
+                and self[dset].compression_opts is None
+            ):
+                self._data._storage_root[dset].compression_opts = self.dataset_spec[
+                    dset
+                ]["compression_opts"]
 
     @property
     def datasets(self):
@@ -455,6 +485,10 @@ class ContainerBase(memh5.BasicCont):
                 # Copy over the data and attributes
                 dset[:] = data[:]
                 memh5.copyattrs(data.attrs, dset.attrs)
+                # TODO Is there a case where these properties don't exist?
+                data.chunks = dset.chunks
+                data.compression = dset.compression
+                data.compression_opts = dset.compression_opts
 
         return new_cont
 
@@ -1064,7 +1098,10 @@ class SiderealStream(
             "distributed_axis": "freq",
             "compression": COMPRESSION,
             "compression_opts": COMPRESSION_OPTS,
-            "chunks": (64, 256, 128),
+            "chunks": (64, 128, 128),
+            "truncate": {
+                "weight_dataset": "vis_weight",
+            },
         },
         "vis_weight": {
             "axes": ["freq", "stack", "ra"],
@@ -1074,7 +1111,8 @@ class SiderealStream(
             "distributed_axis": "freq",
             "compression": COMPRESSION,
             "compression_opts": COMPRESSION_OPTS,
-            "chunks": (64, 256, 128),
+            "chunks": (64, 128, 128),
+            "truncate": True,
         },
         "input_flags": {
             "axes": ["input", "ra"],
@@ -1097,7 +1135,8 @@ class SiderealStream(
             "distributed_axis": "freq",
             "compression": COMPRESSION,
             "compression_opts": COMPRESSION_OPTS,
-            "chunks": (3, 64, 256, 128),
+            "chunks": (3, 64, 128, 128),
+            "truncate": True,
         },
         "nsample": {
             "axes": ["freq", "stack", "ra"],
@@ -1107,7 +1146,7 @@ class SiderealStream(
             "distributed_axis": "freq",
             "compression": COMPRESSION,
             "compression_opts": COMPRESSION_OPTS,
-            "chunks": (64, 256, 128),
+            "chunks": (64, 128, 128),
         },
     }
 
@@ -1245,7 +1284,10 @@ class TimeStream(FreqContainer, VisContainer, TODContainer):
             "distributed_axis": "freq",
             "compression": COMPRESSION,
             "compression_opts": COMPRESSION_OPTS,
-            "chunks": (64, 256, 128),
+            "chunks": (64, 128, 128),
+            "truncate": {
+                "weight_dataset": "vis_weight",
+            },
         },
         "vis_weight": {
             "axes": ["freq", "stack", "time"],
@@ -1255,7 +1297,8 @@ class TimeStream(FreqContainer, VisContainer, TODContainer):
             "distributed_axis": "freq",
             "compression": COMPRESSION,
             "compression_opts": COMPRESSION_OPTS,
-            "chunks": (64, 256, 128),
+            "chunks": (64, 128, 128),
+            "truncate": True,
         },
         "input_flags": {
             "axes": ["input", "time"],
@@ -1441,7 +1484,10 @@ class TrackBeam(FreqContainer, SampleVarianceContainer):
             "distributed_axis": "freq",
             "compression": COMPRESSION,
             "compression_opts": COMPRESSION_OPTS,
-            "chunks": (128, 2, 128, 128),
+            "chunks": (64, 2, 64, 128),
+            "truncate": {
+                "weight_dataset": "weight",
+            },
         },
         "weight": {
             "axes": ["freq", "pol", "input", "pix"],
@@ -1451,7 +1497,8 @@ class TrackBeam(FreqContainer, SampleVarianceContainer):
             "distributed_axis": "freq",
             "compression": COMPRESSION,
             "compression_opts": COMPRESSION_OPTS,
-            "chunks": (128, 2, 128, 128),
+            "chunks": (64, 2, 64, 128),
+            "truncate": True,
         },
         "sample_variance": {
             "axes": ["component", "freq", "pol", "input", "pix"],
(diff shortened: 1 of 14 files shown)


draco/core/containers.py
```diff
@@ -272,6 +282,41 @@ def add_dataset(self, name):
 
         return dset
 
+    def chunkify(self):
```
jrs65, line 285: This should probably have a more sensible name: `ensured_chunked` would be more informative.

draco/core/io.py
```diff
@@ -713,42 +690,69 @@ class Truncate(task.SingleTask):
...
+            Dataset name
+
+        Returns
+        -------
+        Dict or None
+            Returns `None` if the dataset shouldn't get truncated.
+        """
+        # Check if dataset should get truncated at all
+        if self.dataset is None:
+            if dset not in container._dataset_spec or not container._dataset_spec[
```
jrs65, line 728: I wonder if you want to pull the dataset spec from the class rather than the instance, i.e. `container.__class__._dataset_spec`

draco/core/task.py
```diff
@@ -389,8 +393,20 @@ def _save_output(self, output):
             outfile = os.path.expanduser(outfile)
             outfile = os.path.expandvars(outfile)
 
+            if self.output_chunking:
+                try:
+                    output.chunkify()
```
jrs65, line 398: As discussed, I'm not sure if we want this one. I'd like to know if it really is necessary first.

draco/core/containers.py
```diff
@@ -907,7 +952,10 @@ class SiderealStream(FreqContainer, VisContainer, SiderealContainer):
             "distributed_axis": "freq",
             "compression": COMPRESSION,
             "compression_opts": COMPRESSION_OPTS,
-            "chunks": (64, 256, 128),
+            "chunks": (128, 256, 512),
```
jrs65, line 955: I think broadly on the chunk sizes we should not bother chunking/compressing small files (~100 MB or less), as it's probably not worth the overhead. That's things like RFIMask, SystemSensitivity, etc. We should probably keep writing those out as HDF5 too, though I guess that's a pipeline config thing.

jrs65 (changes requested):

draco/core/io.py
nritsche: Why?

draco/core/task.py
nritsche: I ran it without and it's not needed :+1:

draco/analysis/beamform.py
```diff
@@ -607,6 +608,11 @@ def _process_catalog(self, catalog):
         if "position" not in catalog:
             raise ValueError("Input is missing a position table.")
 
+        if not hasattr(self, "epoch"):
+            raise PipelineRuntimeError(
+                "Epoch not set. Was the requested data not available?"
+            )
```
anjakefala, line 614: https://github.com/radiocosmology/draco/issues/133

anjakefala: We are gonna swap this `raise` for a `warning` log.

draco/analysis/beamform.py
```diff
@@ -607,6 +609,10 @@ def _process_catalog(self, catalog):
         if "position" not in catalog:
             raise ValueError("Input is missing a position table.")
 
+        if not hasattr(self, "epoch"):
+            self.log.warning("Epoch not set. Was the requested data not available?")
+            self.data_available = False
```
anjakefala, line 614: We need to return here.

anjakefala: Should probably do the check in `process_data`, bc that is where "no data being available" is discovered.

And return non-zero if no data available, or something.

anjakefala: We are going to stick with the current change because of this bug: https://github.com/radiocosmology/caput/issues/181#issuecomment-924398276

For `BeamFormCat`, `_process_data` in `setup()`, and so it will not be called, if there is no data available.

anjakefala:
@tristpinsm As far as I was concerned, it is ready for review!

draco/core/task.py
```diff
@@ -390,7 +393,13 @@ def _save_output(self, output):
             outfile = os.path.expandvars(outfile)
 
             self.log.debug("Writing output %s to disk.", outfile)
-            self.write_output(outfile, output)
+            self.write_output(
```
tristpinsm, line 396: I don't understand why compression parameters are now held at the `memh5.MemGroup` level. Is it just so that they can be specified in a pipeline config `save` block? Maybe there is a good reason for this to do with Zarr, but it strikes me as inconsistent with the way the compression settings are implemented in `draco` containers and the previous version of `memh5`, where compression is set for every dataset individually. I'll have to take a closer look at the `caput` PR, but in any case it strikes me at least as redundant and maybe conflicting to be specifying compression in multiple places.

tristpinsm (changes requested):
Overall this is looking great. I think the way chunking is defined per-container could be improved (see comments) and I'm confused about how the compression has been refactored.

draco/core/containers.py
tristpinsm: I left some comments addressing this point in my review. More generally, I noted that this PR enables chunking by default for all datasets. Is their a benefit to doing this for datasets that we don't intend to compress? With it on by default we would need to specify those datasets where we don't want chunks. The way it was before you would specify only the datasets where you wanted chunking and compression. I guess which way we set the default should reflect how we intend the typical dataset to be written out.

anjakefala:
@tristpinsm Whose responsibility is it to respond to review comments, btw? Is it me?

It is fine if it is me, I just want to be explicit! This branch does not have a new owner.

tristpinsm:
> @tristpinsm Whose responsibility is it to respond to review comments, btw? Is it me?
> 
> It is fine if it is me, I just want to be explicit! This branch does not have a new owner.

I don't know, as far as I'm concerned anybody is welcome to contribute!

jrs65:
I think the best way forward is probably that we all review this (I think it's just me that hasn't done a last round), and then maybe we meet soon (tomorrow or Friday?) and figure out what changes are important, and we split up making them between us all.

tristpinsm:
> I think the best way forward is probably that we all review this (I think it's just me that hasn't done a last round), and then maybe we meet soon (tomorrow or Friday?) and figure out what changes are important, and we split up making them between us all.

Sounds good to me. I won't be available tomorrow but Friday works.

draco/analysis/beamform.py
jrs65: I guess this one will probably need some work to merge with the current version.

draco/core/io.py
jrs65: I'm not quite sure what I was getting at in here, but my best interpretation 6 months later is that it's debatable if this should get the class or the instance passed in to this method.

draco/core/io.py
```diff
@@ -713,42 +690,69 @@ class Truncate(task.SingleTask):
...
     variance_increase = config.Property(proptype=float, default=None)
+    ensure_chunked = config.Property(proptype=bool, default=True)
+
+    default_params = {
+        "weight_dataset": None,
+        "fixed_precision": 1e-4,
+        "variance_increase": 1e-3,
+    }
+
+    @functools.lru_cache()
```
jrs65, line 709: I think what's done in here is sufficiently quick it doesn't really warrant the LRU cache, but note that if we change this to a per instance call, then it'll never see the same pair twice and it should definitely be removed.

draco/core/task.py
```diff
@@ -262,6 +262,9 @@ class SingleTask(MPILoggedTask, pipeline.BasicContMixin):
 
     output_root = config.Property(default="", proptype=str)
     output_name = config.Property(default="{output_root}{tag}.h5", proptype=str)
+    output_format = config.file_format()
+    output_compression = config.Property(default=None, proptype=str)
+    output_compression_opts = config.Property(default=None)
```
jrs65, line 267: I can't remember if these are just inherited from `BasicContMixin`, so check that...

Also, similar to my comments in `caput.pipeline` it's not clear if we should allow these to be specified *per dataset*. Also is this the right format for these arguments if we do that?

draco/core/task.py
jrs65: Ah, yes. I just made a similar comment above. 100% agree that we should think about how we want to be able to specify these parameters (and if we should at all?).

jrs65 (changes requested):

draco/analysis/beamform.py
anjakefala: Sigh. Yeah.

draco/analysis/beamform.py
jrs65: We can remove the beamform.py changes @anjakefala

draco/analysis/beamform.py
jrs65: Same here.

draco/core/containers.py
```diff
@@ -1567,27 +1640,37 @@ class RingMap(FreqContainer, SiderealContainer):
             "initialise": True,
             "distributed": True,
             "distributed_axis": "freq",
+            "truncate": {
```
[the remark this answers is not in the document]
jrs65: We should review the sanity of these settings.

draco/core/io.py
jrs65: Changes this to be from instance, and remove the lru_cache.

jrs65:
Things to check:

- [ ] How exactly is chunking set by default? Should ensure_chunked set chunking for container types without chunking set? (so ensure chunking should not use default from axes).
- [ ] Prefix ensure chunked with underscore??
- [ ] Copy compression and chunking parameters over in the `.copy()` method.
- [ ] Change the compression options on tasks to be:
    - A single option, either a boolean or a dict (False, ignore all compression and chunks; True is default and use whatever is there)
    - Dict entries are dset_path -> dict of compression options
    - Compression options are: `{"compression": compression_on_off, "compression_opts": dict_of_options_like_h5py, "chunks": (size, of, chunks)}`
    - Unlisted datasets get default settings
    - Needs to pass through all the save/to_file methods
- [ ] Abstract distributed_group_to_*_parallel to give a single implementation for both HDF5 and zarr
- [ ] Add zarr to mkchimeenv.sh 
- [ ] Change ch_pipeline configs to account for everything we are doing above.

tristpinsm:
I was looking at implementing the compression changes, and I think it may be better to parse the config parameters and apply them to the dataset objects in `SingleTask._save_output`. Otherwise, combining the dataset properties and the config-provided `dict` needs to be replicated in every one of the many places datasets are created/written to in `memh5`. All of these places currently get the compression parameters from the dataset objects, and this seems like a pretty efficient way to do it.

The only downside to this approach is that the datasets on the container need to be modified and this will persist downstream in the pipeline, but given that we don't expect this to be a common use-case it might be ok. You could also restore the previous state to the datasets after writing them out I guess.

tristpinsm:
I've made I think the necessary changes to `draco` and `caput` to implement the new way of specifying compression options. I haven't tested my changes yet, and so haven't pushed them. Is there already a test that exists out there, or should I write a config to try it out?

anjakefala:
> I've made I think the necessary changes to draco and caput to implement the new way of specifying compression options. I haven't tested my changes yet, and so haven't pushed them. Is there already a test that exists out there, or should I write a config to try it out?

There exists a config to test with! I can run the tests.

tristpinsm:
this is the config I used for testing:
```pipeline:
  logging:
    root: DEBUG
    h5py: DEBUG
    peewee: INFO
    matplotlib: INFO

  tasks:

    - type: draco.core.task.SetMPILogging
      params:
        level_all: INFO
        level_rank0: DEBUG

    - type: draco.core.io.LoadFilesFromParams
      out: data
      params:
        files: "./test_cmp_data.h5"
          #        distributed: False

    - type: draco.core.io.Truncate
      in: data
      params:
        dataset:
          vis:
            weight_dataset: "vis_weight"
            variance_increase: 0.1 
        ensure_chunked: False
        save: True
        output_name: "./test_cmp_out.zarr"
        output_compression:
          vis:
            compression: "bitshuffle"
            chunks: [64, 128, 512]
          vis_weight:
            compression: "bitshuffle"
            chunks: [64, 128, 512]
```

draco/core/containers.py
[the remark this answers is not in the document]
jrs65: I will skim your changes.

jrs65 (approved):

jrs65:
@tristpinsm did you still have requested changes outstanding in here? Otherwise I'll get it merged.

tristpinsm:
> @tristpinsm did you still have requested changes outstanding in here? Otherwise I'll get it merged.

No, I think we made all the changes that we had discussed!

PR #131, opened 2021-05-10
feat(beamform): refactor how primary beam model is evaluated

ssiegelx:
- Add capability to track sources for a declination
dependent amount of time, so that all sources are
tracked through the same angle on the sky.

- Add option to ignore primary beam model when beamforming,
i.e., use uniform weighting.

- Use the beam method of the telescope object to evaluate
the primary beam model.  Requires [PR 78](https://github.com/chime-experiment/ch_pipeline/pull/78)
of ch_pipeline for the CHIME implementation.

- Move the parameterized fit to the driftscan beam out of
this module and into its own telescope class in ch_pipeline.

- Check if a catalog is already in CIRS coordinates.

- Add the catalog's tag to the output container's tag.

- Create versions of the beamforming tasks that use an
external beam model provided at setup in a GridBeam container.

### DIFF
--- a/draco/analysis/beamform.py
+++ b/draco/analysis/beamform.py
@@ -1,6 +1,34 @@
-"""Beam Forming"""
+"""
+===================================================
+Beamforming tasks (:mod:`~draco.analysis.beamform`)
+===================================================
 
+.. currentmodule:: draco.analysis.beamform
+
+Tools for beamforming visibilities to the location of known sources.
+
+Tasks
+=====
+
+.. autosummary::
+    :toctree: generated/
+
+    BeamForm
+    BeamFormCat
+    BeamFormExternal
+    BeamFormExternalCat
+
+Task Base Classes
+-----------------
+.. autosummary::
+    :toctree: generated/
+
+    BeamFormBase
+    BeamFormExternalBase
+
+"""
 import numpy as np
+import scipy.interpolate
 from skyfield.api import Star, Angle
 
 from caput import config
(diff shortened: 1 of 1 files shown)


ssiegelx:
The main change in this PR is that the beamformer now uses the `beam` method of the telescope object provided at setup to evaluate the primary beam model (instead of Mateus' parameterized fit to the driftscan cylinder beam that was used previously).  This will make it easier to swap in different beam models.  Note that this requires some minor changes to the call signature of the `beam` method that are implemented for CHIME in [PR 78](https://github.com/chime-experiment/ch_pipeline/pull/78) of `ch_pipeline`.  The code that evaluates Mateus' parameterized fit to the driftscan cylinder beam has been moved to `ch_pipeline` (also PR 78) since it is CHIME specific.

In addition, there are two new tasks, `BeamFormExternal` and `BeamFormExternalCat`, which will evaluate the primary beam by interpolating a `GridBeam` container that has been loaded outside the task and provided at setup.

sjforeman (approved):

draco/analysis/beamform.py
```diff
@@ -698,6 +724,173 @@ def process(self, source_cat):
...
+
+        else:
+            raise ValueError(f"Do not recognize beam container: {beam.__class__}")
+
+    def _initialize_beam_with_data(self):
+        """Ensure that the beam and visibilities have the same frequency axis."""
+
+        freq = self.freq_local
+
+        if (freq.size != self._beam_freq.size) or np.any(freq != self._beam_freq):
```
jrs65, line 768:
```suggestion
        if not np.array_equal(freq, self._beam_freq):
```

draco/analysis/beamform.py
```diff
@@ -414,79 +432,78 @@ def _ha_array(self, ra, source_ra_index, source_ra, ha_side, is_sstream=True):
...
-            )
-        else:
-            # XY or YX
-            return (
-                _amp(0, dec, zenith)
-                * np.exp(-(((ha - ha0) / _sig(0, freq, dec)) ** 2))
-                * _amp(1, dec, zenith)
-                * np.exp(-(((ha - ha0) / _sig(1, freq, dec)) ** 2))
-            ) ** 0.5
+            primary_beam[ff] = np.sum(bii * np.dot(bjj.conjugate(), p_stokesI), axis=1)
```
jrs65, line 496: As `pstokes_I` is the identity here can we just skip the `np.dot`?

jrs65 (approved):
This looks like a good refactoring to me. A few minor comments.

ssiegelx:
Thanks @jrs65 and @sjforeman.  I've addressed all of the comments and will merge this now.

Issue #124, opened 2021-03-02
Incompatibility of hybrid ringmap maker with simulated sidereal streams


sjforeman:
The routines in `draco.analysis.ringmapmaker` are incompatible with sidereal streams simulated with non-`CHIME` telescope classes in at least two ways:

1. The following check in `MakeVisGrid` fails:
https://github.com/radiocosmology/draco/blob/0c4b3979e4aeecb2d76efc349241b17a7088b532/draco/analysis/ringmapmaker.py#L59-L65
This appears to come from the `view` of `sstream.prodstack` as `np.uint16`. Changing the type to `np.int` resolves things:
![Screen Shot 2021-03-02 at 9 41 00 AM](https://user-images.githubusercontent.com/14794540/109691599-73716e80-7b3c-11eb-8265-cbc0cb16f1b3.png)

2. Sidereal streams from `draco.synthesis.stream.SimulateSidereal` do not have `reverse_map["stack"]`, causing this to fail in `MakeVisGrid`:
https://github.com/radiocosmology/draco/blob/0c4b3979e4aeecb2d76efc349241b17a7088b532/draco/analysis/ringmapmaker.py#L106-L112
Since `reverse_map["stack"]` is constructed by a few lines in `draco.analysis.transform.CollateProducts`, we could simply copy these lines into `draco.synthesis.stream.SimulateSidereal` to ensure that simulated sidereal streams have this field. This would probably help for future compatibility of simulated sidereal streams and other routines as well.

PR #134, opened 2021-05-26
Fixes for simulating sidereal streams

jrs65:


sjforeman (approved):
This conveniently addresses both items in https://github.com/radiocosmology/draco/issues/124

jrs65:
> This conveniently addresses both items in #124

I live to serve.

jrs65:
Actually, I'm finding some funny segfaults when running this. There seems to be a way it can index out of bounds in the Cython code that I don't quite understand at the moment.

sjforeman:
> Actually, I'm finding some funny segfaults when running this. There seems to be a way it can index out of bounds in the Cython code that I don't quite understand at the moment.

Interesting. I ran it myself for a small non-CHIME telescope without any issues.

jrs65:
@sjforeman the segfault was coming from the prod map coming out with an unexpected type and getting cast incorrectly. I think I've fixed that one, and any other risky cats.

PR #135, opened 2021-05-26
refactor(io): create BaseLoadFiles class

ssiegelx:
Moves the code for resolving selections and loading
files into memh5 containers from the LoadFilesFromParams
task to a new BaseLoadFiles task.  This allows other tasks
that load files in a different pipeline context to use these capabilities.


tristpinsm (approved):
Looks good! You should probably just combine those two commits on `io` before merging.

PR #136, opened 2021-05-27
jrs/transform jy to kelvin

jrs65:
@cahofer I think this is now ready for review. I've made a few changes and checked that they give the same output that your code did:
- Use the non-CHIME specific beamclass to determine the beams we need to get the areas of
- Calculate only for frequencies that we get data for, this also enables the beam area calculation to be more simply distributed
- Add a config option to determine whether the stream is copied or not
- Add a config option to control which direction the conversion is done in
- Rename the task to have consistent units in the name (i.e. Jansky and Kelvin, not Jy and Kelvin)

### DIFF
--- a/draco/analysis/transform.py
+++ b/draco/analysis/transform.py
@@ -894,9 +894,19 @@ class SelectPol(task.SingleTask):
 
 
 class TransformJyKelvin(task.SingleTask):
-    """Task to convert from Jy to Kelvin with driftscan beams"""
+    """Task to convert from Jy to Kelvin and vice-versa.
 
-    convert_jy_to_K = True
+    This integrates over the primary beams in the telescope class to derive the
+    brightness temperature to flux conversion.
+
+    Attributes
+    ----------
+    convert_Jy_to_K : bool
+        If True, apply a Jansky to Kelvin conversion factor. If False apply a Kelvin to
+        Jansky conversion.
+    """
+
+    convert_Jy_to_K = config.Property(proptype=bool, default=True)
 
     def setup(self, manager):
         """This setup uses the manager to set up a telescope object and
--- a/draco/analysis/transform.py
+++ b/draco/analysis/transform.py
@@ -893,7 +893,7 @@ class SelectPol(task.SingleTask):
         return outcont
 
 
-class TransformJyKelvin(task.SingleTask):
+class TransformJanskyToKelvin(task.SingleTask):
     """Task to convert from Jy to Kelvin and vice-versa.
 
     This integrates over the primary beams in the telescope class to derive the
--- a/draco/analysis/transform.py
+++ b/draco/analysis/transform.py
@@ -905,12 +905,22 @@ class TransformJanskyToKelvin(task.SingleTask):
     convert_Jy_to_K : bool
         If True, apply a Jansky to Kelvin conversion factor. If False apply a Kelvin to
         Jansky conversion.
+    reference_declination : float, optional
+        The declination to set the flux reference for. A source transiting at this
+        declination will produce a visibility signal equal to its flux. If `None`
+        (default) use the zenith.
     share : {"none", "all"}
         Which datasets should the output share with the input. Default is "all".
+    nside : int
+        The NSIDE to use for the primary beam area calculation. This may need to be
+        increased for beams with intricate small scale structure. Default is 256.
     """
 
     convert_Jy_to_K = config.Property(proptype=bool, default=True)
+    reference_declination = config.Property(proptype=float, default=None)
     share = config.enum(["none", "all"], default="all")
+
+    nside = config.Property(proptype=int, default=256)
 
     def setup(self, telescope: io.TelescopeConvertible):
         """Set the telescope object.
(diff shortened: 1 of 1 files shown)


draco/analysis/transform.py
```diff
@@ -889,3 +892,130 @@ def process(self, polcont):
...
+        """Set the telescope object.
+
+        Parameters
+        ----------
+        telescope
+            An object we can get a telescope object from. This telescope must be able to
+            calculate the beams at all incoming frequencies.
+        """
+        self.telescope = io.get_telescope(telescope)
+        self.telescope._init_trans(256)
```
cahofer, line 925: I am not sure how important this is, but now nside is fixed. For higher observing frequencies and more cylinders you might want a higher nside parameter.

jrs65: Yeah, good point, I was modifying some of that this morning so I'll push some more changes.

draco/analysis/transform.py
```diff
@@ -889,3 +892,130 @@ def process(self, polcont):
...
+            wavelength[:, np.newaxis, np.newaxis] ** 2
+            / (2 * c.k * om_ij[:, :, np.newaxis])
+            * 10 ** (-26)
+        )
+
+        # Get the container we will apply the conversion to (either the input, or a
+        # copy)
+        if self.share == "all":
+            new_stream = sstream
+        else:  # self.share == "none"
```
cahofer, line 1010: is that next to the else a left over debug statement?

jrs65: No. It's just to state what case the else is catching (as share is a two valued enum type).

draco/analysis/transform.py
```diff
@@ -889,3 +892,130 @@ def process(self, polcont):
...
+            new_stream = sstream.copy()
+
+        # Apply the conversion to the data and the weights
+        if self.convert_Jy_to_K:
```
cahofer, line 1014: I was just thinking that in some cases the data might be also have a gain dataset, should we check for such occasions?

cahofer (approved):
Left some minor comments that you can look over.

draco/analysis/transform.py
jrs65: I'm probably going to leave that one as a todo.

draco/analysis/transform.py
```diff
@@ -889,3 +892,159 @@ def process(self, polcont):
...
+        om_ij = np.zeros((len(local_freq), sstream.vis.shape[1]))
+        for fi, freq in enumerate(local_freq):
+            for bi, (bci, bcj) in enumerate(beamclass_pairs):
+                om_i = self._omega_cache[(bci, freq)]
+                om_j = self._omega_cache[(bcj, freq)]
+                om_ij[fi, bi] = (om_i * om_j) ** 0.5
+
+        # Calculate the Jy to K conversion
+        wavelength = c.c / (local_freq * 10 ** 6)
+        Jy_to_K = (
```
ssiegelx, line 1029: For external beam models that are derived from the data, there may be missing frequencies where the beam is set to zero.  In that case, the beam solid angle will be zero and we may want to use `tools.invert_no_zero` here and elsewhere to prevent NaNs.

draco/analysis/transform.py
```diff
@@ -889,3 +892,159 @@ def process(self, polcont):
...
+
+        # Get the container we will apply the conversion to (either the input, or a
+        # copy)
+        if self.share == "all":
+            new_stream = sstream
+        else:  # self.share == "none"
+            new_stream = sstream.copy()
+
+        # Apply the conversion to the data and the weights
+        if self.convert_Jy_to_K:
```
ssiegelx, line 1043: See comment above about using `tools.invert_no_zero` here.

draco/analysis/transform.py
```diff
@@ -889,3 +892,159 @@ def process(self, polcont):
...
+        # the telescope beams.
+        ap_ref = np.array([[np.pi / 2 - np.radians(self.reference_declination), 0.0]])
+        ap_orig = self.telescope._angpos
+        self.telescope._angpos = ap_ref
+        beam_ref = self.telescope.beam(feed, freq)
+        self.telescope._angpos = ap_orig
+
+        # Normalise omega by the squared magnitude of the beam at the reference position
+        beam_ref = np.sum(np.abs(beam_ref) ** 2)
+        omega /= beam_ref
```
ssiegelx, line 965: See comment below about using `tools.invert_no_zero` here.

draco/analysis/transform.py
```diff
@@ -3,7 +3,8 @@
 This includes grouping frequencies and products to performing the m-mode transform.
 """
 import numpy as np
-from caput import mpiarray, config
+from numpy.lib.recfunctions import structured_to_unstructured
+from caput import mpiarray, config, mpiutil
```
ssiegelx, line 7: I don't think `mpiutil` is being used in this module.

ssiegelx (approved):
This looks good and matches the calculation that I was doing to get the conversion factor.  Just a few minor comments.

PR #138, opened 2021-06-02
feat(flagging): create FindBeamformedOutliers and MaskBeamformedOutliers tasks

ssiegelx:
- Provides the capability to mask beamformed visibilities whose
magnitude deviates significantly from our expectation for thermal noise.

- Adds capability to expand each masked pixel by some fixed
number of pixels to capture nearby contaminated pixels that are
below the threshold.

- Creates three new containers for saving the masks:
RingMapMask, FormedBeamMask, and FormedBeamHAMask.
Each has a single mask dataset of the same shape as the
weight dataset in the container they are named after.

- Scales the weight dataset in the beamformer by a factor of 2 to
account for the fact that we are taking the real componenet of the
beamformed visibility.  This missing factor was identified by
beamforming on a Gaussian noise realization and then comparing the
variance in the formed beams to the expectation based on the
weight dataset.


sjforeman (approved):
This all looks good, but let me ask one question to make sure I've parsed the code correctly: in a ringmap with multiple EW beams, if any of these beams at a given (ra, el) is found to exceed the threshold for outliers, then all such beams are masked, correct?

ssiegelx:
@sjforeman Yeah that's correct.  The weight dataset in the RingMap container does not have a beam axis, so I had to somehow collapse over the beam axis in order to apply the mask using the standard method of zero'ing the weights.

The deconvolved ring maps only have one beam, so I think for the stacking analysis it doesn't matter what choice we make here.

sjforeman:
> @sjforeman Yeah that's correct. The weight dataset in the RingMap container does not have a beam axis, so I had to somehow collapse over the beam axis in order to apply the mask using the standard method of zero'ing the weights.
> 
> The deconvolved ring maps only have one beam, so I think for the stacking analysis it doesn't matter what choice we make here.

@ssiegelx Thanks for the clarification! That's what I figured, but I wanted to make sure. I'll merge this now.

PR #139, opened 2021-06-05
feat: create tasks for generating mock catalogs

sjforeman:
This is heavily based on Mateus's code, but has been reorganized and incorporates a handful of additional options.

Contains tasks to:
- construct an approximate selection function from an input catalog of source positions
- resize the selection function to a desired redshift and angular resolution
- construct a PDF from which to draw simulated source positions, based on an input delta_g map and (optionally) a selection function
- draw mock catalogs from an input PDF
- construct a "catalog" of angular positions of Healpix pixel centers for a desired Nside, which can be fed to beamforming routines to perform stacking on these pixel centers


sjforeman:
(I've made this a draft PR because I still have to do some testing for consistency with the older version of the code, but I expect that things should work as expected)

draco/core/containers.py
```diff
@@ -1011,10 +1011,16 @@ def __init__(self, polarisation=True, *args, **kwargs):
...
+        self.attrs["nside"] = int(np.sqrt(len(self.index_map["pixel"]) / 12))
+
     @property
     def map(self):
         return self.datasets["map"]
 
+    @property
+    def nside(self):
+        return self.attrs["nside"]
+
```
jrs65, line 1023: I don't see why this can't just be a dynamic property.
```suggestion
    @property
    def map(self):
        return self.datasets["map"]

    @property
    def nside(self):
        return int((len(self.index_map["pixel"]) //  12) ** 0.5)

```

draco/core/io.py
```diff
@@ -258,6 +258,12 @@ def next(self):
         # Assign a tag to the stack of maps
         map_stack.attrs["tag"] = group["tag"]
 
+        # Add Nside attribute if none exists
```
jrs65, line 261: Remove this one if `.nside` is dynamic.

draco/synthesis/mockcatalog.py
```diff
@@ -0,0 +1,1045 @@
...
+==================
+
+.. autosummary::
+    :toctree:
+
+    _zlims_to_freq
+    _freq_to_z
+    _z_to_freq
+    _pix_to_radec
+    _radec_to_pix
```
jrs65, line 30: Probably don't document internal functions.

draco/synthesis/mockcatalog.py
```diff
@@ -0,0 +1,1045 @@
...
+.. autosummary::
+    :toctree:
+
+    SelFuncEstimator
+    ResizeSelFuncMap
+    PdfGeneratorBase
+    PdfGeneratorUncorrelated
+    PdfGeneratorNoSelfunc
+    PdfGeneratorWithSelfunc
+    MockCatGenerator
```
jrs65, line 17: For some kind of parallelism with the module name, maybe this task should be called `MockCatalogGenerator`

draco/synthesis/mockcatalog.py
```diff
@@ -0,0 +1,1045 @@
...
+
+See Usage section for usage.
+
+Pipeline tasks
+==============
+
+.. autosummary::
+    :toctree:
+
+    SelFuncEstimator
```
jrs65, line 11: Additionally, it might be better to just call this `SelectionFunctionEstimator` (same with the other tasks). It's not horrendously longer and it is clearer.

draco/synthesis/mockcatalog.py
```diff
@@ -0,0 +1,1045 @@
...
+            Output PDF map.
+        """
+
+        # Get MPI comm
+        comm_ = source_map.comm
+
+        # Get local section of selection function
+        selfunc_local = selfunc.map[:, 0, :]
+
+        # Generate weights for distribution of sources in redshift:
```
jrs65, line 446: Not yet. But Anja's changes will help with this. I think then you'll be able to do:
```
z_weights = selfunc_local.sum(axis=1)
z_weights_sum = self.comm.allreduce(z_weights.sum())
```

Actually, I think you can already do that. Things are easier if you use the Python object based `allreduce`. this probably relies on all ranks having at least one redshift.

draco/synthesis/mockcatalog.py
```diff
@@ -0,0 +1,1045 @@
...
+class MockCatGenerator(task.SingleTask):
+    """Take PDF maps generated by task :class:`PdfGenerator`
+    and use it to draw mock catalogs.
+
+    Attributes
+    ----------
+    nsources : int
+        Number of sources to draw in each mock catalog.
+    ncats : int
+        Number of catalogs to generate.
```
jrs65, line 548: To match the rest of our code, these should probably be `nsource` and `ncat`

draco/synthesis/mockcatalog.py
```diff
@@ -0,0 +1,1045 @@
...
+    sigma_z : float, optional
+        Standard deviation of Gaussian redshift errors (default: None)
+    sigma_z_over_1plusz : float, optional
+        Standard deviation of Gaussian redshift errors will be set to
+        this parameter times (1+z). Only one of this and `sigma_z` can
+        be specified. Default: None
```
jrs65, line 554: Should the redshift error generation be a separate task? I can see it being useful elsewhere, e.g. maybe you want to artificially degrade a real catalog.

draco/synthesis/mockcatalog.py
```diff
@@ -0,0 +1,1045 @@
...
+        pdf_map : :class:`containers.Map`
+            PDF from which to draw positions of sources.
+        """
+
+        # Get PDF map container and corresponding healpix Nside
+        self.pdf = pdf_map
+        self.nside = self.pdf.nside
+
+        # Get MPI communicator and rank
+        self.comm_ = self.pdf.comm
```
jrs65, line 588: Just use `self.comm` everywhere. It should be the same. I think the communicator the task gets is checked for consistency with what the data has.

draco/synthesis/mockcatalog.py
```diff
@@ -0,0 +1,1045 @@
...
+                # Populate local arrays of source redshift, RA, dec,
+                # adding random angular offsets from pixel centers if desired
+                mock_zs[source_count] = z_value
+                mock_ra[source_count] = RAbase
+                mock_dec[source_count] = decbase
+                if not self.srcs_at_pixel_centers:
+                    mock_ra[source_count] += ang_size * rtheta[zi][si]
+                    mock_dec[source_count] += ang_size * rphi[zi][si]
+
+                source_count += 1
```
jrs65, line 753: I think this can all be done faster and more cleanly without the explicit loop of over sources within each redshift.

draco/synthesis/mockcatalog.py
```diff
@@ -0,0 +1,1045 @@
...
+        if (self.sigma_z is not None) or (self.sigma_z_over_1plusz is not None):
+            rzerr = [np.random.normal(size=num) for num in source_numbers]
+
+        # If desired, generate random numbers to randomize position of sources
+        # in each healpix pixel. These are uniform random numbers in [-0.5, 0.5],
+        # which will determine the source's relative displacement from the bin's
+        # central RA and dec.
+        if not self.srcs_at_pixel_centers:
+            rtheta = [np.random.uniform(size=num) - 0.5 for num in source_numbers]
+            rphi = [np.random.uniform(size=num) - 0.5 for num in source_numbers]
```
jrs65, line 699: I think these can all be moved within the loop over redshifts below, which avoids needing to store all the random shifts for all redshifts, and means they don't all need to be generated (i.e. you only need to generate them based on the conditional values).

draco/synthesis/mockcatalog.py
jrs65: i.e use vectorized routines and write the results to the right offsets in the `mock_(ra|dec|zs)` array.

draco/synthesis/mockcatalog.py
```diff
@@ -0,0 +1,1045 @@
...
+        sendbuf = [mock_zerrs, len(mock_zs)]
+        self.comm_.Allgatherv(sendbuf, recvbuf)
+        # Gather theta
+        recvbuf = [mock_ra_full, nsource_tuple, dspls, MPI.DOUBLE]
+        sendbuf = [mock_ra, len(mock_ra)]
+        self.comm_.Allgatherv(sendbuf, recvbuf)
+        # Gather phi
+        recvbuf = [mock_dec_full, nsource_tuple, dspls, MPI.DOUBLE]
+        sendbuf = [mock_dec, len(mock_dec)]
+        self.comm_.Allgatherv(sendbuf, recvbuf)
```
jrs65, line 781: Pro-level use of `Allgatherv`. Kudos.

draco/synthesis/mockcatalog.py
```diff
@@ -0,0 +1,1045 @@
...
+        # If we've created the requested number of mocks, prepare to exit
+        if self._count == self.ncats - 1:
+            self.done = True
+
+        return mock_catalog
+
+    def process_finish(self):
+        """Do nothing when last mock has been created."""
+        return None
+
```
jrs65, line 810: Do you understand why this needs to be in there? I would have though omitting this implementation was fine.

jrs65 (changes requested):
@sjforeman looking good. 

I've made a bunch of suggestions, some of which are performance, and some of which are API conventions to make things a little more consistent.

jrs65:
No major issues though. It's looking great.

draco/synthesis/mockcatalog.py
sjforeman: Okay, I split it into another task. I'd like to add the option for the user to input a custom PDF, but in the interest of getting this merged quickly, maybe I'll leave that for a separate PR

draco/synthesis/mockcatalog.py
sjforeman: Good point, and same for the above comment - it was indeed possible to refactor all this to make it nicer

draco/synthesis/mockcatalog.py
sjforeman: All kudos goes to @fandinomat, who wrote all the nontrivial parts!

sjforeman:
Thanks for the comments @jrs65 - I believe I've addressed all of them. I still have to do some more testing, and then I'll convert this to a non-draft PR when I'm done

sjforeman:
Okay, tests look fine - this is ready for a final look

jrs65 (approved):

jrs65:
@sjforeman this looks great now. I'm going to squash merge it down into a single commit. It's a bit extreme, but it's probably the easiest way of "hiding the sausage" as @nritsche would say, and getting it down to a commit that cleanly tells people what has changed.

jrs65:
Awesome. Merged!

PR #141, opened 2021-06-10
Misc. tweaks to SourceStack

sjforeman:
This collects a few minor changes to `draco.analysis.sourcestack.SourceStack`:

- Allowing the use of input `FormedBeam` containers where the frequency axis is either increasing or decreasing. The convention for this is opposite in the `CHIME` telescope class and a generic `driftscan` telescope, so this change allows for stacking with non-`CHIME` telescopes.
- An option to only stack on sources within a specific frequency bin (useful for isolating the stacking signal within a narrow frequency range, as opposed to averaging across the entire band).
- Changing variables and comments to refer to sources instead of quasars/QSOs.

### DIFF
--- a/draco/analysis/sourcestack.py
+++ b/draco/analysis/sourcestack.py
@@ -79,12 +79,20 @@ class SourceStack(task.SingleTask):
 
         # Get f_mask and qs_indices
         freqdiff = freq[np.newaxis, :] - qso_freq[:, np.newaxis]
-        # Stack axis bin edges to digitize each quasar at.
-        stackbins = self.stack_axis["centre"] + 0.5 * self.stack_axis["width"]
-        stackbins = np.append(
-            stackbins,
-            self.stack_axis["centre"][-1] - 0.5 * self.stack_axis["width"][-1],
-        )
+        # Stack axis bin edges to digitize each quasar at, in either increasing
+        # or decreasing order depending on order of frequencies
+        if self.stack_axis["centre"][0] > self.stack_axis["centre"][-1]:
+            stackbins = self.stack_axis["centre"] + 0.5 * self.stack_axis["width"]
+            stackbins = np.append(
+                stackbins,
+                self.stack_axis["centre"][-1] - 0.5 * self.stack_axis["width"][-1],
+            )
+        else:
+            stackbins = self.stack_axis["centre"] - 0.5 * self.stack_axis["width"]
+            stackbins = np.append(
+                stackbins,
+                self.stack_axis["centre"][-1] + 0.5 * self.stack_axis["width"][-1],
+            )
         # Index of each frequency in stack axis, for each quasar
         qs_indices = np.digitize(freqdiff, stackbins) - 1
         # Indices to be processed in full frequency axis for each quasar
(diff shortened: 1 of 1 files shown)


draco/analysis/sourcestack.py
```diff
@@ -77,101 +83,118 @@ def process(self, formed_beam):
...
 
-            # Gather all ranks of qcount. Report number of quasars stacked
-            full_qcount = comm.reduce(qcount, op=MPI.SUM, root=0)
+            # Gather all ranks of count. Report number of quasars stacked
+            full_count = comm.reduce(count, op=MPI.SUM, root=0)
             if comm.rank == 0:
-                self.log.info(
-                    f"Number of quasars stacked for pol {pstr}: {full_qcount}"
-                )
+                self.log.info(f"Number of quasars stacked for pol {pstr}: {full_count}")
```
ssiegelx, line 195:
```suggestion
                self.log.info(f"Number of sources stacked for pol {pstr}: {full_count}")
```
missed a quasar

draco/analysis/sourcestack.py
```diff
@@ -77,101 +83,118 @@ def process(self, formed_beam):
...
-            qstack.stack[oslc] = np.sum(
-                quasar_stack_full.reshape(comm.size, self.nstack), axis=0
-            ) * invert_no_zero(qstack.weight[oslc])
+            stack.stack[oslc] = np.sum(
+                source_stack_full.reshape(comm.size, self.nstack), axis=0
+            ) * invert_no_zero(stack.weight[oslc])
 
-            # Gather all ranks of qcount. Report number of quasars stacked
-            full_qcount = comm.reduce(qcount, op=MPI.SUM, root=0)
+            # Gather all ranks of count. Report number of quasars stacked
```
ssiegelx, line 192:
```suggestion
            # Gather all ranks of count. Report number of sources stacked.
```

draco/analysis/sourcestack.py
```diff
@@ -77,101 +83,118 @@ def process(self, formed_beam):
...
+        source_indices = np.digitize(freqdiff, stackbins) - 1
+        # Indices to be processed in full frequency axis for each source
+        f_mask = (source_indices >= 0) & (source_indices < self.nstack)
+        # Only sources in the frequency range of the data.
+        source_mask = (np.sum(f_mask, axis=1) > 0).astype(bool)
+        # If desired, also restrict to sources within a specific channel.
+        # This works because the frequency axis is not distributed between
+        # ranks.
+        if self.single_source_bin_index is not None:
+            restricted_chan_mask = (
```
ssiegelx, line 113: I think this expression will fail for edge bins, in the sense that all sources that fall outside the range of frequencies contained in the formed beam will be included in the edge bin's `restricted_chan_mask`.  I'm not sure if this is an issue for your use case.

I think something like:
```
fs = formed_beam.index_map['freq'][self.single_source_bin_index]
restricted_chan_mask = np.abs(source_freq - fs['centre']) < (0.5 * fs['width'])
```
would handle that particular edge case properly.

ssiegelx (approved):
Hey Simon, this looks good, sorry it took me so long to review.

sjforeman:
Thanks for the review @ssiegelx , and sorry for the late follow-up! I implemented your suggestions and rebased to clean up the commit history. I'll merge this once all the CI checks are done

PR #142, opened 2021-06-16
fix(ringmapmaker): MPI calls could get out of sync

jrs65:
This fixes a nasty bug where MPI calls could get out of sync when using
`DeconvolveAnalyticalBeam` in a case where different ranks have
differing numbers of frequencies. To rectify this it refactors the code
such that the beam_mmodes are generate for all frequencies at once which
will prevent the bug from resurfacing in newer implementations.

### DIFF
--- a/draco/analysis/ringmapmaker.py
+++ b/draco/analysis/ringmapmaker.py
@@ -582,7 +582,9 @@ class DeconvolveHybridMBase(task.SingleTask):
         # Dereference datasets
         hv = hybrid_vis_m.vis[:].view(np.ndarray)
         hw = hybrid_vis_m.weight[:].view(np.ndarray)
-        bv = hybrid_beam_m.vis[:].view(np.ndarray)
+
+        # Dereference the beams, and trim to the set of m's present in the input data
+        bv = hybrid_beam_m.vis[:].view(np.ndarray)[: (mmax + 1)]
 
         rmm = rm.map[:]
         rmw = rm.weight[:]
--- a/draco/analysis/ringmapmaker.py
+++ b/draco/analysis/ringmapmaker.py
@@ -547,8 +547,7 @@ class DeconvolveHybridMBase(task.SingleTask):
         m = hybrid_vis_m.index_map["m"]
         mmax = hybrid_vis_m.mmax
 
-        is_odd = np.any(hybrid_vis_m.vis[mmax, 1] != 0.0)
-        nra = 2 * mmax + int(is_odd)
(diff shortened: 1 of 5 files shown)


draco/analysis/ringmapmaker.py
```diff
@@ -821,93 +856,60 @@ def B(phi, u, sigma):
             """Azimuthal beam transfer function."""
             return np.exp(2.0j * np.pi * u * np.sin(phi)) * A(phi, sigma)
 
-        # Deteremine the RA axis from the maximum m-mode in the hybrid visibilities
+        # Determine the RA axis from the maximum m-mode in the hybrid visibilities
         mmax = hybrid_vis_m.mmax
         is_odd = np.any(hybrid_vis_m.vis[mmax, 1] != 0.0)
+        is_odd = hybrid_vis_m.comm.allreduce(is_odd)
```
ssiegelx, line 862: Should we add this `allreduce` to the other places where we determine if it's an even or odd number of RAs?  This is done on line 550 in the `DeconvolveHybridMBase`, and then also in the `_unpack_marray` used by `MModeInverseTransform`.

I guess a scenario where those instances would fail is when there is an odd number of RAs and some missing frequencies that are entirely zero.  If a rank handles only those missing frequencies then this expression would evaluate False for that rank only.

draco/analysis/transform.py
```diff
@@ -595,25 +598,39 @@ def process(self, mmodes):
...
         sstream.vis[:] = ssarray
         # There is no way to recover time information for the weights.
         # Just assign the time average to each baseline and frequency.
-        sstream.weight[:] = mmodes.weight[0, 0, :, :][:, :, np.newaxis] / ntime
+        sstream.weight[:] = mmodes.weight[0, 0, :, :][:, :, np.newaxis] / nra
 
         return sstream
 
 
+class SiderealMModeResample(task.group_tasks(MModeTransform, MModeInverseTransform)):
```
ssiegelx, line 620: Cool!

ssiegelx (approved):
This looks great.  I just have one comment.

draco/analysis/ringmapmaker.py
jrs65: Good point. I think the thing to do here is to keep this as a property on `MContainer`, it can be set by the MModeTransform class as it knows what the original set of RAs are, and then just be looked up by these other tasks.

jrs65:
@ssiegelx I think this is done. I added a new `oddra` property to `MContainer` which tracks if it came from a odd/even number of RA points. This allows the containers to be exactly invertible without hackish checks.

PR #144, opened 2021-06-25
Mockcat fixes and other goodies

jrs65:


sjforeman (approved):
Sorry for the delayed review! This all looks good - I only have a few minor comments about documentation. (I actually have my own unmerged version of the healpix beamforming task, but yours is much cleaner...)

jrs65:
Thanks!

PR #145, opened 2021-06-29
build: raise minimum required numpy

anjakefala:
i"There was a change in the C API
(https://numpy.org/doc/stable/release/1.20.0-notes.html#size-of-np-ndarray-and-np-void-changed)
in numpy 1.20.0. This can result in issues with running draco if the build version of
numpy, and the installed version, differed.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,6 +3,6 @@ caput @ git+https://github.com/radiocosmology/caput.git
 cora @ git+https://github.com/radiocosmology/cora.git
 Cython>0.18
 mpi4py
-numpy>=1.17
+numpy>=1.20.3
 scipy>=0.10
 skyfield


nritsche (approved):

jrs65:
Can you elaborate on why we need to do this? It sounds like it's not a problem with the version of numpy, but one of the build environment, i.e. there is no feature we need in numpy that isn't already in 1.17.

anjakefala:
We continue to be running into this problem on Cedar, because packages are built with > 1.20, and there is an installed < 1.20. I am not quite sure yet why packages are being built with > 1.20, and where these numpys are all located, but the ways folks have gone around this is by updating their numpy version in their virtual environment.

This is why this change was proposed.

anjakefala:
I think I agree that this change will not help. 

Looking at `cedar`:

```
module spider numpy

----------------------------------------------------------------------------------------------------------------------------------------------
  numpy:
----------------------------------------------------------------------------------------------------------------------------------------------
     Versions:
        numpy/1.16.6 (E)
        numpy/1.18.4 (E)
        numpy/1.19.2 (E)
        
        (pipeline)[anja@cedar5 ~]$ module spider numpy/1.19.2

----------------------------------------------------------------------------------------------------------------------------------------------
  numpy: numpy/1.19.2 (E)
----------------------------------------------------------------------------------------------------------------------------------------------
    This extension is provided by the following modules. To access the extension you must load one of the following modules. Note that any module names in parentheses show the module location in the software hierarchy.


       scipy-stack/2020b (StdEnv/2020)


Names marked by a trailing (E) are extensions provided by another module.

```

The problem is probably that when our packages were installed, they were built with `numpy > 1.2.0`, but the installed numpy in the cedar environment is < 1.19.

anjakefala:
I am going to close this PR.

jrs65:
Yeah. I agree with this. The way we have our env set up on cedar it's quite hard for us to upgrade numpy. To get around build issues like this in the past I've needed to turn off pip's build isolation which mostly solves it. Is it just building draco which causes the problem, or is it other packages too?

ashill:
I (and @ssiegelx ) encountered the problem when doing `from draco.core import container`, not when building.

```from draco.core import containers
ValueErrorTraceback (most recent call last)
<ipython-input-1-e0800402881d> in <module>
----> 1 from draco.core import containers
~/venv/lib/python3.8/site-packages/draco/core/containers.py in <module>
     61 from caput import memh5, tod
     62 
---> 63 from ..util import tools
     64 
     65 # Try to import bitshuffle to set the default compression options
~/venv/lib/python3.8/site-packages/draco/util/tools.py in <module>
      7 from numpy.lib.recfunctions import structured_to_unstructured
      8 
----> 9 from ._fast_tools import _calc_redundancy
     10 
     11 
draco/util/_fast_tools.pyx in init draco.util._fast_tools()
ValueError: numpy.ndarray size changed, may indicate binary incompatibility. Expected 88 from C header, got 80 from PyObject
```

In my case, a simple `pip install --upgrade numpy` fixed the problem; Seth and others got around it by creating a clean environment (which I guess either built everything using `numpy < 1.20` or installed `numpy >= 1.20`; seems like either works but not a mix).

This is an awfully obscure bug for users to track down. Is there some way to at least include in the error message that the problem is a mismatch between the build and runtime versions of numpy?

jrs65:
This is the crux of the issue, upgrading numpy sort of works and is the obvious thing that a user would do, but it will then break all other binary modules that build against numpy, and thus require upgrading them, and you're likely to lose the OpenMP parallel version of `np.dot` that is in the provided version on cedar (this is on cedar, right?)

I had thought that the cedar folks setting of the pip config to point at their own "wheelhouse" meant that it was actually very hard to get a new version of numpy (i.e. `pip install -U numpy` would only ever install their own provided version), so they may have put their own version of 1.20 in there, or we disable their pip config.

jrs65:
In summary. I don't know what a great solution is to this problem is. Really the numpy folks should stop breaking their 
ABI so often, and pip should get rid of the stupid build isolation, but those are out of our control.

jrs65:
This *may* be the solution. Replace the numpy build requirement in `pyproject.toml` with https://github.com/scipy/oldest-supported-numpy

jrs65:
I think that should work. Apparently the ABI changes are usually backwards compatible, such that building against an older version, and then using a newer version at run time should be okay. If it does work, then we should look at replacing all our build times deps with it. @anjakefala can you  give it a spin?

anjakefala:
@jrs65 Yup!

PR #147, opened 2021-07-21
Stacking/beamform bug fixes

jrs65:
- fix(RingMapBeamform): incorrect positions when ringmap has a reduced span of ZA
- fix(mockcatalog): NaN's produced if selection func zero for a z slice
- fix: ensure attributes are copied over during container creation
- refactor: use a consistent scheme for matching catalog and data coordinate epochs


sjforeman (approved):
These look great to me! I have a few related changes that will go on top of these, but I'll make a separate PR for those once this has been merged

PR #149, opened 2021-07-28
Task to add eBOSS-like redshift errors to catalog

sjforeman:
This refactors Seth's eBOSS redshift error code into a separate task, and implements a few other minor tweaks.

### DIFF
--- a/draco/synthesis/mockcatalog.py
+++ b/draco/synthesis/mockcatalog.py
@@ -844,9 +844,11 @@ class AddEBOSSZErrorsToCatalog(task.SingleTask, random.RandomTask):
     ----------
     tracer : {"ELG"|"LRG"|"QSO"}
         Generate redshift errors corresponding to this eBOSS sample.
+        If not specified, task will attempt to detect the tracre type from
+        the catalog's `tracer` attribute or its tag. Default: None
     """
 
-    tracer = config.enum(["QSO", "ELG", "LRG"])
+    tracer = config.enum(["QSO", "ELG", "LRG"], default=None)
 
     def process(self, cat):
         """Generate random redshift errors and add to redshifts in catalog.
@@ -861,6 +863,27 @@ class AddEBOSSZErrorsToCatalog(task.SingleTask, random.RandomTask):
         cat_out : :class:`containers.SpectroscopicCatalog`
             Catalog with redshift errors added.
         """
+
+        tracer = self.tracer
+
+        # If tracer not specified in config, check to see whether it's stored
+        # in the catalog's 'tracer' attribute or in its tag
+        if tracer is None:
+            if "tracer" in cat.attrs:
+                tracer = cat.attrs["tracer"]
+            else:
+                for key in _velocity_error_function_lookup.keys():
+                    if key in cat.attrs["tag"]:
+                        tracer = key
+                        break
+
+                if tracer is None:
+                    raise ValueError(
+                        "Must specify eBOSS tracer in config property, "
+                        "catalog 'tracer' attribute, or catalog 'tag' attribute."
+                    )
+
+        self.log.info(f"Applying {tracer} redshift errors.")
 
         # Get redshifts from catalog
         cat_z = cat["redshift"]["z"][:]
@@ -894,12 +917,6 @@ class AddEBOSSZErrorsToCatalog(task.SingleTask, random.RandomTask):
         dz: np.ndarray[nsource,]
             Perturbations to source redshifts based on random velocity errors.
         """
-
-        _velocity_error_function_lookup = {
-            "QSO": self.qso_velocity_error,
-            "ELG": self.elg_velocity_error,
-            "LRG": self.lrg_velocity_error,
-        }
 
         if tracer not in _velocity_error_function_lookup:
             raise ValueError(
@@ -1015,6 +1032,13 @@ class AddEBOSSZErrorsToCatalog(task.SingleTask, random.RandomTask):
         dv = dist.rvs(ELG_LAMBDA, scale=ELG_SIG, size=nsample)
 
         return dv
+
+
+_velocity_error_function_lookup = {
+    "QSO": AddEBOSSZErrorsToCatalog.qso_velocity_error,
+    "ELG": AddEBOSSZErrorsToCatalog.elg_velocity_error,
+    "LRG": AddEBOSSZErrorsToCatalog.lrg_velocity_error,
+}
 
 
 class MapPixelLocationGenerator(task.SingleTask):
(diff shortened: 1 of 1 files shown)


draco/synthesis/mockcatalog.py
```diff
@@ -795,19 +809,213 @@ def process(self, cat):
...
+        err_func = _velocity_error_function_lookup[tracer]
+
+        dv = err_func(z.size, self.rng)
+
+        dz = (1.0 + z) * dv / (C * 1e-3)
+
+        return dz
+
+    @staticmethod
+    def _qso_velocity_error(nsample, rng):
```
jrs65, line 919: This all looks great. My only comment would be to remove the `_` prefixes from these so they're no longer "private" and it's clearer that they could be used elsewhere.

draco/synthesis/mockcatalog.py
```diff
@@ -753,16 +759,23 @@ def process(self):
...
+    catalog.
 
-    TODO: allow for user-specified PDF.
+    Note that the errors are added to the catalog in place, such that
+    if the original catalog is subsequently used in a pipeline, it will
+    have the errors included.
 
     Attributes
     ----------
+    use_catalog_z_errors : bool
```
jrs65, line 775: Yeah, this seems like a good option to have.

draco/synthesis/mockcatalog.py
```diff
@@ -795,19 +809,213 @@ def process(self, cat):
...
+        cat_z_err = cat["redshift"]["z_error"][:]
+
+        # Generate redshift errors for the chosen tracer
+        z_err = self._generate_z_errors(cat_z, self.tracer)
 
         # Add errors to catalog redshifts
         cat_z += z_err
         # Add errors in quadrature with existing stored errors
-        cat_z_err = (z_err ** 2 + cat_z_err ** 2) ** 0.5
+        cat_z_err[:] = (z_err ** 2 + cat_z_err ** 2) ** 0.5
```
jrs65, line 875: It's not obvious this makes any sense. `z_err` is a realisation of a random variable, but you want to add in an estimate of the distribution variance instead. That probably requires another method to calculate that.

jrs65 (changes requested):
A few small changes, otherwise very nice!

draco/synthesis/mockcatalog.py
sjforeman: Changed!

ssiegelx (approved):
This looks great, thanks for pulling it all together.  I just have one comment.

draco/synthesis/mockcatalog.py
sjforeman: Agreed - this somehow snuck through a previous review, and I left it alone as I wasn't sure what would be best here. Do we want to save the standard deviation of the underlying distribution even if the distribution isn't Gaussian? What is the intended use of the `z_error` field - simply a reflection of a (Gaussian) errorbar on `z`?

draco/synthesis/mockcatalog.py
```diff
@@ -795,19 +809,213 @@ def process(self, cat):
...
+    if the original catalog is subsequently used in a pipeline, it will
+    have the errors included.
+
+    Attributes
+    ----------
+    tracer : {"ELG"|"LRG"|"QSO"}
+        Generate redshift errors corresponding to this eBOSS sample.
+    """
+
+    tracer = config.enum(["QSO", "ELG", "LRG"])
```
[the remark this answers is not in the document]
jrs65: That sounds very useful, although fwiw it wouldn't actually help the way I run things (I use `caput-pipeline template-run` to have the tracer be a variable within a template that gets filled in and run for each tracer type).

draco/synthesis/mockcatalog.py
jrs65: The point of the `z_error` field was simply that it was a column available in the SDSS database and it sounded useful to have in our analysis, but I don't think we've ever actually used it, and so we've no real restriction of what it should be.

I think in this case, probably the best thing would be to estimate an effective sigma for each type (even for the non-Gaussian distributions), but I don't think we're really going to need it immediately so I'd be happy just not modifying it in this task and making an issue to do something smarter at a future date.

draco/synthesis/mockcatalog.py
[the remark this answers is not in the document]
jrs65: Just to be sure, maybe I'll modify the Mockcatalog code to add a tracer attribute anyway before I generate all the catalogs.

draco/synthesis/mockcatalog.py
sjforeman: Sure, sounds good to me

draco/synthesis/mockcatalog.py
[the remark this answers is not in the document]
sjforeman: Thanks Seth - I inadvertently dropped this functionality from your code, so thanks for reminding me to put it back in!

draco/synthesis/mockcatalog.py
sjforeman: See https://github.com/radiocosmology/draco/issues/151

draco/synthesis/mockcatalog.py
```diff
@@ -795,20 +809,236 @@ def process(self, cat):
...
+
+    Note also that the errors are added to the catalog in place, such that
+    if the original catalog is subsequently used in a pipeline, it will
+    have the errors included.
+
+    Attributes
+    ----------
+    tracer : {"ELG"|"LRG"|"QSO"}
+        Generate redshift errors corresponding to this eBOSS sample.
+        If not specified, task will attempt to detect the tracre type from
```
jrs65, line 845:
```suggestion
        If not specified, task will attempt to detect the tracer type from
```

draco/synthesis/mockcatalog.py
```diff
@@ -795,20 +809,236 @@ def process(self, cat):
...
+                for key in _velocity_error_function_lookup.keys():
+                    if key in cat.attrs["tag"]:
+                        tracer = key
+                        break
+
+                if tracer is None:
+                    raise ValueError(
+                        "Must specify eBOSS tracer in config property, "
+                        "catalog 'tracer' attribute, or catalog 'tag' attribute."
+                    )
```
jrs65, line 882: Maybe make this detection case insensitive? We've tended to use a bit of a mixed version of the tracer abbreviations.

Also check that the tracer attribute is actually understood.
```suggestion
        if tracer is None:
            if "tracer" in cat.attrs:
                tracer = cat.attrs["tracer"].upper()
                
                if tracer not in _velocity_error_function_lookup:
                    raise ValueError(
                        f"Tracer explicitly set to '{tracer}' in catalog, but value not supported."
                    )
            else:
                for key in _velocity_error_function_lookup.keys():
                    if key in cat.attrs["tag"].upper():
                        tracer = key
                        break

                if tracer is None:
                    raise ValueError(
                        "Must specify eBOSS tracer in config property, "
                        "catalog 'tracer' attribute, or catalog 'tag' attribute."
                    )
```

jrs65 (approved):
A few minor suggestions, but you can probably ignore them if you want.

jrs65:
Looks good to me!

PR #152, opened 2021-07-31
ss/tel-beam-stream

ssiegelx:
Adds a new `CreateBeamStreamFromTelescope` task that evaluates
the beam model in a telescope instance to construct a `GridBeam`,
which is then passed to the `CreateBeamStream` superclass to convert
to `HybridVisStream` for use in the deconvolving ringmap maker.

### DIFF
--- a/draco/analysis/beam.py
+++ b/draco/analysis/beam.py
@@ -49,7 +49,7 @@ class CreateBeamStream(task.SingleTask):
         )
 
     def process(self, data, beam):
-        """Convert the beam model into a format that can be applied to the data.
+        """Convert the beam model into a format that can be deconvolved from data.
 
         Parameters
         ----------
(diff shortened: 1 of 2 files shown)


ssiegelx:
[sourcestack_rev_03_202106_compare_psrc_beam_rev02_and_fit_beam.pdf](https://github.com/radiocosmology/draco/files/6913610/sourcestack_rev_03_202106_compare_psrc_beam_rev02_and_fit_beam.pdf)

I used the code in this PR to create a ringmap with the `CHIMEFitBeam` deconvolved.  Here is a comparison of the resulting source stack to that obtained using the `rev 02` point source beam.  There does not appear to be any significant difference in the amplitude of the signal.  I think this is expected given that the fit beam is approximately equal to the point source beam when averaged over a large enough field.  I would have maybe expected more of an improvement in the noise with the point source beam since we should be appropriately down weighting (up weighting) frequencies and declinations where the response is low (high) because of the the 30 MHz ripples.

jrs65 (approved):

PR #153, opened 2021-08-03
Fix bug using tracer and allow setting of it 

jrs65:
- feat(mockcatalog): add a tracer attribute to pdfs and mockcatalogs
- fix(AddEBOSSZErrorsToCatalog): bug when using tracer from catalog

### DIFF
--- a/draco/synthesis/mockcatalog.py
+++ b/draco/synthesis/mockcatalog.py
@@ -916,7 +916,7 @@ class AddEBOSSZErrorsToCatalog(task.SingleTask, random.RandomTask):
         cat_z_err = cat["redshift"]["z_error"][:]
 
         # Generate redshift errors for the chosen tracer
-        z_err = self._generate_z_errors(cat_z, self.tracer)
+        z_err = self._generate_z_errors(cat_z, tracer)
 
         # Add errors to catalog redshifts
         cat_z += z_err
--- a/draco/synthesis/mockcatalog.py
+++ b/draco/synthesis/mockcatalog.py
@@ -140,6 +140,9 @@ class SelectionFunctionEstimator(task.SingleTask):
         Number of SVD modes used in recovering the selection function from
         the catalog maps.
         Default: 7.
+    tracer : str, optional
+        Set an optional tracer attribute that can be used to identify the type of
+        catalog later in the pipeline.
     """
 
     bcat_path = config.Property(proptype=str, default=None)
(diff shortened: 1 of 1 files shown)


jrs65:
@sjforeman another one for you to review as you're such a reviewing machine. Plus one commit fixes a bug in your code ;-)

draco/synthesis/mockcatalog.py
```diff
@@ -318,8 +318,16 @@ class PdfGeneratorBase(task.SingleTask):
     by the task :class:`MockCatalogGenerator` to draw mock catalogs.
 
     Derived classes must implement process().
+
+    Attributes
+    ----------
+    tracer : str, optional
```
sjforeman, line 324: It would also make sense to have a `tracer` parameter in `SelectionFunctionEstimator`, and have it stored as an attribute in the selection function map and propagated forward from there. It is also nice to have the `tracer` parameter where you've put it, though, in case the user wants to specify the tracer but isn't using a selection function. Perhaps we could have both options, and add code to `PdfGeneratorWithSelectionFunction` to check for a tracer attribute in the selection function container if none is specified in `PdfGeneratorWithSelectionFunction` itself. What do you think?

sjforeman (commented):
Thanks for catching my bug! See my comment about where we should put the `tracer` parameter...

draco/synthesis/mockcatalog.py
jrs65: Good point. I guess maybe the thing to do is to move the functionality onto `draco.core.task.SingleTask` to be a generic way of setting an attribute on the output of any task.

jrs65: Okay. I've added the same code into the SelectionFunctionEstimator. I'll make an issue for the generic functionality.

jrs65:
Issue for generic functionality in #154

draco/synthesis/mockcatalog.py
```diff
@@ -374,12 +391,17 @@ def make_pdf_map(self, source_map, z_weights, selfunc=None):
...
+            attrs_from=selfunc,
         )
 
         # Put computed PDF into local section of container
         pdf_map_local = pdf_map.map[:]
         pdf_map_local[:, 0, :] = pdf
 
+        # Set a tracer attribute
+        if self.tracer is not None:
+            pdf_map.attrs["tracer"] = self.tracer
```
sjforeman, line 403: This will override the `tracer` attribute inherited from the selection function (if any), but I guess that's fine

sjforeman (approved):

draco/synthesis/mockcatalog.py
jrs65: Yeah. I think that's generally the expected behaviour everywhere, i.e. if you explicitly set anything then it's expected to take precedence and overwrite anything else.

PR #156, opened 2021-08-05
Mix data and fix flux cal bug

jrs65:
- fix(ringmapmaker): make complex to real cast explicit
- feat(MixData): a task for generating arbitrary mixes of data and weights
- fix(TransformJanskyToKelvin): incompatibility with CHIMEExternalBeam

### DIFF
--- a/draco/analysis/transform.py
+++ b/draco/analysis/transform.py
@@ -976,26 +976,27 @@ class TransformJanskyToKelvin(task.SingleTask):
 
     def _beam_area(self, feed, freq):
         """Calculate the primary beam solid angle."""
+        import healpy
 
         beam = self.telescope.beam(feed, freq)
         horizon = self.telescope._horizon[:, np.newaxis]
+        beam_pow = np.sum(np.abs(beam) ** 2 * horizon, axis=1)
 
         pxarea = 4 * np.pi / beam.shape[0]
-        omega = np.sum(np.abs(beam) ** 2 * horizon) * pxarea
-
-        # Calculate the beam value at the reference point by temporarily swapping out
-        # the telescopes internal `_angpos` attribute for one that just contains the
-        # reference position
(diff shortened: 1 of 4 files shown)


draco/analysis/transform.py
```diff
@@ -976,26 +976,27 @@ def setup(self, telescope: io.TelescopeConvertible):
...
-        beam_ref = np.sum(np.abs(beam_ref) ** 2)
+        # NOTE: this is slightly less accurate than the previous approach of reseting
+        # the internal `_angpos` property to force evaluation of the beam at the exact
+        # coordinates, but is more generically applicable, and works (for instance) with
+        # the CHIMEExternalBeam class.
+        #
+        # Also, for a reason I don't fully understand it's more accurate to use the
+        # value of the pixel including the reference position, and not do an
+        # interpolation using it's neighbours...
+        beam_ref = beam_pow[
```
sjforeman, line 997: The issues here arise from a more general problem in driftscan, that the current API makes it cumbersome for a user to compute the beam at a given set of angular locations, unless they call a given telescope's `beam_x()` or `beam_y()` methods directly. We could consider adding methods to `drift.core.telescope.UnpolarisedTelescope` and `drift.core.telescope.SimplePolarisedTelescope` to make it easier to do this. In cases like `CHIMEExternalBeam`, where the beam model is not analytical, interpolation of the input beam map to arbitrary locations would be required, and if the included healpy interpolation routines don't work very well, it might take some effort to find a good implementation.

We can definitely punt on this for now, but it would be nice to sort this out in the future...

sjforeman (approved):
Looks good overall!

draco/analysis/transform.py
jrs65: 100% agree here. We need a much better API for this built into the driftscan core.

PR #157, opened 2021-08-06
ss/dayenu

ssiegelx:
Inverse covariance filtering that can handle missing data.
Creates tasks for filtering visibilities in delay and m-mode space,
and for filtering ringmaps in delay space with arbitrary
polarisation/declination dependent cutoffs.


draco/analysis/flagging.py
```diff
@@ -417,14 +417,71 @@ def process(self, data, mask):
...
+        # Multiply the weights by the inverse of the mask
         flag = ~mask.mask[:].view(np.ndarray)
 
         data.weight[:] *= flag.astype(np.float32)
 
         return data
 
 
+class MaskBeamformedWeights(task.SingleTask):
+    """Mask beamformed visibilities with anomalously large weights before stacking."""
```
jrs65, line 433: It's kind of obvious what it is, but ideally there would be a short docstring about the `nmed` attribute.

draco/analysis/dayenu.py
```diff
@@ -0,0 +1,746 @@
...
+
+    ucov = uflag * cov[np.newaxis, :, :]
+
+    pinv = np.linalg.pinv(ucov, hermitian=True) * uflag
+    pinv = pinv[uindex, :, :].reshape(oshp)
+
+    return pinv
+
+
+def instantaneous_m(ha, lat, dec, u, v, w=0.0):
```
jrs65, line 719: This seems like probably something we should move to `caput.interferometry` along with the fringe phase code. I wouldn't really delay this PR on that basis though.

draco/analysis/dayenu.py
```diff
@@ -0,0 +1,746 @@
...
+    cov = np.eye(nra, dtype=np.float64)
+    cov += np.sinc(m_cut * dra / np.pi) / epsilon
+
+    uflag, uindex = np.unique(flag.reshape(-1, nra), return_inverse=True, axis=0)
+    uflag = uflag[:, np.newaxis, :] & uflag[:, :, np.newaxis]
+    uflag = uflag.astype(np.float64)
+
+    ucov = uflag * cov[np.newaxis, :, :]
+
+    pinv = np.linalg.pinv(ucov, hermitian=True) * uflag
```
jrs65, line 713: You're probably better using the `scipy.linalg` version as it'll me MKL accelerated.

draco/analysis/dayenu.py
```diff
@@ -0,0 +1,746 @@
...
+    cov = np.eye(nra, dtype=np.float64) / aeps
+    cov += a * (1.0 - 1.0 / aeps) * np.sinc(m_cut * dra / np.pi)
+
+    uflag, uindex = np.unique(flag.reshape(-1, nra), return_inverse=True, axis=0)
+    uflag = uflag[:, np.newaxis, :] & uflag[:, :, np.newaxis]
+    uflag = uflag.astype(np.float64)
+
+    ucov = uflag * cov[np.newaxis, :, :]
+
+    pinv = np.linalg.pinv(ucov, hermitian=True) * uflag
```
jrs65, line 668: Also use scipy.linalg.

draco/analysis/dayenu.py
```diff
@@ -0,0 +1,746 @@
...
+        * np.cos(m_center * dra)
+    )
+
+    uflag, uindex = np.unique(flag.reshape(-1, nra), return_inverse=True, axis=0)
+    uflag = uflag[:, np.newaxis, :] & uflag[:, :, np.newaxis]
+    uflag = uflag.astype(np.float64)
+
+    ucov = uflag * cov[np.newaxis, :, :]
+
+    pinv = np.linalg.pinv(ucov, hermitian=True) * uflag
```
jrs65, line 620: Also here for scipy.linalg.

draco/analysis/dayenu.py
```diff
@@ -0,0 +1,746 @@
...
+    return pinv
+
+
+def instantaneous_m(ha, lat, dec, u, v, w=0.0):
+    """Calculate the instantaneous fringe-rate.
+
+    Parameters
+    ----------
+    ha : float
+        Hour angle in radians.
```
jrs65, line 725: No docstring for `lat`

jrs65 (changes requested):
Noting that we've tested this pretty heavily and we're going to merge the dayenu and delay code, I've just given this a fairly light review.

sjforeman (approved):
Looks good - my comments are only about documentation

draco/analysis/dayenu.py
ssiegelx: I changed this but then realized that the reason I used the numpy version in the first place is because it accepts a stack of matrices, whereas the scipy version does not.  We stack up all the unique sets of flags and construct the filter with one call, whereas with the scipy version we would have to loop over the unique sets of flags.  I'm not sure how much that speeds up things (if at all) compared to the gains you get from the MKL accelerated version.  In our current scenario we only have one set of flags since complete sidereal coverage is required for the m-mode transform, so scipy with loop is probably faster.  I'm working on timing this for different scenarios now.

jrs65: That's annoying.

Usually the numpy versions are slower implementations and single threaded, so you'd expect ~8x speed up if the matrices are "large enough" (probably anything over N>200 is big enough). However, it might just not be worth the time to fix it at the moment as we're going to refactor everything at somepoint. Could just put a comment about why the numpy versions are being used and the leave it at that.

draco/analysis/dayenu.py
ssiegelx: Yeah that makes sense.  I have been using something like this in several places (constructing the m-space window for the deconvolving ringmap maker and the point source beam calibration).

I can issue a follow up PR that moves this method to `caput`.

draco/analysis/dayenu.py
ssiegelx: I was able to get the total run time of the task down from 35 minutes to 2 minutes (using 4 nodes).

It turned out that applying the filter to the map was very slow.  The filter was being applied to a view of the map in the container at a specific (beam, pol, el) that was not contiguous in memory.  Creating a contiguous array first resulted in a factor of 200 speed up in the filter application.

In the case where we have a single set of frequency flags for all right ascensions, the filter construction is only 5-10% of the total run time (after the fix above).  But if we ever use more complicated masks that change as a function of RA, then that will grow linearly with the number of unique sets of flags.

I went ahead and switched to using the scipy version of pinv.  However my testing actually suggested that this is 5-10% slower than the numpy version.  As mentioned above, this doesn't really matter for our current usage since filter construction is a small fraction of total run time.  I don't want to spend much more time on this now, but we may want to investigate scipy versus numpy a bit more when we refactor this code.

draco/analysis/dayenu.py
jrs65: Sounds good. Although not very high priority!

draco/analysis/dayenu.py
jrs65: Ah nice. That sounds awesome!

jrs65 (approved):

jrs65:
I think this can be merged whenever you're happy @ssiegelx

PR #160, opened 2021-09-16
RA integration window for sidereal streams and improve redshift errors

jrs65:


draco/analysis/transform.py
```diff
@@ -441,8 +441,17 @@ class MModeTransform(task.SingleTask):
 
     The maximum m used in the container is derived from the number of
     time samples, or if a manager is supplied `telescope.mmax` is used.
+
+    Attributes
+    ----------
+    remove_integration_window : bool
+        Deconvolve the effect of the finite width of the RA integration (presuming it
+        was a rectangular integration window). This is applied to both the visibilies
```
sjforeman, line 449:
```suggestion
        was a rectangular integration window). This is applied to both the visibilities
```

draco/analysis/transform.py
```diff
@@ -513,6 +522,19 @@ def process(self, sstream: containers.SiderealContainer) -> containers.MContaine
...
+        if self.remove_integration_window:
+
+            m = np.arange(mmax + 1)
+            w = np.sinc(m / nra)
+            inv_w = tools.invert_no_zero(w)
+
+            sl_vis = (slice(None),) + (np.newaxis,) * (len(ma.vis.shape) - 1)
+            ma.vis[:] *= inv_w[sl_vis]
+
+            sl_weight = (slice(None),) + (np.newaxis,) * (len(ma.vis.shape) - 1)
```
sjforeman, line 535: I think you want the shape of the weight dataset here (this doesn't matter for `MModes`, but makes a difference for `HybridVisMModes`):
```suggestion
            sl_weight = (slice(None),) + (np.newaxis,) * (len(ma.weight.shape) - 1)
```

draco/analysis/transform.py
```diff
@@ -573,16 +595,25 @@ class MModeInverseTransform(task.SingleTask):
...
     Attributes
     ----------
     nra : int
         Number of RA bins in the output. Note that if the number of samples does not
         Nyquist sample the maximum m, information may be lost. If not set, then try to
         get from an `original_nra` attribute on the incoming MModes, otherwise determine
         an appropriate number of RA bins from the mmax.
+    apply_integration_window : bool
+        Apply the effect of the finite width of the RA integration (presuming a
+        rectangular integration window). This is applied to both the visibilies and the
```
sjforeman, line 610:
```suggestion
        rectangular integration window). This is applied to both the visibilities and the
```

draco/analysis/transform.py
```diff
@@ -610,6 +641,19 @@ def process(self, mmodes: containers.MContainer) -> containers.SiderealContainer
...
+        if self.apply_integration_window:
+
+            m = np.arange(mmodes.mmax + 1)
+            w = np.sinc(m / nra)
+            inv_w = tools.invert_no_zero(w)
+
+            sl_vis = (slice(None),) + (np.newaxis,) * (len(mmodes.vis.shape) - 1)
+            mmodes.vis[:] *= w[sl_vis]
+
+            sl_weight = (slice(None),) + (np.newaxis,) * (len(mmodes.vis.shape) - 1)
```
sjforeman, line 654: Same comment as above
```suggestion
            sl_weight = (slice(None),) + (np.newaxis,) * (len(mmodes.weight.shape) - 1)
```

draco/synthesis/mockcatalog.py
```diff
@@ -1017,14 +1019,14 @@ def lrg_velocity_error(nsample, rng):
             Velocity errors in km / s.
         """
 
-        LRG_SIG = 91.8
+        LRG_SIG = 65.6
```
sjforeman, line 1022: Should also change the docstring - here's a suggestion:
```
This is taken from Ross et al. 2020 (https://arxiv.org/abs/2007.09000).
Figure 2 shows the distribution of redshift differences for repeated
observations of the same object; this is well fit by a Gaussian with 
width 92 km/s. They state that this corresponds to a Gaussian distribution 
for the single-measurement redshift errors, with width 65.6 km/s.
There is a bit of a tail that is not being captured in their Gaussian fit, and 
is hence not simulated in this routine.
```

sjforeman (approved):
Looks good - just a few suggestions!

PR #161, opened 2021-09-28
feat(util.tools.invert_no_zero): Handle subnormals.

tristpinsm:
Avoid overflow when subnormal floats are present.

I'm not sure if this is the desired behaviour for this function, but I couldn't think of a better place to handle this.

The reason I had to make this change is that I was running into overflows when trying to run the deconvolved ringmap pipeline (Tikhonov) using fitbeam. I think the issue was that fitbeam evaluates to very tiny values in some places, resulting in subnormal floats. This change has at least allowed the pipeline to get past that stage.


tristpinsm:
Yeah, I could see either returning 0 or float_max making sense. If you return float_max though it kind of suggests that you can invert the operation, but really you've lost the information that was in the subnormal float. And practically, that would require having two separate conditions to check (`== 0` and `< tol`).

PR #163, opened 2021-10-14
feat(synthesis.noise.GaussianNoiseDataset): Allow choice of dataset.

tristpinsm:


tristpinsm:
> The pattern we use almost everywhere else we deal with multiple possible container types with different datasets is for the task itself to figure it out. Typically we use an internal dict mapping from the type to the dataset name (e.g.
> 
> https://github.com/radiocosmology/draco/blob/20bb39875554ff3e73f309c1ab14a0533527a799/draco/analysis/delay.py#L225
> 
> ) so could you change this to do that?
> You may or may not want to leave in the option to override the dataset name.

Sure I can do that. Might be worth noting that `DelaySpectrumEstimatorBase` doesn't have that, which is what I modelled it on.

PR #164, opened 2021-10-15
New QSO z-error model

jrs65:
- feat(HealpixBeamForm): add an option to smooth the map before extraction
- feat(mockcatalog): add a new redshift dependent QSO z-error model

### DIFF
--- a/draco/analysis/beamform.py
+++ b/draco/analysis/beamform.py
@@ -1199,7 +1199,16 @@ class HealpixBeamForm(task.SingleTask):
     Unless it has an explicit `epoch` attribute, the Healpix map is assumed to be in the
     same coordinate epoch as the catalog. If it does, the input catalog is assumed to be
     in ICRS and then is precessed to the CIRS coordinates in the epoch of the map.
+
+    Attributes
+    ----------
+    fwhm : float
+        Smooth the map with a Gaussian with the specified FWHM in degrees. If `None`
+        (default), leave at native map resolution. This will modify the input map in
+        place.
     """
+
+    fwhm = config.Property(proptype=float, default=None)
 
     def setup(self, hpmap: containers.Map):
         """Set the map to extract beams from at each catalog location.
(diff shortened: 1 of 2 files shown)


draco/synthesis/mockcatalog.py
```diff
@@ -997,6 +997,60 @@ def qso_velocity_error(z, rng):
...
+
+        QSO_F_highz = 35.0
+        QSO_ztrans = 1.0
+        QSO_zwidth = 0.05
+
+        def smooth_step_function(z, zt, zw, fl, fh):
+            return (1 + np.tanh((z - zt) / zw)) * (fh - fl) / 2 + fl
+
+        def invfz(z):
+            return smooth_step_function(z, QSO_ztrans, QSO_zwidth, 0, 1 / QSO_F_highz)
```
sjforeman, line 1033: If I'm reading this right, it looks like the prefactor of the (1+tanh(...)) factor ends up being 1/70, whereas it is 1/35 in Eq. (131) of the stacking paper. I assume the code is correct and there is a typo in the paper?

draco/analysis/beamform.py
```diff
@@ -1199,6 +1208,13 @@ def setup(self, hpmap: containers.Map):
         """
 
         self.map = hpmap
+        mv = self.map.map[:]
+
+        self.log.info("Smoothing input Healpix map.")
+        for lfi, _ in mv.enumerate(axis=0):
+            mv[lfi, 0] = healpy.smoothing(
```
sjforeman, line 1215: Do we also want to loop over the pol axis here instead of just smoothing the zeroth pol, since in line 1273 we form beams for each pol present in the map container?

sjforeman (changes requested):

draco/analysis/beamform.py
jrs65: Yes. Not quite clear why it doesn't do that already.

draco/synthesis/mockcatalog.py
sjforeman: I'm saying that what is written in the code differs from Eq. (131) of the stacking paper by a factor of 2 in the prefactor

jrs65: I see. Let me check. Clearly the sims are done with what's in here. I'll need to check that the fit redshift error model matches that though....

Yes, we're good. `smooth_step_function` has the same definition in my comparison plots so it's just a typo in that equation.

jrs65: Maybe the one above it too. Will check.

draco/analysis/beamform.py
jrs65: This should be fixed now.

PR #167, opened 2021-10-26
doc(io.LoadFileFromParams): Remove note about index selection.

tristpinsm:
Distributed reads with index selection should now work.


jrs65:
I think this is now not going to work for zarr datasets, so I'll close this one and amend the message in the zarr PR.

PR #170, opened 2022-01-20
fix(SingleTask): pull tag from input_tags, when not explicitly configured

anjakefala:
Fixes bug introduced in bbbc6b4505e7065ca1c50f0784e11a8c7775dbe6. Input
tags stopped being referenced when setting the tag on the output.

Needed to tag delayspectrum files properly in the daily pipeline.

This reverts to previous behaviour. If this was behaviour that was intentionally removed, we will need to ensure that the tag is properly set on `delayspectrum` output containers.

### DIFF
--- a/draco/analysis/delay.py
+++ b/draco/analysis/delay.py
@@ -411,7 +411,9 @@ class DelaySpectrumEstimator(task.SingleTask, random.RandomTask):
         delays = np.fft.fftshift(np.fft.fftfreq(ndelay, d=self.freq_spacing))  # in us
 
         # Initialise the spectrum container
-        delay_spec = containers.DelaySpectrum(baseline=baselines, delay=delays)
+        delay_spec = containers.DelaySpectrum(
+            baseline=baselines, delay=delays, attrs_from=ss
+        )
         delay_spec.redistribute("baseline")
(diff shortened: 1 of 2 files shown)


jrs65:
There seems to be a few aspects to this one.

I was on the fence for a while about whether it should blindly copy the tag over from the input container. It previously did that, but it was all a bit of a hack as to how it worked, and so I had thought maybe it would be better for people to have to do it explicitly. But now I think I'm convinced now that it should do that, it's certainly better than the backup of just using the `count` as the tag.

The other aspect that confused me was why was this happening only in the DelaySpectrumEstimator. It looks like that's because that task does not use an `attrs_from` argument when constructing it's output container, which would have explicitly copied over the `tag`. Could you make that change, as it'll ensure both that tag is explicitly set, and that the other attributes are propagated?

jrs65 (changes requested):
This looks good, just a request to add the `attrs_from` argument (see my other comment).

anjakefala:
@jrs65 Made the requested change, and tested it without the `input_tags` change. `delayspectrum` files get tagged appropriately. =)

jrs65 (approved):

PR #171, opened 2022-02-10
fix(beamform): return before requiring self.epoch

anjakefala:
Needed along with c83f70577f631b70e6


jrs65:
Looks good, but I thought we'd already fixed this.

jrs65 (approved):

jrs65:
I guess this is a slightly different bug?!

anjakefala:
@jrs65 Yes :sob:

PR #176, opened 2022-02-21
style:

Arnab-half-blood-prince:
This will generate the delay spectrum at each RA and DEC using Wiener filter method. 
The DelaySpectrumWienerBase class in draco.analysis.delay.py will take ringmap and corresponding delay power spectrum as input and estimate the delay spectrum for all RA and DEC.

### DIFF
--- a/draco/analysis/delay.py
+++ b/draco/analysis/delay.py
@@ -676,6 +676,303 @@ class DelaySpectrumEstimatorBase(task.SingleTask, random.RandomTask):
         return delay_spec
 
 
+class DelaySpectrumWienerBase(task.SingleTask, random.RandomTask):
+    """Calculate the delay spectrum for each RA and DEC of any container with a frequency axis.
+
+    The spectrum is calculated by Wiener filtering method given the delay power spectrum at each DEC (or baselines) [https://arxiv.org/abs/2202.01242 see:- eqn A6]
+
+    The delay spectrum output is indexed by  `baseline` and 'RA' axis. The baseline axis is the
+    composite axis of all the axis (DEC, pol and beam axes) in the container except the RA and frequency axis.
+    These constituent axes are included in the index map,  which are ['beam', 'pol', 'ra', 'el'], and their
+    order is given by the `baseline_axes` attribute.
+
+    Attributes
+    ----------
+    delayps : str
+        The delay power spectrum for all baselines. This will be used in Wiener filter to generate the delay spectrum for all RA.
+
+    freq_zero : float, optional
+        The physical frequency (in MHz) of the *zero* channel. That is the DC
+        channel coming out of the F-engine. If not specified, use the first
+        frequency channel of the stream.
+    freq_spacing : float, optional
+        The spacing between the underlying channels (in MHz). This is conjugate
+        to the length of a frame of time samples that is transformed. If not
+        set, then use the smallest gap found between channels in the dataset.
+    nfreq : int, optional
+        The number of frequency channels in the full set produced by the
+        F-engine. If not set, assume the last included frequency is the last of
+        the full set (or is the penultimate if `skip_nyquist` is set).
+    skip_nyquist : bool, optional
+        Whether the Nyquist frequency is included in the data. This is `True` by
+        default to align with the output of CASPER PFBs.
+    dataset : str
+        Calculate the delay spectrum of this dataset (e.g., "vis", "map", "beam").
+    average_axis : str
+        Name of the axis to take the average over.
+    """
+
+    delayps = config.Property(proptype=str, default=None)
+    freq_zero = config.Property(proptype=float, default=None)
+    freq_spacing = config.Property(proptype=float, default=None)
+    nfreq = config.Property(proptype=int, default=None)
+    skip_nyquist = config.Property(proptype=bool, default=True)
+
+    dataset = config.Property(proptype=str, default="vis")
+    average_axis = config.Property(proptype=str)
+
+    def setup(self, telescope: io.TelescopeConvertible):
+        """Set the telescope needed to generate Stokes I.
+
+        Parameters
+        ----------
+        telescope : TransitTelescope
+        """
+        self.telescope = io.get_telescope(telescope)
+
+    def process(self, ss: FreqContainerType) -> containers.delayspec:
+        """Estimate the delay spectrum.
+
+        Parameters
+        ----------
+        ss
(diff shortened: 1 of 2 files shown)


ssiegelx (commented):
This looks good Arnab.  I have a bunch of comments, but most are just explaining our coding conventions.  The two substantial comments are about line 896 of delay.py, which I think would actually cause errors in the results.

I will likely have a few additional comments once these are addressed.  I am hoping we can come up with a way of defining the DelaySpectrum containers so that we don't have to flatten all of the axes into a single baseline axis, as I think reshaping these datasets is going to be annoying if we are using them frequently.  But I need to think about this a bit more before proposing something.

draco/analysis/delay.py
```diff
@@ -718,6 +718,212 @@ def process(self, ss: FreqContainerType) -> containers.DelaySpectrum:
...
+    """Calculate the delay spectrum for each RA and DEC of any container with a frequency axis.
+
+    The spectrum is calculated by Wiener filtering method given the delay power spectrum
+    at each DEC/baselines [https://arxiv.org/abs/2202.01242 see:- eqn A6].
+
+    The delay spectrum output is indexed by  `baseline` and 'sample' axis. The baseline axis is the
+    composite axis of all the axis (DEC, pol and beam axes) in the container except the 'sample' and frequency axis.
+    The 'sample' axis is the average axis over which the average is taken to estimate the delay power spectrum.
+    These constituent axes are included in the index map,  which are ['beam', 'pol', 'sample', 'el'], and their
+    order is given by the `baseline_axes` attribute.
```
sjforeman, line 731: Try to keep docstring lines under 75 characters (see https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard)

draco/analysis/delay.py
```diff
@@ -718,6 +718,212 @@ def process(self, ss: FreqContainerType) -> containers.DelaySpectrum:
...
+        Calculate the delay spectrum of this dataset (e.g., "vis", "map", "beam").
+    average_axis : str
+        Name of the axis to take the average over.
+
+    apply_window = config.Property(proptype=bool, default=True)
+
+    window = config.enum(
+        ["nuttall", "blackman_nuttall", "blackman_harris"], default="nuttall"
+    )
+
```
sjforeman, line 746: Update to have correct docstring formatting

draco/analysis/delay.py
```diff
@@ -718,6 +718,212 @@ def process(self, ss: FreqContainerType) -> containers.DelaySpectrum:
...
+    complex_timedomain : bool, optional
+        Whether to assume the original time samples that were channelized into a
+        frequency spectrum were purely real (False) or complex (True). If True,
+        `freq_zero`, `nfreq`, and `skip_nyquist` are ignored. Default: False.
+
+    """
+
+    dataset = config.Property(proptype=str, default="vis")
+    average_axis = config.Property(proptype=str)
+    # complex_input = config.Property(proptype=bool, default=False)
```
sjforeman, line 756: Remove commented out line

draco/analysis/delay.py
```diff
@@ -718,6 +718,212 @@ def process(self, ss: FreqContainerType) -> containers.DelaySpectrum:
         return delay_spec
 
 
+class DelaySpectrumWienerBase(task.SingleTask):
+    """Calculate the delay spectrum for each RA and DEC of any container with a frequency axis.
```
sjforeman, line 722: This description isn't quite correct: in principle this task could operate on visibilities, in which case the input container would not have a dec axis

draco/analysis/delay.py
```diff
@@ -718,6 +718,212 @@ def process(self, ss: FreqContainerType) -> containers.DelaySpectrum:
...
+    These constituent axes are included in the index map,  which are ['beam', 'pol', 'sample', 'el'], and their
+    order is given by the `baseline_axes` attribute.
+
+    Attributes
+    ----------
+
+    dataset : str
+        Calculate the delay spectrum of this dataset (e.g., "vis", "map", "beam").
+    average_axis : str
+        Name of the axis to take the average over.
```
sjforeman, line 739: Unlike in the Gibbs delay power spectrum estimator, no average is taken in this task, so a different name should probably be used here

draco/analysis/delay.py
```diff
@@ -718,6 +718,212 @@ def process(self, ss: FreqContainerType) -> containers.DelaySpectrum:
...
+        # ==== Figure out the frequency structure and delay values ====
+
+        self.freq_zero = ss.freq[0]
+
+        self.freq_spacing = np.abs(np.diff(ss.freq[:])).min()
+
+        if self.complex_timedomain:
+            self.nfreq = len(ss.freq)
+            channel_ind = np.arange(self.nfreq)
+        # ndelay = self.nfreq
```
sjforeman, line 812: Remove commented-out line

draco/analysis/delay.py
```diff
@@ -1171,6 +1377,115 @@ def _draw_ps_sample(d):
...
+
+    complex_timedomain : bool, optional
+        If True, assume input data arose from a complex timestream. If False, assume
+        input data arose from a real timestream, such that the first and last frequency
+        channels have purely real values. Default: False.
+
+    Returns
+    -------
+    y_spec : np.ndarray[nRA,ndelay]
+            The 2D array containing the delay spectrum for each RA.
```
sjforeman, line 1416: Make sure whitespace is consistent with formatting guidelines

sjforeman (commented):
Hi Arnab - this functionality will be great to have, so thanks for working on it!

My specific comments in this review are mostly about formatting and clarity. Seth has also raised a good point about whether to continue with the scheme where all axes except frequency and one other one are compressed into a single axis. Here is a possible solution:
- Have separate `DelaySpectrum` and `DelayTransform` containers for each input container we want to consider (e.g. `SiderealStreamDelaySpectrum`, `RingMapDelaySpectrum`)
- Refactor `DelaySpectrumEstimatorBase` and `DelaySpectrumWienerBase` such that there are derived classes (e.g. `DelaySpectrumEstimatorSiderealStream` - unfortunately getting long-winded) that specify which axis is averaged over, but with the majority of the code contained in the base classes.

Since there are only a few possibilities for input containers into these estimators, it would only add a moderate amount of code to split the tasks up this way, but would make the output delay spectra easier to work with.

Separately, there is also a lot of repeated code between `DelaySpectrumEstimatorBase` and `DelaySpectrumWienerBase`, and it would be worth thinking about how to reduce some of that redundancy.

PR #177, opened 2022-03-21
feat(delay): allow user to specify apodisation window and allow for complex-valued delay spectrum

sjforeman:


draco/analysis/delay.py
```diff
@@ -846,8 +889,56 @@ def fourier_matrix_c2r(N, fsel=None):
     return Fr
 
 
+def fourier_matrix_c2c(N, fsel=None):
+    """Generate a Fourier matrix to represent a complex to complex FFT.
```
jrs65, line 893: This seems

draco/analysis/delay.py
```diff
@@ -846,8 +889,56 @@ def fourier_matrix_c2r(N, fsel=None):
...
+    N : integer
+        Length of timestream that we are transforming to.
+    fsel : array_like, optional
+        Indices of the frequency channels to include in the transformation
+        matrix. By default, assume all channels.
+
+    Returns
+    -------
+    F : np.ndarray
+        An array performing the Fourier transform from a complex time series to
```
jrs65, line 906: This seems subtle, and certainly isn't true as written.

I can take an arbitrary complex time series, apply this matrix to it, and I will get the usual *complex* Fourier transform frequencies inserted into the `[0::2]` elements of the output vector, and `1j` times them in the `[1::2]` elements.

If the matrix here was being used to define the transpose (i.e. mapping from real-valued, real and imaginary frequency components to a complex time stream) this would be correct, but it's not.

I really think that the elements in `F[0::2]` are all thats needed to represent this, but it would require changes in the way that `delay_spectrum_gibbs` is implemented. Alternatively, you could have this mapping from real-imag alternating floats on both sides of the matrix (which is probably what you'd get from `F.view(np.float64)`).

draco/analysis/delay.py
```diff
@@ -887,40 +982,44 @@ def delay_spectrum_gibbs(
...
+    # spectrum is assumed to be real)
     Ni_r = np.zeros(2 * Ni.shape[0])
     Ni_r[0::2] = np.where(is_real_freq, Ni, Ni / 2**0.5)
     Ni_r[1::2] = np.where(is_real_freq, 0.0, Ni / 2**0.5)
 
     # Create the Hermitian conjugate weighted by the noise (this is used multiple times)
-    FTNih = F.T * Ni_r[np.newaxis, :] ** 0.5
-    FTNiF = np.dot(FTNih, FTNih.T)
+    FTNih = F.T.conj() * Ni_r[np.newaxis, :] ** 0.5
+    FTNiF = np.dot(FTNih, FTNih.T.conj())
```
jrs65, line 1022: Given the comments above, I'm not 100% sure that this is actually what we want to be doing in here.

jrs65 (changes requested):

draco/analysis/delay.py
sjforeman: > If the matrix here was being used to define the transpose (i.e. mapping from real-valued, real and imaginary frequency components to a complex time stream) this would be correct, but it's not.

I wrote this routine such that the transpose of its output is correct, since we only ever use the transpose. (It always struck me as strange that `fourier_matrix_c2r()` was defined in such a way that its transpose needs to be taken to get the result we want, but I wrote `fourier_matrix_c2c()` to match this convention.) I remember checking the outputs against the numpy ifft routine (which motivated the comments that describe the comparison), so I believe the transpose of `fourier_matrix_c2c()` is correct.

jrs65: The rationale is that as it's a classic inverse problem, you first define the forward model, which is given by F. You do kind of use it within the $\mathbf{F}^\dagger \mathbf{N}^{-1} \mathbf{F}$ term.

I'll try and think about it all today, and get a coherent picture in my head. I think correctly expressing the forward model in the way we've written requires using a matrix like
```math
\begin{pmatrix}
\mathbf{F} / 2 & \mathbf{F}^* / 2 \\
-i \mathbf{F} / 2 & -i \mathbf{F}^* / 2
\end{pmatrix}
\begin{pmatrix}
\mathbf{t} \\
\mathbf{t}^*
\end{pmatrix}
```
with $\mathbf{F}_{ab} = e^{2 \pi i a b / N}$, which forces you back into the full statistics of complex Gaussian numbers (i.e. needing both correlation and relation matrices).

jrs65: Hmmm.... I think Github's new math rendering is meant to have rendered the block above, but it's only done the inline parts. Not quite sure what's going on with that.

jrs65: Made it work 🥳

sjforeman:
@jrs65 Thanks again for your feedback. I've gone through and attempted to re-implement the c2c transform more carefully, using the alternating real/imaginary representation in both sides of the transform.

The plots below are the results of a test similar to what you showed in the appendix of the stacking paper, with half of the 1024 frequencies randomly masked. The c2c version is noisier in the low-power region, but otherwise the performance matches the r2c transform.

r2c_test

c2c_test

jrs65:
It seems quite weird that it's so spiky in the noise dominated region. The power spectrum seems to be going down at least two orders of magnitude lower in certain locations. I wouldn't have expected there to be that much difference between that and the c2r version.

How do the results compare to the old c2c version?

draco/analysis/delay.py
```diff
@@ -846,34 +889,126 @@ def fourier_matrix_c2r(N, fsel=None):
...
+
+    Returns
+    -------
+    out : array_like
+        Output array of alternating real and imaginary components. These components are
+        expanded along the last axis, such that if `array` has `N` complex elements in
+        its last axis, `out` will have `2N` real elements.
+    """
+
+    return array.astype(np.complex128, order="C").view(np.float64).copy()
```
jrs65, line 948: Is it necessary to copy this here, rather than just returning the view. 

If you want to return a copy, I'd change the first line of the docstring to avoid the word view.

draco/analysis/delay.py
```diff
@@ -846,34 +889,126 @@ def fourier_matrix_c2r(N, fsel=None):
...
+
+    Returns
+    -------
+    out : array_like
+        Output array of complex numbers, derived from compressing the last axis (if
+        `array` has `N` real elements in the last axis, `out` will have `N/2` complex
+        elements).
+    """
+
+    return (array[..., 0::2] + 1.0j * array[..., 1::2]).astype(np.complex128)
```
jrs65, line 967: Similar to the comment above, you could actually generate a view in here. I think `array.astype(np.float64, order="C").view(np.complex128)` would work (potentially with a reshape).

draco/analysis/delay.py
```diff
@@ -846,34 +889,126 @@ def fourier_matrix_c2r(N, fsel=None):
...
     fsel : np.ndarray[freq], optional
         Indices of channels that we have data at. By default assume all channels.
     niter : int, optional
         Number of Gibbs samples to generate.
     rng : np.random.Generator, optional
         A generator to use to produce the random samples.
+    complex_input : bool, optional
+        If True, assume input data arose from a complex timestream. If False, assume
+        input data arose from a real timestream, such that the first and last frequency
+        channels have purely real values. Default: False.
```
jrs65, line 1011: I find the name of this parameter a bit confusing, as I keep thinking it refers to the task input, but it really refers to some abstract input that we never see. There's probably a more descriptive name. `complex_timedomain`?

draco/analysis/delay.py
```diff
@@ -999,7 +1156,14 @@ def _draw_ps_sample(d):
...
 
-        spec.append(S_samp)
+        # If we're dealing with a complex delay spectrum, S_samp will be the power
+        # spectrum of the (alternating) real and imaginary delay spectrum components,
+        # so we need to compress S_samp down to the power spectrum of the complex
+        # elements
+        if complex_input:
+            spec.append(S_samp[0::2] + S_samp[1::2])
+        else:
+            spec.append(S_samp)
```
jrs65, line 1166: I am a bit unclear about whether this is quite what we want. $\DeclareMathOperator{\var}{var}$

We are calculating delay power spectra, and the underlying assumption is that the underlying statistics are circularly symmetric complex Gaussian (so $D = \var(d_\text{real}) + \var(d_\text{imag})$ ), and that's what we output in the end. However, in the middle we calculate the separate real and imaginary PS samples, and then feed that back into the drawing of the signal sample. That means that the delay noise perturbation which are added into the signal sample do not have circularly symmetric statistics.

Ultimately I'm trying to understand why the output seems "noisier", and while I don't quite know what effect this is going to have with respect to that, it's a bit weird.

jrs65 (changes requested):

sjforeman:
@jrs65 It turned out that the noisy results in the low-power region were due to an error in how the separate real and imaginary parts of the delay spectrum were sampled; for samples that happened to be low, this error drove those samples to zero in subsequent Gibbs iterations. I've fixed that, and have addressed your other comments as well.

Below are some plots that demonstrate good agreement between the r2c and c2c transforms:
spec_r2c
spec_c2c
lowpower_comparison
highpower_comparison

jrs65:
These changes look great. Thanks @sjforeman . I'm going to rebase and tidy up the commits and then merge.

jrs65 (approved):

PR #178, opened 2022-06-03
fix(invert_no_zero): issue when operating on MPIArrays

jrs65:
I think the fix turns out to be trivial in here, as all the new MPIArray code means that `.empty_like` does exactly what you would want it to.


tristpinsm (approved):
good that there is a simple solution!

jrs65:
Actually I don't think I want a ravel in there as it's more likely to make a copy.

PR #179, opened 2022-06-30
fix: resolve MPIArray related warnings

jrs65:
This is one of a set of PRs to resolve the warnings spat out by the daily pipeline with the new MPIArray

### DIFF
--- a/draco/analysis/flagging.py
+++ b/draco/analysis/flagging.py
@@ -592,18 +592,21 @@ class SmoothVisWeight(task.SingleTask):
         weight = data.weight[:]
         # Data will be distributed in frequency.
         # So a frequency loop will not be too large.
+
+        weight_local = weight.local_array
+
         for lfi, gfi in weight.enumerate(axis=0):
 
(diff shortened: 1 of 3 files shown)


draco/analysis/flagging.py
```diff
@@ -592,18 +592,21 @@ def process(self, data):
         weight = data.weight[:]
         # Data will be distributed in frequency.
         # So a frequency loop will not be too large.
+
+        weight_local = weight.local_array
+
         for lfi, gfi in weight.enumerate(axis=0):
 
             # MPIArray takes the local index, returns a local np.ndarray
```
tristpinsm, line 600: I guess this is no longer true

tristpinsm (approved):
Presumably there are many other instances that should be changed to use `.local_array`. It would be tedious to go through and find them all, but should we have a brief document explaining the "best practice" for this common pattern, and advertise it? Then maybe people can fix them as they are encountered. Or maybe I'm overestimating how frequent this will be.

jrs65:
This was basically merged in #180

PR #183, opened 2022-07-21
fix(containers): fixes bug setting compression parameters for copy

ssiegelx:
It looks like the copy was accidentally swapped with the original container when setting the compression parameters.


jrs65 (approved):
Good catch!

PR #185, opened 2022-07-28
Ljg/threashold vis weight

ljgray:
Updates `ThresholdVisWeight` to be baseline-independent, returning a `RFIMask` container. 

The mask is created by taking a median over all baselines, clipping any values below an absolute threshold, then taking a median in time/ra to get a single threshold for each frequency. This threshold is evaluated relative to the median of all baselines to get a 2D mask in frequency+ra/time - i.e., the same mask would be applied to all baselines. 

I'm not fully familiar with how the re-gridding is done but I'd imagine that masking this way could let outliers slip through, since the mask is relative to the median baseline. Alternatively, we could mask all baselines individually and then stack those masks so that the only ra/frequency points allowed through are those that aren't above the threshold in any baseline, but that could be overly aggressive.

### DIFF
--- a/draco/analysis/flagging.py
+++ b/draco/analysis/flagging.py
@@ -5,6 +5,7 @@ data; and pre-map making flagging on m-modes.
 """
 
 from typing import Union
+import warnings
 import numpy as np
 import scipy.signal
 from scipy.ndimage import median_filter
@@ -659,8 +660,11 @@ class ThresholdVisWeight(task.SingleTask):
             mean_baseline > self.absolute_threshold, mean_baseline, np.nan
         )
         # Average across the time (ra) axis to get per-frequency thresholds,
-        # ignoring any nans
-        threshold = np.nanmean(threshold, axis=2, keepdims=True)
+        # ignoring any nans. np.nanmean will give a warning if an entire band is
+        # nan, which we expect to happen in some cases.
+        with warnings.catch_warnings():
+            warnings.filterwarnings(action="ignore", message="Mean of empty slice")
+            threshold = np.nanmean(threshold, axis=2, keepdims=True)
         # Create a 2D baseline-independent mask.
         mask = ~(
             mean_baseline.local_array
--- a/draco/analysis/flagging.py
+++ b/draco/analysis/flagging.py
@@ -612,11 +612,10 @@ class SmoothVisWeight(task.SingleTask):
 
 
 class ThresholdVisWeight(task.SingleTask):
-    """Set any weight less than the user specified threshold equal to zero.
+    """Create a mask to remove all weights below a per-frequency threshold.
 
-    Threshold is determined as `maximum(absolute_threshold,
-    relative_threshold * mean(weight))` and is evaluated per product/stack
-    entry.
+    A single relative threshold is set for each frequency along with an absolute
+    minimum weight threshold. Masking is done relative to the mean baseline.
 
     Parameters
     ----------
@@ -628,10 +627,10 @@ class ThresholdVisWeight(task.SingleTask):
     """
 
     absolute_threshold = config.Property(proptype=float, default=1e-7)
-    relative_threshold = config.Property(proptype=float, default=0.0)
+    relative_threshold = config.Property(proptype=float, default=0.9)
 
-    def process(self, timestream):
-        """Apply threshold to `weight` dataset.
+    def process(self, stream):
+        """Make a baseline-independent mask.
 
         Parameters
         ----------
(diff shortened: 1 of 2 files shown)


ljgray:
> I'm not fully familiar with how the re-gridding is done but I'd imagine that masking this way could let outliers slip through, since the mask is relative to the median baseline. Alternatively, we could mask all baselines individually and then stack those masks so that the only ra/frequency points allowed through are those that aren't above the threshold in any baseline, but that could be overly aggressive.

Updated to use the latter method as some testing found that the fraction of masked weights is comparable.

ljgray:
Using mean rather than median to get around radiocosmology/caput#205. Doesn't seem to have much of an effect on the result. Should be ready for review.

ljgray:
The following must be added to the pipeline job template, between `SiderealRegridder` and `RingMapMaker`
 
```yaml
    # Flag out low weight samples to remove transient RFI artifacts at the edges of
    # flagged regions
    - type: draco.analysis.flagging.ThresholdVisWeight
      in: sstream
      out: weight_mask
      params:
          relative_threshold: 0.5

    # Apply the RFI mask. This will modify the data in place.
    - type: draco.analysis.flagging.ApplyRFIMask
      in: [sstream, weight_mask]
      out: sstream_mask
```

ljgray:
This Works<sup>TM</sup> in that the pipeline runs without crashing or throwing any warnings.

ljgray:
With the implementation here, the following issues are addressed:
- flagging is independent of baseline
- masking is done per-frequency, accounting for different weights across frequencies
- returns a mask container. This has been tested in a pipeline run
It seems to do well at flagging gaps caused by RFI without removing anything else. I'm not yet convinced that it's able to catch the edge regions properly

draco/core/io.py
```diff
@@ -111,7 +111,7 @@ def _list_or_glob(files: Union[str, List[str]]) -> List[str]:
     return files
 
 
-def _list_of_filegroups(groups: Union[List[Dict] or Dict]) -> List[Dict]:
+def _list_of_filegroups(groups: Union[List[Dict], Dict]) -> List[Dict]:
```
jrs65, line 114: Weird. I wonder how that got in.

draco/analysis/flagging.py
```diff
@@ -1121,8 +1131,11 @@ def process(self, tstream, rfimask):
         else:  # self.share == "none"
             tsc = tstream.copy()
 
-        # Mask the data
-        tsc.weight[:] *= (~rfimask.mask[sf:ef][bcast_slice]).astype(np.float32)
+        # Mask the data. This will throw an mpiarray warning because the mask being
+        # applied is a numpy array matching the local array directly
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore")
+            tsc.weight[:] *= (~rfimask.mask[sf:ef][bcast_slice]).astype(np.float32)
```
jrs65, line 1138: I'm on the fence about this one. It might be better to create an MPIArray with the mask and then apply directly. Turning off warnings seems a bit heavy handed as maybe some other part of the code could throw a warning in here.

jrs65 (changes requested):

draco/analysis/flagging.py
ljgray: Maybe an optional argument to cast the mask to an MPIarray? I agree that it might be a bit much to ignore the warning

jrs65: It could be reasonable to add a method to `MemDatasetCommon` which returns the dataset as an MPIArray distributed along a given axis. Or a classmethod on `MPIArray` which given a common numpy array, builds a distributed MPIArray from it.

draco/analysis/flagging.py
```diff
@@ -628,49 +628,59 @@ class ThresholdVisWeight(task.SingleTask):
...
-
-        timestream.comm.Allreduce(keep_sum, keep_total, op=MPI.SUM)
-        keep_frac = keep_total / float(np.prod(weight.global_shape))
-
-        self.log.info(
-            "%0.5f%% of data is below the weight threshold"
-            % (100.0 * (1.0 - keep_frac))
+        # Average across the time (ra) axis to get per-frequency thresholds,
+        # interpolating any spots where the entire band is nan
+        threshold = np.mean(threshold, axis=2, keepdims=True)
```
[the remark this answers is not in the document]
ljgray: Good catch

draco/analysis/flagging.py
```diff
@@ -628,49 +628,59 @@ class ThresholdVisWeight(task.SingleTask):
...
+        nans, fx = np.isnan(threshold), lambda z: z.nonzero()[0]
+        threshold[nans] = np.interp(fx(nans), fx(~nans), threshold[~nans])
```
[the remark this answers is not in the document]
ljgray: `np.nanmean` will still return `nan` for a frequency band which is entirely nans, which would happen if the entire band failed the absolute threshold check. the comparison 
```python
mean_baseline < threshold * self.relative_threshold
``` 
will return False anywhere where the threshold is nan, since anything compared to nan is False. A cheaper method than interpolation would be to fill the nans in `threshold` with something like double the minimum threshold, but those nans would still have to be filled.

draco/analysis/flagging.py
ljgray: Is there a specific reason behind not making the draco RFImask distributed?

jrs65: I think it's too small to make it worth the overhead. I agree that on reflection it doesn't seem like the smartest idea, we could consider changing it but I think we need to figure out exactly where it could cause problems downstream first.

ljgray: It might be worth just letting the warning get logged and leaving it as-is. Since `MemDatasetDistributed` exists, it doesn't really make sense to add anything to it related to mpi. This comparison of an mpiarray to a np.ndarray is the intended behaviour here, and it should always work, so maybe it isn't worth trying to hack around it

draco/analysis/flagging.py
```diff
@@ -628,49 +627,56 @@ class ThresholdVisWeight(task.SingleTask):
...
-
-        self.log.info(
-            "%0.5f%% of data is below the weight threshold"
-            % (100.0 * (1.0 - keep_frac))
+        # Average across the time (ra) axis to get per-frequency thresholds,
+        # ignoring any nans
+        threshold = np.nanmean(threshold, axis=2, keepdims=True)
+        # Create a 2D baseline-independent mask.
+        mask = ~np.squeeze(
+            mean_baseline.local_array > threshold * self.relative_threshold
```
jrs65, line 666: I don't think the criterion actually works as written.

Take a single frequency (after baseline averaging) with two values, one a very small value epsilon below the absolute threshold and one epsilon above, i.e. `[t - eps, t + eps]`. You end up with threshold on (L658) becoming `[NaN, t + eps]`, then on L663 after the nanmean becoming `[t + eps]`. A typical value of relative_threshold might be ~0.5, and provided `eps < t * 0.5` *neither* of the times within that one frequency is masked.

I think what you need to do is either generate masks for both relative and absolute thresholds and then combine, or  I think you can just do `mean_baseline.local_array > np.fmax(threshold * self.relative_threshold, self.absolute_threshold)`

jrs65 (changes requested):

jrs65 (changes requested):

draco/analysis/flagging.py
```diff
@@ -628,49 +627,57 @@ class ThresholdVisWeight(task.SingleTask):
...
-        keep_frac = keep_total / float(np.prod(weight.global_shape))
-
-        self.log.info(
-            "%0.5f%% of data is below the weight threshold"
-            % (100.0 * (1.0 - keep_frac))
+        # Average across the time (ra) axis to get per-frequency thresholds,
+        # ignoring any nans
+        threshold = np.nanmean(threshold, axis=2, keepdims=True)
+        # Create a 2D baseline-independent mask.
+        mask = ~np.squeeze(
```
jrs65, line 665: I thought I'd added this comment earlier, but maybe I forgot to click. Apologies if this is the second time you get it.

I think it's best to avoid the squeeze in here. If you get an input array which has only one frequency it would squeeze out that one too, which would cause problem. As you know explicitly which axes have been manipulated you can do the selection explicitly (probably just needs `[:, 0, :]`).

jrs65 (changes requested):

draco/analysis/flagging.py
```diff
@@ -628,49 +628,60 @@ class ThresholdVisWeight(task.SingleTask):
...
+            > np.fmax(threshold * self.relative_threshold, self.absolute_threshold)
+        )[:, 0, :]
+        # Collect all parts of the mask. Method .allgather() returns a np.ndarray
+        mask = mpiarray.MPIArray.wrap(mask, axis=0).allgather()
+        if self.comm.rank == 0:
+            # Log the percent of data masked
+            drop_frac = np.sum(mask) / np.prod(mask.shape)
+            self.log.info(
+                "%0.5f%% of data is below the weight threshold" % (100.0 * (drop_frac))
+            )
```
jrs65, line 680: I think in here you should remove the `self.comm.rank == 0` conditional, and output on all ranks. We generally favour making log messages output on all ranks, and then controlling what gets put in the logs with the rank dependent level cutoffs (you'll see that in the configs in `SetMPILogging` with the `level_rank0` etc parameters).

jrs65 (approved):
One last comment to fix. I've approved this an advance, so please just rebase + merge when you've fixed it.

PR #186, opened 2022-08-15
fix(ApplyRFIMask): don't apply mask to local array.

ljgray:
Trying to apply the mask to the .local_array results in an
AttribteError. This should fix the issue, although an mpiarray warning
will occur. This reverts a change made in #185

### DIFF
--- a/draco/analysis/flagging.py
+++ b/draco/analysis/flagging.py
@@ -1131,9 +1131,8 @@ class ApplyRFIMask(task.SingleTask):
         else:  # self.share == "none"
             tsc = tstream.copy()
 
-        # Mask the data. This will throw an mpiarray warning because the mask being
-        # applied is a numpy array matching the local array directly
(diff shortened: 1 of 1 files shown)


jrs65:
Can you explain why the `AttributeError` happens?

jrs65:
I don't understand it at all. If the result of `tsc.weight[:]` isn't an MPIArray then none of the code above it should work.

ljgray:
```python
AttributeError: can't set attribute 'local_array'
```

Looking at `MPIArray`, there doesn't appear to be a setter for `.local_array`, so I think that's what throwing the `AttributeError`.

jrs65:
Oh, then I think you just need to do `.local_array[:] *=`

ljgray:
PR should be updated for that change.

jrs65 (approved):

jrs65:
Does that change actually work? I think it might need to be `tsc.weight[:].local_array[:]`. The first slice is needed to get the MPIArray out, and then second to avoid the missing local_array setter.

PR #187, opened 2022-08-18
perf(SmoothVisWeight): Use caput `moving_weighted_median`

ljgray:
~~The previous implementation looped over each frequency individually.
This can be accomplished by setting the window size to 1 in the median
filter, meaning that we only need one median filter call. The filter can
also be done in-place, removing an extra memory allocation (this
probably won't change much since it's a fairly small array).~~
Looping is still done and we instead use `caput.moving_weighted_median` for speed

### DIFF
--- a/draco/analysis/flagging.py
+++ b/draco/analysis/flagging.py
@@ -11,7 +11,6 @@ from typing import Union, overload
 import warnings
 import numpy as np
 import scipy.signal
-from scipy.ndimage import median_filter
 
 from caput import config, weighted_median, mpiarray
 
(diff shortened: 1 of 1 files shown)


jrs65 (approved):
I'm happy to merge this as it is a cleaner, although in somes ways it seems like a sideways move as probably uses more memory. It constructs the full sized boolean array `zeromask` rather than keep it per frequency, so if there's more 4 frequencies per node this is probably more expensive memory wise (zeromask is all frequencies, 1 byte per element, versus the previous version with the intermediate 4 byte per element temporary).

ljgray:
I'm going to hold off on merging for the moment. The change I made is cleaner, but in testing there are no noticeable performance changes. This is actually quite slow overall, and is logically easily to do in parallel, so I would like to see if I can tweak it to make use of all available cores.

jrs65:
@ljgray you probably don't need to worry about replicating the edge behaviour of the original code. I think the scheme used by the `moving_weighted_median` code (basically truncating the window at the edges) is reasonable and unbiased, and on top of that the time range pushed through the daily pipeline is deliberately padded to ensure that we don't have edges issues affecting the data we care about.

ljgray:
> @ljgray you probably don't need to worry about replicating the edge behaviour of the original code. I think the scheme used by the `moving_weighted_median` code (basically truncating the window at the edges) is reasonable and unbiased, and on top of that the time range pushed through the daily pipeline is deliberately padded to ensure that we don't have edges issues affecting the data we care about.

Ah, great.

draco/analysis/flagging.py
```diff
@@ -606,27 +605,25 @@ def process(self, data):
...
-            # Median filter. Mode='nearest' to prevent steps close to
-            # the end from being washed
-            weight_local[lfi] = median_filter(
-                weight_local[lfi], size=(1, self.kernel_size), mode="nearest"
+            zeromask = weight_local[i] == 0.0
+
+            weight_local[i] = weighted_median.moving_weighted_median(
+                data=weight_local[i],
+                weights=np.ones_like(weight_local[i]),
+                size=(1, self.kernel_size),
```
ljgray, line 621: Not sure if there's a better/more efficient way to make the weights. Maybe setting dtype to int instead of float

ljgray:
@jrs65 with the changes made in radiocosmology/caput#220, we get about a 3x performance increase with this task in the daily pipeline, and it lets us further improve the `moving_weighted_median` independently

draco/analysis/flagging.py
```diff
@@ -606,27 +605,25 @@ def process(self, data):
...
-            zeromask = weight_local[lfi] == 0.0
-            # Median filter. Mode='nearest' to prevent steps close to
-            # the end from being washed
-            weight_local[lfi] = median_filter(
-                weight_local[lfi], size=(1, self.kernel_size), mode="nearest"
+            zeromask = weight_local[i] == 0.0
+
+            weight_local[i] = weighted_median.moving_weighted_median(
+                data=weight_local[i],
+                weights=np.ones_like(weight_local[i], dtype=int),
```
jrs65, line 620: Actually in here I think you want to take advantage of the fact it's a weighted median and weight the zero entries to zero. So it wants to be something like (
```suggestion
                weights=(weights_local[i] > 0).astype(int),
```

jrs65 (changes requested):

draco/analysis/flagging.py
jrs65: This is obviously not what the original code does, but the input code has the issue of being biased by long regions of zero weights.

PR #188, opened 2022-08-20
convert ch_pipeline ringmap (daily) to draco ringmap container

Arnab-half-blood-prince:
Covert the daily (CSD)  ringmap (output of ch_pipeline) to draco ringmap. The daily ringmap of ch_pipeline has rms axis instead of weight, so we are changing rms to weight and storing it in a new draco ringmap container.

### DIFF
--- a/draco/analysis/delay.py
+++ b/draco/analysis/delay.py
@@ -10,7 +10,7 @@ from caput import mpiarray, config
 from cora.util import units
 
 from ..core import containers, task, io
-from ..util import random
+from ..util import random, tools
 
 
 class ConvertRingmap(task.SingleTask):
@@ -45,7 +45,7 @@ class ConvertRingmap(task.SingleTask):
         rmap_out.redistribute("freq")
 
         rmap_out.map[:] = rmap_in.map[:]
-        rmap_out.weight[:] = invert_no_zero(rmap_in.rms[:][..., np.newaxis] ** 2)
+        rmap_out.weight[:] = tools.invert_no_zero(rmap_in.rms[:][..., np.newaxis] ** 2)
 
         if "dirty_beam" in rmap_in.datasets:
             rmap_out.add_dataset("dirty_beam")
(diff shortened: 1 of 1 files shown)


draco/analysis/delay.py
```diff
@@ -10,7 +10,48 @@
...
 
 from ..core import containers, task, io
-from ..util import random
+from ..util import random, tools
+
+
+class ConvertRingmap(task.SingleTask):
+    """Covert the  daily (CSD)  ringmap (output of ch_pipeline) to draco ringmap.
+    The daily ringmap of ch_pipeline has rms axis instead of weight, so we
+    are changing rms to weight and store it in a new draco ringmap container.
```
ssiegelx, line 19:
```suggestion
    """Covert a ch_pipeline RingMap to a draco RingMap.
    
    The ch_pipeline RingMap container has an rms dataset instead of
    a weight dataset.  This task converts rms to weight and stores the
    result in a draco RingMap container.
```
Docstrings should conform to the numpy style guide.  The first line should be a summary of the task.  Then there should be a blank line.  Then you can have an (optional) paragraph that provides additional details.  Also, I would remove references to the daily pipeline and CSD, because this task is more generic and can be used to convert any ringmap generated by ch_pipeline into a format that can be processed by draco tasks.

draco/analysis/delay.py
```diff
@@ -10,7 +10,48 @@
...
+    """
+
+    def process(self, rmap_in):
+        """converting daily ringmap to draco ringmap.
```
ssiegelx, line 23:
```suggestion
        """Convert a ch_pipline RingMap to a draco RingMap.
```

ssiegelx (changes requested):
This looks good @Arnab-half-blood-prince.   I only have a few comments related to the docstrings.

However, I think `draco.analysis.delay` is not the right place for this task.  I would move it  to the bottom of `ch_pipeline.analysis.ringmap` instead.  Could you close this pull request and create a new pull request in ch_pipeline that contains this code?  There are two comments related to importing the appropriate container from draco that you will also need to implement in order to get the task to work in ch_pipeline.

Arnab-half-blood-prince:
I think we can not use draco dependency in ch_pipeline. ch_pipeline dependencies are ch_util, driftscan, and caput. Because of this, I put this task here.

ssiegelx:
I think you have it backwards.  `draco` cannot depend on `ch_pipeline` or `ch_util`, because it is experiment agnostic (i.e., can be used by CHIME, CHORD, HIRAX, etc.).  It provides generic tasks that can be used to process radio data acquired with any drift-scan telescope.  `ch_pipeline` provides the CHIME-specific stuff and does indeed depend on `draco`, you can check the `requirements.txt` file in that repository.

Arnab-half-blood-prince:
Okay, got it. I am closing this pull request here and will generate the same in ch_pipeline.

PR #189, opened 2022-08-22
updated z_weights_sum calculation

josephwkania:
Updated z_weights_sum calculation in mockcatalog.py to use new MPIArray allreduce method instead of
the allreduce function which was failing.


jrs65 (approved):

jrs65:
Thanks @josephwkania

PR #190, opened 2022-08-24
Update CollateProducts to remove redistribute calls

ljgray:
CollateProducts is a major bottleneck due almost entirely to multiple `redistibute` calls. Removing the `redistribute` calls reduces cumulative time from ~2500 seconds to ~75 seconds. It may not be feasible to do this robustly while distributed across frequency.

### DIFF
--- a/draco/analysis/transform.py
+++ b/draco/analysis/transform.py
@@ -182,7 +182,6 @@ class CollateProducts(task.SingleTask):
         )
 
         bt_freq = ss.index_map["freq"][freq_ind]
-
         # Determine the input product map and conjugation.
         # If the input timestream is already stacked, then attempt to redefine
         # its representative products so that they contain only feeds that exist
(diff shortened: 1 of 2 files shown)


ljgray:
I've tested with a daily pipeline run without issue, but there may be edge cases that I've missed

draco/analysis/transform.py
```diff
@@ -257,71 +256,81 @@ def process(self, ss):
...
-            else:
-                wss = (ssw[freq_ind, ss_pi] > 0.0).astype(np.float32)
-                wss.local_array[:] *= nprod_in_stack[np.newaxis, ss_pi, stt:ett]
-
-            # Accumulate visibilities, conjugating if required
-            if feedconj == conj:
-                spv[:, sp_pi] += wss * ssv[freq_ind, ss_pi]
-            else:
-                spv[:, sp_pi] += wss * ssv[freq_ind, ss_pi].conj()
+        # Figure out mapping between the local frequencies
```
ljgray, line 268: I tested a daily pipeline run without issue, but I think this will break if the local slice of frequencies in the file and the local slice of frequencies in the telescope instance don't contain all the same frequencies

draco/analysis/transform.py
```diff
@@ -257,71 +256,81 @@ def process(self, ss):
...
+        freq_ind_loc = tools.find_keys(
+            ss.freq[ssv.local_bounds],
+            self.telescope.frequencies[ssv.local_bounds],
+            require_match=True,
+        )
```
ljgray, line 273: I tested a daily pipeline run without issue, but I think this will break if the local slice of frequencies in the file and the local slice of frequencies in the telescope instance don't contain all the same frequencies

PR #191, opened 2022-08-26
Speed up `CollateProducts` by avoiding `redistribute` when possible.

ljgray:
The call to redistribute is only needed if the frequency axis has to be reordered. In many cases this doesn't have to happen, so we can do the task while distributed across frequency, avoiding an expensive `redistribute` call. This checks to see if the frequency mapping is sorted, and if it is it does not redistribute. Without redistributing, the task takes 70-80s total in the daily pipeline, down from around 2500s when redistributing every time.

### DIFF
--- a/draco/analysis/transform.py
+++ b/draco/analysis/transform.py
@@ -182,7 +182,6 @@ class CollateProducts(task.SingleTask):
         )
 
         bt_freq = ss.index_map["freq"][freq_ind]
-
         # Determine the input product map and conjugation.
         # If the input timestream is already stacked, then attempt to redefine
         # its representative products so that they contain only feeds that exist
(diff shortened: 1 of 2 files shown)


draco/analysis/transform.py
```diff
@@ -226,13 +225,19 @@ def process(self, ss):
             **output_kwargs,
         )
 
+        # Check if frequencies are already ordered
+        no_redistribute = freq_ind == sorted(freq_ind)
```
jrs65, line 229: This test is not sufficient I think as `freq_ind` may not represent all the indices in the input array (i.e. it may be down selecting), but the selection may already be sorted.

At the least this test needs changing, probably to something like:
```suggestion
        no_redistribute = freq_ind == list(range(len(ss.freq)))
```

However, I wonder if it would be better to just remove the frequency down selection code and put into a separate task? One that takes a `Telescope` object as a requirement and then just does the same frequency selection that's in this task?

jrs65 (changes requested):

draco/analysis/transform.py
ljgray: If I'm understanding correctly, you mean we would have one task that downselects the frequencies followed by `CollateProducts` which would re-order them (if needed)?

ljgray:
@jrs65 Improved the test for frequency mapping, this should be good to merge now

jrs65 (approved):

PR #192, opened 2022-09-07
feat(RFIMask): calculate RFIMask container rather than applying in place.

jrs65:
This also removes the deprecated destriping functionality.

### DIFF
--- a/draco/analysis/flagging.py
+++ b/draco/analysis/flagging.py
@@ -4,7 +4,7 @@ This includes data quality flagging on timestream data; sun excision on sidereal
 data; and pre-map making flagging on m-modes.
 """
 
-from typing import Union
+from typing import Union, overload
 import warnings
 import numpy as np
 import scipy.signal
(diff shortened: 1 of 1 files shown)


draco/analysis/flagging.py
```diff
@@ -957,30 +957,44 @@ class RFIMask(task.SingleTask):
...
+    @overload
+    def process(self, sstream: containers.SiderealStream) -> containers.SiderealRFIMask:
+        ...
+
+    @overload
+    def process(self, sstream: containers.TimeStream) -> containers.RFIMask:
+        ...
+
+    def process(
+        self, sstream: Union[containers.TimeStream, containers.SiderealStream]
```
ljgray, line 975: Not super familiar with the @overload decorator, but is the typing needed here if it's already given by @overload? Seems redundant

ljgray (approved):

draco/analysis/flagging.py
jrs65: I believe it is, or the function body isn't type checked as far as I understand. I was surprised by this as well, it's a little annoying. This may be type checker dependent but that's why mypy says.

PR #195, opened 2022-09-15
feat(MaskBaselines): mask baselines under an average weight threshold

jrs65:


draco/analysis/flagging.py
```diff
@@ -233,23 +234,36 @@ def process(self, ss):
         ss.redistribute("freq")
 
         baselines = self.telescope.baselines
+
+        # The masking array. True will *retain* a sample
```
ljgray, line 238: I guess we should flag this to switch conventions in the future for consistency.

ljgray (approved):

draco/analysis/flagging.py
jrs65: Good point. It's only internal so I can just changing the naming to make it clearer.

jrs65: Okay. I've done that, but it was slightly invasive, so it would be good if you could look again.

ljgray (approved):

PR #199, opened 2022-09-23
feat(flagging): add task to mask time/freq samples with low weight at…

sjforeman:
… any baseline

This task is intended to be run on individual acquisition files, to catch samples with small weights that won't be caught after the weights are smoothed during CHIME daily processing


draco/analysis/flagging.py
```diff
@@ -702,6 +702,87 @@ def process(self, stream):
...
+        mask = np.any(
+            (local_weight < self.relative_threshold * time_median[:, :, np.newaxis])
+            & (local_weight != 0),
+            axis=1,
+        )
+
+        # If absolute threshold is also specified, include this in mask
+        if self.absolute_threshold is not None:
+            mask = np.any(
+                (local_weight < self.absolute_threshold) & (mask[:, np.newaxis, :]),
```
tristpinsm, line 768: won't this only mask times/frequencies that have already been masked by the relative threshold?

sjforeman: Agreed. How about this?
```
if self.absolute_threshold is not None:
    mask = np.any(
        ((local_weight < self.absolute_threshold) & (local_weight != 0)) | mask[:, np.newaxis, :],
        axis=1,
    )
```

tristpinsm: If I understand correctly, I think you'd want to combine the masks with an `|` after the `np.any`, and maybe also exclude zero weights from the comparison, as done above.

tristpinsm:
As a more general comment, it seems like this could be integrated with the existing `VisThresholdWeight` task, where you'd specify the axis to take the median over. It might not be worth the effort to do that right now.

tristpinsm (approved):

PR #202, opened 2022-09-29
Move _invert_no_zero to caput and add deprecation warning

ljgray:
Accompanies radiocosmology/caput#221

### DIFF
--- a/draco/util/_fast_tools.pyx
+++ b/draco/util/_fast_tools.pyx
@@ -262,26 +262,3 @@ ctypedef fused real_or_complex:
     double complex
     float
     float complex
-
-@cython.wraparound(False)
-@cython.boundscheck(False)
-@cython.cdivision(True)
-cpdef _invert_no_zero(real_or_complex [:] array, real_or_complex [:] out):
-
-    cdef bint cond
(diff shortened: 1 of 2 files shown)


jrs65:
I think this is linked to the other PR, but we should move the routine at the level of the `draco.util.tools.invert_no_zero` rather than the `_fast_tools._invert_no_zero` level.

ljgray:
> I think this is linked to the other PR, but we should move the routine at the level of the `draco.util.tools.invert_no_zero` rather than the `_fast_tools._invert_no_zero` level.

What I'm thinking here is that we would just remove invert_no_zero from draco altogether, and this is just temporary so that anyone using it can switch to importing from caput. If you think it's fine to make a breaking change then we could just remove it altogether

jrs65:
Agreed. But you should remove `_invert_no_zero` entirely, move the remaining functionality of `draco.util.tools.invert_no_zero` into caput, and turn `draco.util.tools.invert_no_zero` into a thin shim calling `caput.tools.invert_no_zero`.

ljgray:
Ah, I see what you mean

jrs65 (approved):

PR #203, opened 2022-09-30
feat(flagging): improve code for masking low weights

sjforeman:
Addresses https://github.com/radiocosmology/draco/issues/201, by rewriting `draco.analysis.flagging.ThresholdVisWeightBaseline` such that:
- It outputs a container containing a baseline-dependent mask, instead of applying this mask in-place
- The user can choose whether to consider only co-pol baselines or all baselines in the mask. (In some CHIME daily acquisitions, certain cross-pol baselines have low weights while the co-pol baselines are fine. In analyses that only use co-pol, we might not want to mask all baselines at a given freq/time sample if only cross-pol baselines are problematic, which is why I included this option.)
- The user can choose whether to use a mean or median as the averaging procedure

I've also added containers to hold baseline-dependent masks, and a task to collapse such a mask down into a `RFIMask`.

Note that this `ThresholdVisWeightBaseline` does *not* incorporate the behavior of the "old `ThresholdVisWeight`" that is reintroduced in https://github.com/radiocosmology/draco/pull/200, so if we want to keep both, one or the other of these tasks will need to be renamed.


draco/analysis/flagging.py
```diff
@@ -1755,3 +1674,207 @@ def process(self, data):
...
+            if np.sum(x != 0) > 0
+            else 0.0
+        )
+
+        # Apply averaging function to weights for each baseline
+        average_weight = np.array(
+            [average_nonzero_func(w) for w in local_weight.transpose(1, 0, 2)],
+            dtype=local_weight.dtype,
+        )
+
```
ljgray, line 1777:
```suggestion
    average_weight = average_func(
        np.ma.array(local_weight, mask=(local_weight==0.0)), 
        axis=(0,2)
    ).data
```
This combined with using np.ma.mean and np.ma.median should do the same as the combo of the lambda function + average_weight loop above (np.ma.array shouldn't make a copy of the data).

draco/analysis/flagging.py
```diff
@@ -1755,3 +1674,207 @@ def process(self, data):
...
+        # If only flagging co-pol baselines, make separate mask to select those,
+        # and multiply into low-weight mask
+        if self.pols_to_flag == "copol":
+
+            # Get local section of stack axis
+            stack_lo = stream.weight.local_offset[1]
+            stack_ls = stream.weight.local_shape[1]
+            local_stack = stream.stack[stack_lo : stack_lo + stack_ls]
```
ljgray, line 1795:
```suggestion
            local_stack = stream.stack[local_weight.local_bounds]
```
MPIArray.local_bounds is a fairly new method to simplify getting a slice like this

ljgray (changes requested):

draco/analysis/flagging.py
sjforeman: Thanks! I double-checked that this doesn't affect the results, so I've made this change

draco/analysis/flagging.py
sjforeman: I had to use `stream.weight[:].local bounds`, but otherwise it works out

ljgray: Right, yeah the local_weight would already by a numpy array

jrs65 (changes requested):
I don't think we really want to remove the old ThresholdVisWeight at this time. Can you reintroduce it such that we can keep using it in the rev06 run?

Also, please hide how the sausage was made in here and rebase the commits down to the externally visible changes and hide all the internal ones.

jrs65 (approved):

PR #204, opened 2022-10-20
feat(misc): add task to test mpi communication

ljgray:
This task just confirms that all mpi processes can communicate with each other. Not sure if there is a better place to put the task

### DIFF
--- a/draco/core/misc.py
+++ b/draco/core/misc.py
@@ -5,6 +5,7 @@ appropriate module, or enough related tasks end up in here such that they can
 all be moved out into their own module.
 """
 
+import logging
 
 import numpy as np
 
@@ -12,6 +13,8 @@ from caput import config
 
 from ..core import task, containers
 from ..util import tools
+
+logger = logging.getLogger(__name__)
 
 
 class ApplyGain(task.SingleTask):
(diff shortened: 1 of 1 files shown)


draco/core/misc.py
```diff
@@ -214,6 +217,36 @@ def finish(self):
...
+    across all nodes.
+    """
+
+    timeout = config.Property(proptype=int, default=60)
+
+    def setup(self):
+        from caput import mpiutil
+        import time
+
+        comm = mpiutil._comm
```
jrs65, line 231:
```suggestion
        import time

        comm = self.comm
```

draco/core/misc.py
```diff
@@ -214,6 +217,36 @@ def finish(self):
...
+        results = []
+
+        # Send and receive across all ranks
+        for i in range(comm.size):
+            results.append(comm.Irecv(np.zeros(1), (comm.rank - i) % comm.size))
+            comm.Isend(np.ones(1), (comm.rank + i) % comm.size)
+
+        time.sleep(self.timeout)
```
jrs65, line 239: I might send something larger in here to stress it a bit more. Maybe a few MB per message.

Also, you might want to put a testable pattern in the message to ensure it's the correct, uncorrupted message, e.g. alternating elements of the sending and receiving rank.

draco/core/misc.py
```diff
@@ -214,6 +217,36 @@ def finish(self):
         return items
 
 
+class CheckMPIEnvironment(task.MPILoggedTask):
+    """Check that the current MPI environment can communicate
+    across all nodes.
+    """
+
+    timeout = config.Property(proptype=int, default=60)
```
jrs65, line 225: This might not be long enough. The job start up time can often be a few minutes, and you might imagine that could vary a lot between ranks. If one starts up a minute before the others then this will fail regardless of whether they could all communicate.

jrs65 (changes requested):
Looking about perfect!

jrs65 (approved):

draco/core/misc.py
```diff
@@ -214,6 +217,50 @@ def finish(self):
...
+        def prep_buf(x):
+            return mpiutil.typemap(x.dtype)
+
+        # Send and receive across all ranks
+        for i in range(comm.size):
+            send = (comm.rank + i) % comm.size
+            recv = (comm.rank - i) % comm.size
+
+            results.append(comm.Irecv(prep_buf(recvs[recv, :]), recv))
+            comm.Isend(prep_buf(sends[comm.rank, :]), send)
```
jrs65, line 247: I don't think you need a `prep_buf` in here. mpi4py should figure it out automatically
```suggestion
        # Send and receive across all ranks
        for i in range(comm.size):
            send = (comm.rank + i) % comm.size
            recv = (comm.rank - i) % comm.size

            results.append(comm.Irecv(recvs[recv, :], recv))
            comm.Isend(sends[comm.rank, :], send)
```

jrs65 (changes requested):

draco/core/misc.py
ljgray: Ah, yeah, I had needed it in a previous iteration but should be able to remove it now

jrs65: You might also want to give an explicit width type for the numpy array. It's easier to reason about. I think `float` -> `np.float64` but I'm never entirely sure.

ljgray: Fair

draco/core/misc.py
```diff
@@ -214,6 +217,58 @@ def finish(self):
...
+    across all nodes.
+    """
+
+    timeout = config.Property(proptype=int, default=240)
+
+    def setup(self):
+        import time
+
+        comm = self.comm
+        n = 500000  # Corresponds to a 4 Mb buffer
```
jrs65, line 231:
```suggestion
        n = 500000  # Corresponds to a 4 MB buffer
```
I think it's 4 megabytes, not megabits.

jrs65 (approved):
Very minor suggestion.

draco/core/misc.py
ljgray: Ah, yes, good catch, I specifically changed that and apparently did it wrong

PR #206, opened 2022-10-23
fix(CheckMPIEnvironment): use the logger on the task

jrs65:


ljgray (approved):
Looks good, should also be able to remove the logging import now I think

PR #209, opened 2022-10-27
refactor(flagging): clean up various versions of ThresholdVisWeight.

ljgray:
This includes removing the redundant BaselineAlt copy and renaming the frequency-dependent version for consistency.

### DIFF
--- a/draco/analysis/flagging.py
+++ b/draco/analysis/flagging.py
@@ -631,7 +631,7 @@ class SmoothVisWeight(task.SingleTask):
         return data
 
 
-class ThresholdVisWeight(task.SingleTask):
+class ThresholdVisWeightFrequency(task.SingleTask):
     """Create a mask to remove all weights below a per-frequency threshold.
 
     A single relative threshold is set for each frequency along with an absolute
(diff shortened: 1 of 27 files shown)


ljgray:
`ThresholdVisWeightBaseline` will function the same as `ThresholdVisWeightBaselineAlt` with proper config, including
- using the mean
- keeping the ignore_weight threshold as 0
- masking all pols
So, we can remove `ThresholdVisWeightBaselineAlt`. 

We aren't currently using `ThresholdVisWeightFrequency`, so I'm leaving that in for now. The changes look like a bit of a mess because I moved a couple of classes around in the file to try to improve some organization.

draco/analysis/flagging.py
```diff
@@ -697,9 +697,210 @@ def process(self, stream):
...
+    average_type : string, optional
+        Type of average to use ("median" or "mean"). Default: "median".
+    absolute_threshold : float, optional
+        Any weights with values less than this number will be set to zero.
+        Default: 1e-7.
+    relative_threshold : float, optional
+        Any weights with values less than this number times the average weight
+        will be set to zero. Default: 1e-6.
+    ignore_absolute_threshold : float, optional
+        Any weights will values less than this number will be ignored when
```
sjforeman, line 731: Fixing my own typo:
```suggestion
        Any weights with values less than this number will be ignored when
```

draco/analysis/flagging.py
```diff
@@ -697,9 +697,210 @@ def process(self, stream):
...
+        telescope : TransitTelescope
+        """
+
+        self.telescope = io.get_telescope(telescope)
+
+    def process(
+        self,
+        stream,
+    ) -> Union[containers.BaselineMask, containers.SiderealBaselineMask]:
+        """Construct mask and apply to input weights, if desired.
```
sjforeman, line 758: This task used to have the option to apply the mask to the input weights, but it doesn't any more
```suggestion
        """Construct baseline-dependent mask.
```

sjforeman (approved):
Thanks Liam, this looks good to me! My only comments are corrections to things that I wrote a while ago.

When this is merged, we might want to create an issue or draft PR for `ch_pipeline` that includes your note about the relationship between `ThresholdVisWeightBaselineAlt` and `ThresholdVisWeightBaseline`, so we can make that change in the next revision of the daily processing.

PR #210, opened 2022-11-04
Lyman-alpha cross-correlation

tristpinsm:
I'm about to start writing the paper up for this analysis, so I thought it would be good to also get the code reviewed and merged.


draco/synthesis/noise.py
```diff
@@ -332,3 +332,60 @@ def process(self, data_exp):
                 )
 
         return data_exp
+
+
+class GaussianNoiseExternal(task.SingleTask, random.RandomTask):
```
jrs65, line 337: It seems like this could be done by using the existing tasks `GaussianNoise` to generate a realisation and then `MixData` to combine them, rather than adding an extra task?!

draco/synthesis/lyalpha.py
```diff
@@ -0,0 +1,55 @@
...
+    delta = lognormal_transform(delta, axis=1)
+
+    F = np.exp(-A[:, np.newaxis] * (1 + delta) ** alpha)
+    F /= F.mean(axis=1)[:, np.newaxis]
+
+    return F
+
+
+class GenerateFluxTransmission(SingleTask):
+    # TODO: should this use the HI density or matter density as an input?
```
jrs65, line 29: Good question. I guess really it's the HI density, though you'd have the most flexibility if you just started from the matter density. 

Also this needs docstrings.

draco/core/io.py
```diff
@@ -352,6 +342,328 @@ def process(self):
...
+    Attributes
+    ----------
+    delta_files : list or dict
+        A dictionary specifying a file group, or a list of them.
+    """
+
+    delta_files = config.Property(proptype=_list_of_filegroups)
+
+    def process(self):
+        """Load spectra from all files and combine into a single
```
jrs65, line 469: Initial docstring should be a single line. Longer comments can go in a paragraph below.

draco/core/io.py
```diff
@@ -352,6 +342,328 @@ def process(self):
...
+                        continue
+                    to_send = array[slices[r]].flatten()
+                    sends.append(self.comm.Isend([to_send, mpitype], dest=r))
+                for req in sends:
+                    req.wait()
+            else:
+                local_shape = list(shape)
+                local_shape[axis] = n_all[self.comm.rank]
+                local = np.zeros(local_shape, dtype=dtype)
+                self.comm.Recv([local, mpitype], source=root)
```
jrs65, line 636: You typically don't need to give the mpitype in these send/recv calls. They'll get inferred.

draco/core/io.py
```diff
@@ -352,6 +342,328 @@ def process(self):
...
+        self.comm.Bcast(z, root=0)
+        self.comm.Bcast(lam, root=0)
+        self.comm.Bcast(obj_id, root=0)
+        self.log.debug(f"(rank {self.comm.rank}) got axes.")
+
+        # convert wavelength (A) to frequency (MHz)
+        freq = units.c / lam * 1e4
+
+        # scatter the distributed datasets over frequency
+        def scatter_array(array, root=0, axis=0):
```
jrs65, line 599: I think you can easily leverage existing MPIArray routines to implement a scatter. Basically you add an extra length-1 axis at the front on rank=0, and then pretend that it is a (badly) distributed MPIArray, and then redistribute to axis=0, and then peel off the spurious first axis. Something like:
```
shape = array.shape if self.comm.rank == 0 else None
shape = self.comm.bcast(shape, root=0)
array_ext = array[np.newaxis, ...] if self.comm.rank == 0 else np.empty((0,) + shape, dtype=np.float64)
arr0 = mpiarray.MPIArray.wrap(array_ext, axis=0)
arr_dist = arr0.redistribute(axis=1)[0]
```

jrs65: This is definitely better than reimplementing a scatter directly in here.

draco/core/io.py
```diff
@@ -352,6 +342,328 @@ def process(self):
...
+        out = np.zeros((3, n), dtype=np.float64)
+        out[0, :] = pos["ra"][zsel]
+        out[1, :] = pos["dec"][zsel]
+        out[2, :] = z[zsel]
+
+        return out
+
+
+class LoadFITSCatalogDLA2(LoadFITSCatalog):
+    """Load DLA positions from a FITS source catalog structured like the one
```
jrs65, line 420: Again this needs a bit of tweaking to fit the numpydoc format.

jrs65 (changes requested):
I'm still going, but here are some first round comments.

draco/synthesis/noise.py
tristpinsm: yeah I guess so, I hadn't thought of using `MixData`

draco/core/io.py
tristpinsm: I see, yeah that is nicer... Well at least I got to practice some MPI calls.

PR #212, opened 2022-11-23
fix(misc): Log MPI check as successful only after all ranks complete

ljgray:
I noticed this issue in the logs of rev_06 3237, if one rank fails it will log the failure after the timeout, but other ranks might have already logged that they were successful, which is confusing. This should fix that by only logging success after all ranks have completed.


jrs65:
Has this task caught many issues?

jrs65 (approved):

ljgray:
> Has this task caught many issues?

It caught its first one today, but we also haven't had any of the errors that it was intended to catch until now

PR #214, opened 2022-11-27
feat(io): allowing zarr zip jobs to run in the background

jrs65:


ljgray:
@jrs65 Has this been fairly well tested? I'd be interested in trying to get it included in rev_07 if it's mostly ready to go

jrs65:
Yes. I think it's pretty well tested. In the end I think it helped a bit, but not as much as I was hoping. It's worth trying again though.

draco/core/io.py
```diff
@@ -982,6 +982,131 @@ def process(self):
...
+
+    def setup(self):
+        """Check the parameters and determine the ranks to use."""
+
+        if not self.output_name.endswith(".zarr.zip"):
+            raise ConfigError("File output name must end in `.zarr.zip`.")
+
+        # Trim off the .zip suffix and fix the file format
+        self.output_name = self.output_name[:-4]
+        self.output_format = fileformats.Zarr
```
ljgray, line 1017: This seems to break right now because the `config` property for `output_format` expects a string and tries to call `.lower()` on it. I submitted a PR (radiocosmology/caput#237) to address it, as I think that setting it this way should be valid.

draco/core/io.py
```diff
@@ -982,6 +982,131 @@ def process(self):
...
+        for h in self._handles:
+
+            self.log.debug(f"Waiting on job processing {h.filename}")
+
+            if h.handle is not None:
+                returncode = h.handle.wait()
+
+                if returncode != 0 or not os.path.exists(h.filename):
+                    self.log.debug("Error occurred while zipping. Debug logs follow...")
+                    self.log.debug(f"stdout={h.handle.stdout}")
```
ljgray, line 1102: If we're including this sort of debugging, I think this would be useful
```suggestion
                    self.log.debug(f"returncode={returncode}")
                    self.log.debug(f"stdout={h.handle.stdout}")
```

ljgray (changes requested):
Overall it looks good. I'm assuming the pipeline config would be something like what's below, but maybe there's something that has to be subtly different.

```
  - type: some.task
    ...
    out: some_file

  - type: ...SaveZarrZip
    in: some_file
    out: some_file_handle
    params:
        save: true
        output_name: "some_file.zarr.zip"

  - type: ...WaitZarrZip
    in: some_file_handle
```

ljgray (approved):
This seems good to me to merge in, I went ahead and added a missing f-string in `SaveZarrZip` and it all seems to work as expected

PR #217, opened 2022-12-23
Refactor delay-space measurement code and add features

sjforeman:
This is meant to supersede #176 and #216:
- Refactors `DelaySpectrumEstimator`, `DelaySpectrumEstimatorBase`, and `DelaySpectrumWienerBase` to eliminate repeated code blocks and enable future extensions
- Implements the fixes and features from #216 

Notably, I've kept the scheme of compressing many axes into a `baseline` axis. I don't think this is ideal in the long run, but in the short term, I think it's higher-priority to get the Wiener filter and #216 features all merged coherently

### DIFF
--- a/draco/analysis/delay.py
+++ b/draco/analysis/delay.py
@@ -526,7 +526,6 @@ class DelayGibbsSamplerBase(DelayTransformBase, random.RandomTask):
         out_cont : `contaiers.DelayTransform` or `containers.DelaySpectrum`
             Output delay spectrum or delay power spectrum.
         """
-
         nbase = out_cont.spectrum.global_shape[0]
 
         # Set initial conditions for delay power spectrum
@@ -629,7 +628,6 @@ class DelayGeneralContainerBase(DelayTransformBase):
         out_cont : `containers.DelayTransform` or `containers.DelaySpectrum`
             Container for output delay spectrum or power spectrum.
         """
-
         if self.dataset not in ss.datasets:
             raise ValueError(
                 f"Specified dataset to delay transform ({self.dataset}) not in "
@@ -718,6 +716,7 @@ class DelayPowerSpectrumStokesIEstimator(DelayGibbsSamplerBase):
         Parameters
         ----------
         telescope : TransitTelescope
+            Telescope object we'll use for baseline and polarization information.
         """
         self.telescope = io.get_telescope(telescope)
 
@@ -738,7 +737,6 @@ class DelayPowerSpectrumStokesIEstimator(DelayGibbsSamplerBase):
         out_cont : `containers.DelayTransform` or `containers.DelaySpectrum`
             Container for output delay spectrum or power spectrum.
         """
-
         tel = self.telescope
 
         # Construct the Stokes I vis, and transpose from [baseline, freq, ra] to
@@ -792,6 +790,7 @@ class DelaySpectrumWienerEstimator(DelayGeneralContainerBase):
         Parameters
         ----------
         dps : `containers.DelaySpectrum`
+            Delay power spectrum for signal part of Wiener filter.
         """
         self.dps = dps
 
@@ -812,7 +811,6 @@ class DelaySpectrumWienerEstimator(DelayGeneralContainerBase):
         out_cont : `containers.DelaySpectrum`
             Output delay spectrum.
         """
-
         nbase = out_cont.spectrum.global_shape[0]
 
         # Read the delay power spectrum to use as the signal covariance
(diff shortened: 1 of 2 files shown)


draco/core/containers.py
```diff
@@ -2300,10 +2300,46 @@ class DelaySpectrum(ContainerBase):
...
+    def spectrum(self):
+        return self.datasets["spectrum"]
+
+    @property
+    def weight_boost(self):
+        return self.attrs["weight_boost"]
+
+
+class DelayTransform(ContainerBase):
+    """Container for a delay spectrum."""
```
ljgray, line 2317: I think it could be beneficial to elaborate a bit more on the difference between a delay power spectrum and a delay spectrum here, especially since we've used the terms somewhat interchangeably (specifically, why does `DelayTransform` say it holds a delay spectrum and not `DelaySpectrum`?)

draco/analysis/delay.py
```diff
@@ -416,52 +403,100 @@ def process(self, ss):
...
-        # Initialise the spectrum container
-        delay_spec = containers.DelaySpectrum(
-            baseline=baselines, delay=delays, attrs_from=ss
-        )
-        delay_spec.redistribute("baseline")
-        delay_spec.spectrum[:] = 0.0
+        # Get views of data and weights appropriate for the type of processing we're
+        # doing. Also get container for output (DelayTransform or DelaySpectrum)
+        data_view, weight_view, out_cont = self._process_data(ss)
+
```
ljgray, line 427: `DelayTransformBase` should probably have `_process_data` and `_evaluate` methods, both of which would just raise a `NotImplementedError`. It makes the exception more useful if someone tries to use this class on its own. Those methods would then have docstings describing their intended use.

draco/analysis/delay.py
```diff
@@ -416,52 +403,100 @@ def process(self, ss):
...
+        # Evaluate frequency->delay transform
+        output_spectrum = self._evaluate(data_view, weight_view, out_cont)
```
ljgray, line 429: This is very minor. I think it could be more clear here that `output_spectrum` and `out_cont` are actually the same object. Possibly having `_evaluate` not return anything, although I'm not sure if that would actually be more clear.

draco/analysis/delay.py
```diff
@@ -1182,6 +1266,74 @@ def _draw_ps_sample(d):
     return spec
 
 
+def wiener_filter(
```
ljgray, line 1269:
```suggestion
def delay_spectrum_wiener_filter(
```
For clarity about what the function does and consistency with `delay_spectrum_gibbs`

ljgray (changes requested):
This looks good to me. I have a couple of nitpicky stylistic comments but that's about it

draco/core/containers.py
sjforeman: Agreed, the conventions are confusing. I've added some notes to the `DelaySpectrum` docstring to clarify this.

What I'd really like to do is rename `DelaySpectrum` -> `DelayPowerSpectrum` and `DelayTransform` -> `DelaySpectrum` to make the meanings clearer, but this would break all sorts of other code. A reasonable compromise might be `DelaySpectrum` -> `DelayPowerSpec` and `DelayTransform` -> `DelaySpec`, and then aliasing `DelaySpectrum` to `DelayPowerSpec` for backwards-compatibility.

draco/analysis/delay.py
sjforeman: Done

draco/analysis/delay.py
sjforeman: I've changed everything to `out_cont` and added a few clarifying comments - let me know if this doesn't help enough

draco/analysis/delay.py
sjforeman: Done. I also renamed `delay_spectrum_gibbs` -> `delay_power_spectrum_gibbs` to distinguish between delay spectra and power spectra, and aliased the former to the latter for backwards-compatibility

ljgray (commented):
These changes look good to me overall. I spoke to @jrs65 on Friday and he mentioned having some thoughts about modifying the delay spectrum implementation to be split into multiple pipeline tasks. So I don't have anything else to change here, but we should let him weight in as well.

Arnab-half-blood-prince:
Hi, I am also checking this and it looks good to me. However, I have one suggestion to add the freq axis in the index_map of the DelaySpectrum and DelayTransform container. This is needed to estimate various cosmological quantities downstream. For example, we need redshift to estimate comoving distances and also chan resolution to get the k_parallel axis.  

Probably something like this within the task -
 # Create the index map for the frequency axis
  delay_spec.create_index_map("freq",ss.freq)

And add this in the  container - 

    @property
    def freq(self):
        return self.index_map["freq"]

ljgray (approved):

PR #218, opened 2023-01-09
fix(band_covariance): fix out of bounds index in _band_wiener_covariance

### LINKED ISSUE #215: Numerical issues in `SmoothVisWeights` -> `SiderealRegridder` tasks.
This is an issue to track the NaN/inf values appearing in the `SiderealRegridder` task. Based on the current testing:
1. Using a weighted median filter in `SmoothVisWeights` (draco#187) and setting weights to zero where vis_weights are zero avoids the issue, suggesting that it's related to the inclusion of zeros in the median calculation in `SmoothVisWeights`. Setting all weights to 1 or using `scipy.ndimage.median_filter` produces the error.
2. The error occurs on different ranks/subsets of the data, making me think it's related to some floating point error in very small/large values
3. Oddly, saving the data returned by `SmoothVisWeights` and then feeding that into `SiderealRegridder` in either a notebook or a new pipeline task avoids the issue. The only thing I can think of here is that the save/load process affects the data in a subtle way

ljgray:
An internal loop indexed 1 past the end index, causing an out of bounds access at the end of an array, which resulted in NaNs being returned for some datasets. 

Closes #215

### DIFF
--- a/draco/util/_fast_tools.pyx
+++ b/draco/util/_fast_tools.pyx
@@ -23,13 +23,11 @@ cpdef _band_wiener_covariance(double [:, ::1] Rn, double [::1] Ni,
 
     cdef double [:, ::1] Ci = np.zeros((bw+1, Rn.shape[0]), dtype=np.float64)
 
-    cdef int N, M
     cdef int alpha, beta, betap, j, alpha_start, si, ei
 
     cdef double t
 
-    N = Rn.shape[0]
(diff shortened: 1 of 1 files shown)


tristpinsm (approved):
I couldn't find any other use of `band_wiener_covariance` in `draco` or `ch_pipeline`, so it's probably fine that this changes the implicit convention for defining the end index.

PR #221, opened 2023-01-26
Mask gain interpolation region after bad calibration

ljgray:
Gains are set to 1 when calibration determines some frequencies are bad. The gain interpolation region will also produce junk data, so this is designed to mask that. 

This PR has 3 commits:
1. A couple of documentation and naming updates (maybe shouldn't be in this PR)
2. A task to mask any weight above or below flat thresholds, mostly for avoiding numerical issues
3. A task to mask the gain boundary regions

 This task is run for each individual timestream file then masks are combined and applied. It also requires subsequently using `ExpandMask` from `ch_pipeline` to actually mask the gain interpolation region AFTER individual timestreams have been grouped into a full day.

### DIFF
--- a/draco/analysis/flagging.py
+++ b/draco/analysis/flagging.py
@@ -577,6 +577,60 @@ class RadiometerWeight(task.SingleTask):
         return stream
 
 
+class SanitizeWeights(task.SingleTask):
+    """Flags weights outside of a valid range.
+
+    Flags any weights above a max threshold and below a minimum threshold.
+    Baseline dependent, so only some baselines may be flagged.
+
+    Attributes
+    ----------
+    max_thresh : float
+        largest value to keep
+    min_thresh : float
+        smallest value to keep
+    """
+
+    max_thresh = config.Property(proptype=np.float32, default=1e10)
+    min_thresh = config.Property(proptype=np.float32, default=1e-10)
+
+    def setup(self):
+        """Check the max and min values.
+
+        Raises
+        ------
+        ValueError
+            if min_thresh is larger than max_thresh
+        """
+        if self.min_thresh >= self.max_thresh:
+            raise ValueError("Minimum threshold is larger than maximum threshold.")
+
+    def process(self, data):
+        """Mask any weights outside of the threshold range.
+
+        Parameters
+        ----------
+        data : :class:`andata.CorrData` or :class:`containers.VisContainer` object
+            Data containing the weights to be flagged
+
+        Returns
+        -------
+        data : same object as data
+            Data object with high/low weights masked in-place
+
+        """
+
+        # Ensure data is distributed in frequency
+        data.redistribute("freq")
+
+        weight = data.weight[:].local_array
+
+        weight[weight > self.max_thresh] = 0.0
+        weight[weight < self.min_thresh] = 0.0
+
+        return data
+
+
 class SmoothVisWeight(task.SingleTask):
     """Smooth the visibility weights with a median filter.
 
--- a/draco/analysis/flagging.py
+++ b/draco/analysis/flagging.py
@@ -2,6 +2,9 @@
 
 This includes data quality flagging on timestream data; sun excision on sidereal
 data; and pre-map making flagging on m-modes.
+
+The convention for flagging/masking is `True` for contaminated samples that should
+be excluded and `False` for clean samples.
 """
 
 from typing import Union, overload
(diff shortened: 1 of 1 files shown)


draco/analysis/flagging.py
```diff
@@ -577,6 +577,60 @@ def process(self, stream):
...
+        """
+        if self.min_thresh >= self.max_thresh:
+            raise ValueError("Minimum threshold is larger than maximum threshold.")
+
+    def process(self, data):
+        """Mask any weights outside of the threshold range.
+
+        Parameters
+        ----------
+        data : :class:`andata.CorrData` or :class:`containers.TimeStream` object
```
sjforeman, line 649: Actually, it looks like this task should work for any `containers.VisContainer`, since it only relies on the `weight` property

draco/analysis/flagging.py
```diff
@@ -416,6 +416,42 @@ def process(self, data):
         return out
 
 
+class BadGainMask(task.SingleTask):
+    """Get a mask of regions with bad gain.
+
+    Assumes that bad gains are set to 1.
```
sjforeman, line 422: Would it be worth adding a `bad_gain_value` attribute, with default value of 1.0, in case there are situations in the future where some other number has been used to identify bad gains?

draco/analysis/flagging.py
```diff
@@ -416,6 +416,42 @@ def process(self, data):
...
+        -------
+        mask : RFIMask container
+            Time-freq mask
+        """
+
+        # Ensure data is distributed in frequency
+        data.redistribute("freq")
+
+        # Boolean mask where gains are bad across all baselines.
+        mask = np.all(data.gain[:] == 1.0, axis=1).allgather()
```
sjforeman, line 443: This kind of floating-point comparison makes me nervous, but I can't find a simple `np.allclose()` analog for `np.all(..., axis=1)`, so I guess it's fine?

sjforeman (changes requested):

draco/analysis/flagging.py
ljgray: I've changed this check in a way that should work

draco/analysis/flagging.py
```diff
@@ -1361,8 +1459,8 @@ def process(self, tstream, rfimask):
         return tsc
 
 
-class MaskFreq(task.SingleTask):
-    """Apply a mask to the frequency axis.
+class FreqMask(task.SingleTask):
```
jrs65, line 1462: I'm not so convinced about this rename. I like it when task names describe an action, and this one is now pretty unclear (i.e could it be a container with a frequency mask in it?). The old name wasn't great, but I think this is going backwards slightly.

draco/analysis/flagging.py
```diff
@@ -1263,10 +1360,11 @@ def process(
         return mask_cont
 
 
-class ApplyRFIMask(task.SingleTask):
-    """Apply an RFIMask to the data.
+class ApplyTimeFreqMask(task.SingleTask):
```
jrs65, line 1363: This rename makes sense.

draco/analysis/flagging.py
```diff
@@ -413,6 +416,47 @@ def process(self, data):
...
 
 
+class BadGainMask(task.SingleTask):
+    """Get a mask of regions with bad gain.
+
+    Assumes that bad gains are set to 1.
+    """
+
+    threshold = config.Property(proptype=float, default=1.0)
+    thresold_tol = config.Property(proptype=float, default=1e-5)
```
jrs65, line 426:
```suggestion
    threshold_tol = config.Property(proptype=float, default=1e-5)
```

draco/analysis/flagging.py
```diff
@@ -413,6 +416,47 @@ def process(self, data):
...
+        mask : RFIMask container
+            Time-freq mask
+        """
+
+        # Ensure data is distributed in frequency
+        data.redistribute("freq")
+
+        # Boolean mask where gains are bad across all baselines.
+        mask = np.all(
+            data.gain[:] <= self.threshold + self.thresold_tol, axis=1
```
jrs65, line 447:
```suggestion
            data.gain[:] <= self.threshold + self.threshold_tol, axis=1
```

draco/analysis/flagging.py
```diff
@@ -413,6 +416,47 @@ def process(self, data):
         return out
 
 
+class BadGainMask(task.SingleTask):
```
jrs65, line 419: Along the lines of the comment below, this might be better called `MaskBadGains`

jrs65 (changes requested):

jrs65 (approved):

PR #222, opened 2023-02-01
Timestamp fixes (1/1): Remove some unnecessary time axis copies

ljgray:
Instead, the copy_from parameter will handle directly copying all index_map entries, including the time axis.


tristpinsm:
this seems fairly innocuous to me, but it's a piece of a larger set of changes for which I don't know the context very well, so not sure it's very useful for me to approve.

jrs65 (approved):
I guess we shouldn't merge without all the other changes going in too?!

ljgray:
> I guess we shouldn't merge without all the other changes going in too?!

I guess this one probably would be fine to go in now, but I'll just merge all the timestamp stuff together.

PR #227, opened 2023-02-18
Timestamp Alignment(2/4): Copy index_map dataset attributes and remove `time` property

ljgray:
`index_map` dataset attributes will be copied when `axes_from` is set. The `.time` property is removed from `containers.TODContainer` as it will be moved to `caput.tod.TOData`.

Related PRs:
- https://github.com/radiocosmology/caput/pull/234
- https://github.com/chime-experiment/ch_pipeline/pull/168
- https://github.com/chime-experiment/ch_util/pull/52

Note: this PR will not pass CI checks until the `caput` PR is merged

Another note: This raises a deprecation warning from `numpy.array_equal`. This is an internal numpy issue and should be fixed in numpy 1.24 (we're running 1.22.2 on cedar)

### DIFF
--- a/draco/core/containers.py
+++ b/draco/core/containers.py
@@ -194,19 +194,27 @@ class ContainerBase(memh5.BasicCont):
 
             # Check if axis is specified in initialiser
             if axis in kwargs:
-                # If axis is an integer, turn into an arange as a default definition
-                if isinstance(kwargs[axis], int):
-                    axis_map = np.arange(kwargs[axis])
-                else:
-                    axis_map = kwargs[axis]
+                axis_map = kwargs[axis]
+                copy_axis_attrs = False
 
-            # If not set in the arguments copy from another object if set
-            elif axes_from is not None and axis in axes_from.index_map:
-                axis_map = axes_from.index_map[axis]
+                # If axis is an integer, turn into an arange as a default definition
+                if isinstance(axis_map, int):
+                    axis_map = np.arange(axis_map)
+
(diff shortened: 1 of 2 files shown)


test/test_containers.py
```diff
@@ -16,6 +16,7 @@ def ss_container():
     ss.attrs["test_attr1"] = "hello"
     ss.vis.attrs["test_attr2"] = "hello2"
     ss.weight.attrs["test_attr3"] = "hello3"
+    ss.index_attrs["freq"]["alignment"] = 1
```
ljgray, line 19: Some of these tests must have been accidentally removed because this shouldn't work

draco/core/containers.py
```diff
@@ -438,18 +446,80 @@ def axes(self):
...
-        # classes where
-        # the axes get added at run time)
+        # classes where the axes get added at run time)
         axes |= set(self.__dict__.get("_axes", []))
 
         # This must be the same order on all ranks, so we need to explicitly sort to
         # get around the hash randomization
         return tuple(sorted(axes))
 
+    def _sanitize_axis_map(self, axis, axis_map, axes_from):
```
ljgray, line 456: This is a pretty iffy way to do this check. I think that I've removed any cases where an axis is set in kwargs directly from a property (i.e. time=data.time[:]), so we can probably remove this, but I wonder about having some way to warn a user about this specific case.

jrs65 (approved):

PR #229, opened 2023-02-22
feat(flagging): introduce a task to apply a baseline-dependent mask.

ljgray:
Most baseline dependent masking has been done in place. To follow the convention of separating masking tasks into those that create masks and those that apply masks, this task lets us directly apply `BaselineMask` and `SiderealBaselineMask` containers to timestream or sidereal stream without having to collapse it to just a time-freq mask.

This is needed to properly replicate the behaviour of the previous `ThresholdVisWeightBaselineAlt` task, and more generally to apply any baseline-dependent mask. Optionally, `ThresholdVisWeightBaseline` could be given an `in_place` flag

### DIFF
--- a/draco/analysis/flagging.py
+++ b/draco/analysis/flagging.py
@@ -1464,6 +1464,101 @@ class ApplyTimeFreqMask(task.SingleTask):
         return tsc
 
 
+class ApplyBaselineMask(task.SingleTask):
+    """Apply a distributed mask that varies across baselines.
+
+    No broadcasting is done, so the data and mask should have the same
+    axes. This shouldn't be used for non-distributed time-freq masks.
+
+    This task may produce output with shared datasets. Be warned that
+    this can produce unexpected outputs if not properly taken into
+    account.
+
+    Attributes
+    ----------
+    share : {"all", "none", "vis", "map"}
+        Which datasets should we share with the input. If "none" we create a
+        full copy of the data, if "vis" or "map" we create a copy only of the modified
+        weight dataset and the unmodified vis dataset is shared, if "all" we
+        modify in place and return the input container.
+    """
+
+    share = config.enum(["none", "vis", "map", "all"], default="all")
+
+    @overload
+    def process(
+        self, data: containers.TimeStream, mask: containers.BaselineMask
+    ) -> containers.TimeStream:
+        ...
+
+    @overload
+    def process(
+        self, data: containers.SiderealStream, mask: containers.SiderealBaselineMask
+    ) -> containers.SiderealStream:
+        ...
+
+    def process(self, data, mask):
+        """Flag data by zeroing the weights.
+
+        Parameters
+        ----------
+        data
+            Data to apply mask to. Must have a stack axis
+        mask
+            A baseline-dependent mask
+
+        Returns
+        -------
+        data
+            The masked data. Masking is done in place.
+        """
+
+        if isinstance(mask, containers.BaselineMask):
+            if not hasattr(data, "time"):
+                raise TypeError(f"Expected a timestream-like type. Got {type(data)}.")
+
+            if not (data.time.shape == mask.time.shape) and np.allclose(
+                data.time, mask.time
+            ):
+                raise ValueError("timestream and mask have different time axes.")
+
+        elif isinstance(mask, containers.SiderealBaselineMask):
+            if not hasattr(data, "ra"):
+                raise TypeError(
+                    f"Expected a sidereal stream like type. Got {type(data)}."
+                )
+
+            if not (data.ra.shape == mask.ra.shape) and np.allclose(data.ra, mask.ra):
+                raise ValueError("sidereal stream and mask have different RA axes.")
+
+        else:
+            raise TypeError(
+                f"Require a BaselineMask or SiderealBaselineMask. Got {type(mask)}."
+            )
+
+        # Validate remaining axes
+        if not np.array_equal(data.stack, mask.stack):
+            raise ValueError("data and mask have different baseline axes.")
+
+        if not (data.freq.shape == mask.freq.shape) and np.allclose(
+            data.freq, mask.freq
+        ):
+            raise ValueError("data and mask have different freq axes.")
+
+        if self.share == "all":
+            tsc = data
+        elif self.share == "vis":
+            tsc = data.copy(shared=("vis",))
+        elif self.share == "map":
+            tsc = data.copy(shared=("map",))
+        else:
+            tsc = data.copy()
+
+        tsc.weight[:] *= (~mask.mask[:]).astype(np.float32)
+
+        return tsc
+
+
 class MaskFreq(task.SingleTask):
     """Make a mask for certain frequencies.
 


draco/analysis/flagging.py
```diff
@@ -1361,6 +1361,95 @@ def process(self, tstream, rfimask):
...
+        """Flag data by zeroing the weights.
+
+        Parameters
+        ----------
+        data : timestream or sidereal stream
+            A timestream or sidereal stream like container. For example,
+            `containers.TimeStream`, `andata.CorrData` or
+            `containers.SiderealStream`.
+        mask : containers.BaselineMask  or containers.SiderealBaselineMask
+            An RFI mask for the same period of time.
```
sjforeman, line 1395: More accurate:
```suggestion
            A baseline-dependent mask for the same period of time.
```

draco/analysis/flagging.py
```diff
@@ -1361,6 +1361,95 @@ def process(self, tstream, rfimask):
...
+        -------
+        data : timestream or sidereal stream
+            The masked data. Masking is done in place.
+        """
+
+        if isinstance(mask, containers.BaselineMask):
+            if not hasattr(data, "time"):
+                raise TypeError(f"Expected a timestream-like type. Got {type(data)}.")
+
+            if not np.array_equal(data.time, mask.time):
```
sjforeman, line 1407: `np.array_equal` can have issues when comparing floating-point numbers, so `np.allclose` is probably safer
```suggestion
            if not np.allclose(data.time, mask.time):
```

draco/analysis/flagging.py
```diff
@@ -1361,6 +1361,95 @@ def process(self, tstream, rfimask):
...
+                raise ValueError("timestream and mask have different time axes.")
+
+        elif isinstance(mask, containers.SiderealBaselineMask):
+            if not hasattr(data, "ra"):
+                raise TypeError(
+                    f"Expected a sidereal stream like type. Got {type(data)}."
+                )
+
+            if not np.array_equal(data.ra, mask.ra):
```
sjforeman, line 1416: As above:
```suggestion
            if not np.allclose(data.ra, mask.ra):
```

draco/analysis/flagging.py
```diff
@@ -1361,6 +1361,95 @@ def process(self, tstream, rfimask):
...
+                raise ValueError("sidereal stream and mask have different RA axes.")
+
+        else:
+            raise TypeError(
+                f"Require a BaselineMask or SiderealBaselineMask. Git {type(mask)}."
```
sjforeman, line 1421: Typo:
```suggestion
                f"Require a BaselineMask or SiderealBaselineMask. Got {type(mask)}."
```

draco/analysis/flagging.py
```diff
@@ -1361,6 +1361,95 @@ def process(self, tstream, rfimask):
...
+            )
+
+        # Validate remaining axes
+        if not np.array_equal(data.stack, mask.stack):
```
sjforeman, line 1425: `array_equal` is probably fine here, since the `stack` axis should be integers, but might as well change it for consistency:
```suggestion
        if not np.allclose(data.stack, mask.stack):
```

draco/analysis/flagging.py
```diff
@@ -1361,6 +1361,95 @@ def process(self, tstream, rfimask):
...
+            raise ValueError("data and mask have different baseline axes.")
+
+        if not np.array_equal(data.freq, mask.freq):
```
sjforeman, line 1428:
```suggestion
        if not np.allclose(data.freq, mask.freq):
```

sjforeman (changes requested):
@ljgray , apologies for not remembering earlier that the old `ThresholdVisWeightBaselineAlt` applied baseline-dependent masking in-place - I could have potentially saved a few rounds of revisions here. Anyway, I only had a few minor requests here.

I recall @jrs65 being reluctant to introduce functionality for applying baseline-dependent masks, but since this new task is only meant to reproduce what we had already been doing with `ThresholdVisWeightBaselineAlt`, hopefully he won't object...

ljgray:
> @ljgray , apologies for not remembering earlier that the old `ThresholdVisWeightBaselineAlt` applied baseline-dependent masking in-place - I could have potentially saved a few rounds of revisions here. Anyway, I only had a few minor requests here.
> 
> I recall @jrs65 being reluctant to introduce functionality for applying baseline-dependent masks, but since this new task is only meant to reproduce what we had already been doing with `ThresholdVisWeightBaselineAlt`, hopefully he won't object...

If we don't want to add a task to apply a baseline-dependent mask, we could also just add an `in-place` flag to `ThresholdVisWeightBaseline` and do it there. I'm not a huge fan of having a task that could return a mask container or a sidereal stream container, so I figure that this implementation is cleaner, but I'll discuss with @jrs65 before merging.

sjforeman:
> If we don't want to add a task to apply a baseline-dependent mask, we could also just add an `in-place` flag to `ThresholdVisWeightBaseline` and do it there. I'm not a huge fan of having a task that could return a mask container or a sidereal stream container, so I figure that this implementation is cleaner, but I'll discuss with @jrs65 before merging.

I agree with you: I remember considering an in-place flag when I refactored `ThresholdVisWeightBaseline`, but decided (possibly in consultation with others, I can't remember) that outputting the mask container was cleaner.

draco/analysis/flagging.py
ljgray: The benefit of `array_equal` is that it also checks for equal shape, whereas `allclose` assumes that the shape are the same and will throw a different exception if the shapes don't match. I agree though that we should probably use it for this sort of comparison, so I'm also checking for equal shape separately

draco/analysis/flagging.py
```diff
@@ -1464,6 +1464,111 @@ def process(self, tstream, rfimask):
...
+            raise ValueError("data and mask have different baseline axes.")
+
+        if not (data.freq.shape == mask.freq.shape) and np.allclose(
+            data.freq, mask.freq
+        ):
+            raise ValueError("data and mask have different freq axes.")
+
+        # Distribute both across baselines
+        data.redistribute("stack")
+        mask.redistribute("stack")
```
ljgray, line 1552: This actually might not be needed

draco/analysis/flagging.py
```diff
@@ -1464,6 +1464,111 @@ def process(self, tstream, rfimask):
...
+            if not (data.ra.shape == mask.ra.shape) and np.allclose(data.ra, mask.ra):
+                raise ValueError("sidereal stream and mask have different RA axes.")
+
+        else:
+            raise TypeError(
+                f"Require a BaselineMask or SiderealBaselineMask. Got {type(mask)}."
+            )
+
+        # Validate remaining axes
+        if not (data.stack.shape == mask.stack.shape) and np.allclose(
```
jrs65, line 1540: Is `allclose` appropriate here? The contents shouldn't have limited precision so they should be exactly equal.

draco/analysis/flagging.py
```diff
@@ -1464,6 +1464,111 @@ def process(self, tstream, rfimask):
...
+            data.stack, mask.stack
+        ):
+            raise ValueError("data and mask have different baseline axes.")
+
+        if not (data.freq.shape == mask.freq.shape) and np.allclose(
```
jrs65, line 1545: `allclose` probably is appropriate for the frequencies though.

draco/analysis/flagging.py
jrs65: Agreed. I think if mask is the same shape there's not need to redistribute at all (on both entry and exit)

draco/analysis/flagging.py
```diff
@@ -1464,6 +1464,111 @@ def process(self, tstream, rfimask):
...
+        if self.share == "all":
+            tsc = data
+        elif self.share == "vis":
+            tsc = data.copy(shared=("vis",))
+        elif self.share == "map":
+            tsc = data.copy(shared=("map",))
+        else:
+            tsc = data.copy()
+
+        tsc.weight[:].local_array[:] *= (~mask.mask[:].local_array[:]).astype(
```
jrs65, line 1563: I believe this should with fine without the explicit local array calls.
```suggestion
        tsc.weight[:] *= (~mask.mask[:]).astype(np.float32)
```

jrs65 (changes requested):
Putting aside my objections to baseline-freq-time dependent masking, this looks good to me!

draco/analysis/flagging.py
ljgray: True, this one probably doesn't have to use `allclose`

ljgray:
> Putting aside my objections to baseline-freq-time dependent masking, this looks good to me!

Ideally we can figure out a way to avoid having any baseline-dependent masking

jrs65 (approved):

PR #231, opened 2023-03-20
feat(containers): remove '.copy' method from ContainerBase.

ljgray:
This method has been implemented in 'MemDiskGroup' in 'caput.memh5', so there is no need to overwrite it here now.

Relies on radiocosmology/caput#239 to properly implement shared datasets.

### DIFF
--- a/draco/core/containers.py
+++ b/draco/core/containers.py
@@ -499,62 +499,6 @@ class ContainerBase(memh5.BasicCont):
 
         return selections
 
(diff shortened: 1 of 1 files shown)


ljgray:
@ketiltrout This should now be fine after merging radiocosmology/caput#239

ketiltrout (approved):
I think this looks fine.  Certainly, the draco test suite seems sufficient to let us know if this broke something.

PR #233, opened 2023-03-23
perf(task): only share across ranks if nan is found

ljgray:


ljgray:
Interestingly this actually slowed things down, so closing

jrs65:
I think this was likely to break too if any NaN's were actually found. All ranks need to participate in the collective `allreduce` call, but ones where `found == False` would skip it which would usually lead to everything hanging.

ljgray:
> I think this was likely to break too if any NaN's were actually found. All ranks need to participate in the collective `allreduce` call, but ones where `found == False` would skip it which would usually lead to everything hanging.

Yeah, that would definitely happen. Most of the time here is actually spent at the first barrier rather than in the actual `allreduce` call anyway

jrs65:
My guess would be that is because it's taking a while for the different ranks to sync up as some of them are taking a long time to NaN check. It might be worth trying to rewrite and parallelise the NaN/Inf check with Cython.

ljgray:
At the very least I want to remove the `.sum()` call unless a nan/inf is actually found, just calling `np.isnan(x).any()` to check should be considerably faster

PR #235, opened 2023-03-31
Use ruff for CI linting

ljgray:
Similar to radiocosmology/caput#238

Also moves the unit tests for `invert_no_zero` to caput, since that method was moved previously. An associated PR can be found at radiocosmology/caput#240


draco/analysis/flagging.py
```diff
@@ -132,10 +131,12 @@ def process(self, mmodes):
         Parameters
         ----------
         mmodes : containers.MModes
+            Mmode container to mask
 
         Returns
         -------
         mmodes : containers.MModes
+            Same object as input with maskig applied
```
jrs65, line 139:
```suggestion
            Same object as input with masking applied
```

jrs65 (approved):
Looks good. Just spotted one typo.

draco/analysis/flagging.py
ljgray: Cool, I'll add that

ljgray:
This PR seems to be breaking the sphinx autosummary...

PR #236, opened 2023-04-03
fix(flagging): increase default threshold for static weight masking

ljgray:
Since this is just supposed to catch numerical issues, I'm moving the default thresholds much closer to the max/min float32 values.

I'm also interested in feedback about making this baseline-independent and just spitting out  a mask, since it's adding some baseline-dependent flagging at the moment.

### DIFF
--- a/draco/analysis/flagging.py
+++ b/draco/analysis/flagging.py
@@ -627,8 +627,8 @@ class SanitizeWeights(task.SingleTask):
         smallest value to keep
     """
 
-    max_thresh = config.Property(proptype=np.float32, default=1e10)
-    min_thresh = config.Property(proptype=np.float32, default=1e-10)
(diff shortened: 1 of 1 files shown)


ljgray:
Reverting to a draft because there's something else I want to try

sjforeman:
The change in the default values seems fine, since the new defaults are used in the rev07 daily config anyway.

I don't have any opinions about making `SanitizeWeights` baseline-independent, since I don't know what fraction of the data the current `SanitizeWeights` typically catches, or whether the flagged samples are localized in frequency or time (vs. spread out more randomly, which would incur a larger loss of data by moving to a baseline-independent scheme). I'd be fine with punting that issue to the future...

sjforeman (approved):

PR #238, opened 2023-04-11
feat(beam): use GridBeam in telescope coordinates

ssiegelx:
Allows the user to generate a HybridVisStream from a GridBeam in telescope coordinates.  This avoids problems at the pole when using GridBeams in celestial coordinates that do not span 360 degrees in hour angle.

### DIFF
--- a/draco/analysis/beam.py
+++ b/draco/analysis/beam.py
@@ -16,6 +16,7 @@ Tasks
 
 import numpy as np
 import scipy.constants
+import scipy.interpolate
 
 from caput import interferometry
 
@@ -45,6 +46,9 @@ class CreateBeamStream(task.SingleTask):
             % (self.telescope.latitude, self.telescope.rotation_angle)
         )
 
+        self.latitude = np.radians(self.telescope.latitude)
+        self.rotation_angle = np.radians(self.telescope.rotation_angle)
+
     def process(self, data, beam):
         """Convert the beam model into a format that can be deconvolved from data.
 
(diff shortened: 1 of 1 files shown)


tristpinsm:
Here are plots generated using the telescope beam input to `CreateBeamStream`. The ringing near the NCP is a lot more localised in declination, but still present. Also, whereas before the NCP itself appeared to be masked, that no longer seems to be the case.
mmode_beam_477_Y
mmode_beam_477_X

ssiegelx:
Thanks @tristpinsm.  I think the remaining ringing in declination is due to the |hour angle| < 90 degree limit that is placed to ignore the antipodal transit.  There is roughly +/- 2.5 degrees around the NCP where the beam response is significant at 90 degrees.

I'm not sure there are any great options as that assumptions that are made in the beam generation begin to break down as you approach the NCP.  Perhaps we should just set the weight dataset for the beam stream to zero for these declinations?

tristpinsm:
> I'm not sure there are any great options as that assumptions that are made in the beam generation begin to break down as you approach the NCP. Perhaps we should just set the weight dataset for the beam stream to zero for these declinations?

Yeah that seems reasonable, there is not much sky area there anyway...

draco/analysis/beam.py
```diff
@@ -119,30 +125,24 @@ def process(self, data, beam):
...
 
         # Transpose the first two dimensions from (freq, pol) to (pol, freq)
         bweight = bweight.swapaxes(0, 1)
-        bvis = beam.beam[:].swapaxes(0, 1)
+        bvis = beam.beam[:].local_array.swapaxes(0, 1)
 
         # Create output container
         out = containers.HybridVisStream(
-            ra=nra,
+            ra=ra,
```
tristpinsm, line 145: In the celestial beam case, `ra` is defined based on the `phi` axis of the beam container, which doesn't span the full RA range and causes a crash later on. I will fix this and set the test job running again.

draco/analysis/beam.py
```diff
@@ -69,48 +73,50 @@ def process(self, data, beam):
...
+        if beam.coords == "celestial":
+            dec = np.radians(beam.theta)
+            el_beam = np.sin(dec - self.latitude)
 
-        el_beam = np.sin(np.radians(dec - self.telescope.latitude))
-        el_data = data.index_map["el"]
+            ha = beam.phi
+            ra = (ha + 360.0) % 360.0
+            nra = int(round(360.0 / np.abs(ha[1] - ha[0])))
+            delta_ra = 360.0 / nra
```
tristpinsm, line 86:
```suggestion
            ra_beam = (ha + 360.0) % 360.0
            nra = int(round(360.0 / np.abs(ha[1] - ha[0])))
            delta_ra = 360.0 / nra
            ra = np.arange(nra) * delta_ra
```

With this change (and propagating `ra_beam` in the next few lines) the code ran to completion and I confirmed that the result is the same as with the previous version by comparing the ringmap and weights for about 20 unflagged randomly selected frequencies in the lower band.

ssiegelx:
@Arnab-half-blood-prince This PR enables `CreateBeamStream` to handle `GridBeam` containers in telescope coordinates.  Unfortunately it never got merged and I don't remember the status of testing.  Do you want to try running it?

tristpinsm:
I can push the changes I proposed here, and I think with those it was working well.

PR #241, opened 2023-04-27
feat(transform): task to apply reduction ops to a container

ljgray:
Apply a reduction op across up to n-1 axes of a rank-n dataset. The reduction can be applied to more than one dataset in the container, and axis downselection can be applied to n-1 axes as well. A weight mask can optionally be generated, which masks values where the weights are zero and is broadcast to the dataset shape if possible. This mask can also be applied before doing the reduction, using a masked array.

The task tries to optimize redistribution by first trying to minimize the number of redistributions that happen and then trying to choose the best possible axes to redistribute over if needed. This is just being done by an inverse sort by axis length.

I've tested this a fair bit, but I wouldn't be surprised if there are still some edge cases that could break. It generally works as expected when being used to generate variance-over-freq and variance-over-el datasets in the chime daily pipeline.

Requires radiocosmology/caput#242

### DIFF
--- a/draco/analysis/transform.py
+++ b/draco/analysis/transform.py
@@ -2,11 +2,12 @@
 
 This includes grouping frequencies and products to performing the m-mode transform.
 """
-from typing import Optional, Union, overload
+from typing import Optional, Union, Tuple, overload
 
 import numpy as np
 from numpy.lib.recfunctions import structured_to_unstructured
 from caput import mpiarray, config
+from caput.tools import invert_no_zero
 
 from ..core import containers, task, io
 from ..util import tools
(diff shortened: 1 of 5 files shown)


draco/analysis/transform.py
```diff
@@ -1265,3 +1265,235 @@ def process_finish(self) -> Union[containers.SiderealStream, containers.RingMap]
...
+        op = getattr(np, self.op)
+
+        # Order axes based on decreasing length to help decide the best
+        # redistribution options
+        ax_priority = sorted(
+            data.index_map.keys(), key=lambda x: len(data.index_map[x]), reverse=True
+        )
+
+        def _select_axes(options: list, current_ax: int) -> tuple:
+            """Select the best distriuted axes for selection and reduction."""
```
sjforeman, line 1337: Typo:
```suggestion
            """Select the best distributed axes for selection and reduction."""
```

draco/analysis/transform.py
```diff
@@ -1265,3 +1265,235 @@ def process_finish(self) -> Union[containers.SiderealStream, containers.RingMap]
...
+        out.attrs["reduced"] = True
+        out.attrs["reduction_op"] = self.op
+        out.attrs["reduction_axes"] = np.array(self.axes)
+        out.attrs["downselections"] = self.selections
+        out.attrs["reduced_datasets"] = np.array(self.datasets)
+
+        # Iterate over the datasets and apply operation
+        for name, ds in data.datasets.items():
+            if name not in self.datasets:
+                # Just set this dataset to be zero if it was initialized
```
sjforeman, line 1410: I'm confused about this block: I would have thought that the `copy_from=data` option when `out` was created implies that datasets will be copied from `data` to `out`, so if there's a dataset that's in `out` but not in `self.datasets`, we'd want to leave it alone since it's not being reduced. Am I misunderstanding the logic?

draco/analysis/transform.py
```diff
@@ -1265,3 +1265,235 @@ def process_finish(self) -> Union[containers.SiderealStream, containers.RingMap]
...
+    op : str
+        Reduction operation to apply. Must be the name of a numpy
+        attribute, ex. "var". Default is "mean".
+    generate_mask : bool, optional
+        If true, generate a boolean mask to remove samples that are flagged.
+        When this is set, reduction can only be applied to datasets with
+        the same axes as the weights.
+    apply_mask : bool, optional
+        Whether to apply the weight mask to the datasets before reduction.
+        This won't do anything unless `weight_mask` is True.
```
sjforeman, line 1301: I think you mean:
```suggestion
        This won't do anything unless `generate_mask` is True.
```

draco/analysis/transform.py
sjforeman: I think I may understand now: if there's a dataset that is not being reduced, we don't care about preserving it in the reduced container, so we just set its entries to zero. Is that right? If so, it would be helpful to say that in the comment

draco/analysis/transform.py
```diff
@@ -1265,3 +1265,235 @@ def process_finish(self) -> Union[containers.SiderealStream, containers.RingMap]
...
+
+            out[name][:] = reduced[:]
+
+            # Add the weights back in. If a weight mask was used, the resulting
+            # weights array will be the same for each dataset, so we don't have
+            # to care about overwriting it each time (other than performance)
+            if self.generate_mask:
+                out.weight[:] = weight[:]
+            else:
+                out.weight[:] = 1
```
sjforeman, line 1497: We just set the weights to unity if we're not using a mask because we don't care about the weights in the reduced dataset? Is that right?

sjforeman (commented):
I don't see any obvious flaws in this implementation (modulo a few points where I got confused and wrote comments), but the logic is sufficiently complicated that it would be good to have someone else take a look

draco/analysis/transform.py
ljgray: My idea was exactly what you said, if we aren't reducing a dataset then we don't care about preserving it, but in that case it probably just shouldn't be initialized in the first place. I think I can tweak the code a bit to make that more obvious, and add a comment + docstring entry.

draco/analysis/transform.py
ljgray: That's the idea, but I'm very open to thoughts about how best to handle the weights through a reduction like this, since it's a bit convoluted and maybe not super useful the way it's done here

draco/analysis/transform.py
sjforeman: Got it, thanks!

draco/analysis/transform.py
sjforeman: Hmm, it would depend on what we might want to do with the weights in a reduced dataset. You could consider allowing the user to also specify a reduction operation to be applied to the weights, in case we want to retain some of that information when examining the corresponding reduced dataset...but this could also wait until we've spent some time working with reduced products, so we know what would be useful

ljgray: I've done a fair bit of a rework with how the weights are handled, which I think should make more sense and be a bit more useful

draco/analysis/transform.py
```diff
@@ -1265,3 +1266,299 @@ def process_finish(self) -> Union[containers.SiderealStream, containers.RingMap]
...
+                    mask = np.broadcast_to(weight, arr.shape, subok=False)
+                except ValueError as e:
+                    raise ValueError(
+                        f"Could not broadcast weights. Got dataset {name} with "
+                        f"shape {arr.shape} and weight with shape {weight.shape}."
+                    ) from e
+
+            if name in self.apply_weight:
+                # Apply the weights to the dataset
+                arr = arr * mask
```
jrs65, line 1547: I think this weighting isn't quite what we want. For reductions that are essentially summation you want to do something like:
```
arr = arr * weight / weight.sum(axis=apply_over) * arr.shape[apply_over]
```

jrs65 (changes requested):
I'm going to echo Simon's comments here. This is nice, and I think could be broadly useful, but I fear it might be too smart for its own good. The actual task we want to do (variance over freq, variance over elevation) could be implemented in probably < 20 lines. This generic version is 300, and might not be flexible enough for what we want.

One of my fears is that it's pretty hard to specify a sensible generic reduction of the data when you would like to be able to use different weightings (unweighted, masked, and weighted) and have things turn out sensibly. Some of the common reductions have fairly sensible definitions (e.g. mean, variance), but others don't, i.e. how do you do a weighted `prod`, or a weighted `min`?

I'm not sure how exactly to go about simplifying this. One thing I might suggest is defining a common API, like:

```python
def reduction(data: np.ndarray, weight: np.ndarray, axis: int, weighting: Literal["none", "masked", "weighted"]) -> tuple[np.ndarray, np.ndarray]:
    """Perform a reduction for data and weight to return a reduced data and weight."""
    ...
```
and then having specific implementations for each of a more limited set of operations you actually want to support and be well defined.

Something like:

```python
def reduction_var(data: np.ndarray, weight: np.ndarray, axis: int, weighting: Literal["none", "masked", "weighted"]) -> tuple[np.ndarray, np.ndarray]:
    if weighting == "none":
        v = data.var(axis=axis)
    else:
        if weighting == "masked":
            w = (weight > 0).astype(weight.dtype)
        else:
            w = weight.copy()
        w *= w.sum(axis=axis) / w.shape[axis]
        v = np.var(w**0.5 * v, axis=axis)
    return v, np.ones_like(v)
```
That above is a little ugly, so maybe there's a better approach, but I imagine you get the vague idea.

draco/analysis/transform.py
```diff
@@ -1265,3 +1266,343 @@ def process_finish(self) -> Union[containers.SiderealStream, containers.RingMap]
         self.mixed_data = None
 
         return data
+
+
+class Downselect(task.SingleTask):
```
jrs65, line 1271: You might consider merging some of the internals of this into `draco.core.containers.copy_datasets_filter` to generalise this (that one only works on a single axis).

draco/analysis/transform.py
ljgray: I've cleaned this up a fair bit. I'm not 100% sure on the weighting for a multi-axis variance, so it would helpful to check that.

draco/analysis/transform.py
ljgray: That's a good idea but I might push it off until post-revision_07

sjforeman (approved):
Looks good to me! Although it could still probably use a once-over from someone else, just to be sure...

draco/analysis/transform.py
```diff
@@ -1265,3 +1257,213 @@ def process_finish(self) -> Union[containers.SiderealStream, containers.RingMap]
...
+
+    Attributes
+    ----------
+    selections : dict
+        Dict formatted as {<ax_name>:<selection>,} for any axes
+        that should be downselected before applying the reduction.
+        Selections can be either slices or a list/tuple of indices.
+    """
+
+    selections = config.Property(proptype=dict)
```
jrs65, line 1279: I don't think we should reinvent another way of specifying the `selections` in here. Just re-use what's available in `draco.core.io.BaseLoadFiles`. You'll need to find some way of using the `_resolve_sel` method here, cleanest would be moving it all out into a `SelectionsMixin` class that both inherit from, but tbh you could just inherit from `BaseLoadFiles` and ignore all it's other functionality.

draco/analysis/transform.py
```diff
@@ -1265,3 +1257,213 @@ def process_finish(self) -> Union[containers.SiderealStream, containers.RingMap]
...
+        # Redistribute bcak to the original axis, again using the axis name
+        out.redistribute(ds_axes[original_ax_id])
+
+        return out
+
+    def _make_output_container(
+        self, data: containers.ContainerBase
+    ) -> containers.ContainerBase:
+        """Create the output container."""
+        output_axes = {ax: np.array([-1], dtype=int) for ax in self.axes}
```
jrs65, line 1418: Why -1? Something like the mean of the reduced axis might be more meaningful.

draco/core/containers.py
```diff
@@ -2775,32 +2774,60 @@ def empty_timestream(**kwargs):
...
+    # Try to clean up selections
+    for ax in list(selection):
+        sel = selection[ax]
+        # Remove any unnecessary slices
+        if sel == slice(None):
+            del selection[ax]
+        # Convert any indexed selections to slices where possible
+        elif type(sel) in {list, tuple, np.ndarray}:
+            if list(sel) == list(range(sel[0], sel[-1])):
+                selection[ax] = slice(sel[0], sel[-1])
```
jrs65, line 2830: Nice. Good idea to do this.

test/test_containers.py
```diff
@@ -81,3 +81,48 @@ def test_copy(ss_container):
...
+
+def test_copy_filter(ss_container):
+    """Test copying datasets between container while filtering an axis.
+
+    This test should ensure that slices and selections can be appplied to
+    various axes properly, that arguments can be passed correctly, and that
+    the function fails as expected when bad arguments are provided.
+    """
+    new = containers.SiderealStream(axes_from=ss_container, attrs_from=ss_container)
+
```
jrs65, line 94: You should test a multi-axis selection that should succeed in here.

jrs65 (changes requested):

draco/analysis/transform.py
ljgray: This was just an arbitrary choice to make a length 1 axis that explicitly doesn't have meaning. I'm not convinced that something like a mean would be better - for example, if we took a variance over RA then the RA axis mean would be 180 deg, and I'm not sure if that's really meaningful (and maybe confusing since the data wouldn't actually correspond to that position). It would probably be better to use something a bit more explicit though.

ljgray: Swapped to using the first value of the original index map by default after some discussion

ketiltrout (approved):
I think this looks okay.  It might be worth adding a comment to `_make_output_container` pointing out the choice of `index_map[0]` for the value of the rolled-up axis.

PR #243, opened 2023-05-05
ci(ruff): add extra linting rules: isort, pyupgrade, numpy, ruff-specific

ljgray:
These additions are useful for improving overall quality and consistency of code and eliminating extra things for PR reviewers to check. The ruff python version is set to python 3.8, since this is the minimum supported version. 

- isort: formats and alphabetically orders imports
- pyupgrade: upgrade syntax for newer versions of python (based on the version specified in pyproject.toml)
- numpy: check deprecated types
- ruff-specific: ambiguous unicode, collection instead of concatenation
- flake8-blind-except: don't allow blind `except` statements
- flake8-comprehensions: use comprehensions instead of concatenation where possible
- flake8-return: consistent and clear return statements

Additional ignored rules:
- NPY002: replace legacy numpy.random calls with np.random.Generator

I'm happy to add more rules if there's anything else of interest

### DIFF
--- a/draco/synthesis/mockcatalog.py
+++ b/draco/synthesis/mockcatalog.py
@@ -84,16 +84,15 @@ Below is an example workflow:
 ...
 """
 
-import numpy as np
 import healpy as hp
+import numpy as np
 import scipy.stats
-
+from caput import config, mpiarray, mpiutil
 from cora.util import units
-from caput import config
-from caput import mpiarray, mpiutil
-from ..core import task, containers
-from ..util import random, tools
 from mpi4py import MPI
+
+from ..core import containers, task
+from ..util import random, tools
 
 # Constants
 C = units.c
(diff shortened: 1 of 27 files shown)


jrs65:
I don't think you can use the | syntax instead on Union until 3.10.

ljgray:
Good catch. This PR will probably look pretty messy until I nail down exactly which rules to use.

ljgray:
> I don't think you can use the | syntax instead on Union until 3.10.

Actually, in hindsight this is exactly what `from __future__ import annotations` is for, so we could potentially include that everywhere it's needed

draco/synthesis/mockcatalog.py
```diff
@@ -1079,7 +1071,7 @@ def elg_velocity_error(z, rng):
         percentiles:
             "Additionally, we can assess with repeats that 99.5, 95, and 50
             percent of our redshift estimates have a precision better than
-            300 km s−1, 100 km s−1, and 20 km s−1, respectively."
```
ketiltrout, line 1082: Why is this converting minus signs to dashes?  Is unicode not allowed in docstrings?

If the − really isn't wanted, it might be better to just write "km/s" here.

ketiltrout:
> > I don't think you can use the | syntax instead on Union until 3.10.
> 
> Actually, in hindsight this is exactly what `from __future__ import annotations` is for, so we could potentially include that everywhere it's needed

`from __future__ import annotations` causes Python to treat annotations as if they were strings.  So, while this import allows using | syntax when _running_ in pre-3.10 Python, a pre-3.10 type checker will still break when encountering | instead of `Union`.

ljgray:
@ketiltrout @jrs65 I've done another run through to make sure everything is working as expected here, so I want to ping you both for review again.

jrs65 (approved):

PR #245, opened 2023-05-08
Ac/power spectrum map

Arnab-half-blood-prince:
This will estimate power spectrum (3D,2D,1D) from a ringmap. The short descriptions of tasks are mentioned below:

1. TransformJyPerBeamToKelvin : This task will convert a ringmap in Jy/beam unit to Kelvin unit. It will estimate psf solid angle based on longest baseline and will be same for both XX and YY pol. 
2. DelayTransformMapFFT : Transform the ringmap from freq to delay by just taking FFT. It will only work when the map is foreground filtered and close to noise. We can also estimate the delay spectrum via Gibbs+Wiener method and pass the delay spectrum to the next step (3).
 3. DelayTransformMapFFT : Spatial transform the delay transformed map from (RA,DEC) to (u,v) domain, by taking 2D FFT.
4. CrossPowerSpectrum3D : Estimate 3D cross power spectrum of two data cubes in (delay,kx,ky) domain.
6. AutoPowerSpectrum3D : Estimate the 3D auto power spectrum of a data cube in (delay,kx,ky) domain.
7. CylindricalPowerSpectrum2D : Estimate the cylindrically averaged 2D power spectrum in (kpar,kperp) domain
8. SphericalPowerSpectrum2Dto1D: Estimate the spherically averaged 1D power spectrum from a 2D cylindrically averaged power spectrum.
9. SphericalPowerSpectrum2Dto1D: Estimate the spherically averaged 1D power spectrum from a 3D power spectrum cube.
This task is for check. The 1D estimate from 3D to 1D or 2D to 1D should match. I checked that and it is consistent.

A example run on HI signal is stored here : /project/rpp-chime/arnab92/work/21cm/halofit_new/ringmap/masked_map/ps_HI
Follow the config file there. It will take ~3mins to run. One can also just run inside a login node by caput-pipeline run *.yaml,  for time saving.


draco/analysis/powerspec.py
```diff
@@ -0,0 +1,1568 @@
...
+        vis_cube.data_tau_uv[:] = 0.0
+
+        # Save the central freq of the band and the
+        # corresponding redshift and the cube volume in the attrs
+        vis_cube.attrs["freq_center"] = nu_c
+        vis_cube.attrs["redshift"] = redshift
+        vis_cube.attrs["volume"] = vol_cube
+
+        # Create index map for (u,v) coordinates and k_parallel
+        fourier_axes = ["u", "v", "k_parallel", "uv_mask"]
```
ljgray, line 364: I mention this in the `containers` file as well, but I think that these should actually be datasets in the container rather than index maps, since two containers with the same `kx` `ky` axes could have different `u` and `v` depending on the cosmology used (I think).

It's also cleaner since `index_map` entries are generally expected to be used by at least one dataset, so having some extra axes doesn't fit into the container code design as well.

draco/analysis/powerspec.py
```diff
@@ -0,0 +1,1568 @@
...
+        for ff in fourier_axes:
+            vis_cube.create_index_map(ff, locals()[ff])
+
+        # Save the spatial window as an attribute
+        if self.apply_spatial_window:
+            vis_cube.attrs["window_spatial"] = self.spatial_window
```
ljgray, line 370: I think this can also be a dataset instead of an attribute.

draco/analysis/powerspec.py
```diff
@@ -0,0 +1,1568 @@
...
+
+        # Spatial transform to Fourier domain
+        # loop over pol
+        for pp, psr in enumerate(pol):
+            self.log.debug(f"Estimating spatial FFT for pol: {psr}")
+            # loop over delays
+            for lde, de in vis_cube.data_tau_uv[:].enumerate(axis=1):
+                slc = (pp, lde, slice(None), slice(None))
+                data = np.ascontiguousarray(data_view.local_array[slc])
+                data_uv, NEB_ra, NEB_dec = image_to_uv(
```
ljgray, line 382: Like mentioned above, this could probably be sped up using `caput.fftw`

draco/analysis/powerspec.py
```diff
@@ -0,0 +1,1568 @@
...
+        # tapering window function used for FFT
+        volume_cube = data_1.attrs["volume"]
+        NEB_freq = data_1.attrs["effective_bandwidth"]
+        NEB_ra = data_1.attrs["effective_ra"]
+        NEB_dec = data_1.attrs["effective_dec"]
+        NEB = 1 / (NEB_freq * NEB_ra * NEB_dec)
+        ps_norm = volume_cube * NEB
+
+        # Initialise the 3D power spectrum container
+        ps_cube = containers.Powerspec3D(
```
ljgray, line 461: This could just be `axes_from=data_1`

draco/analysis/powerspec.py
```diff
@@ -0,0 +1,1568 @@
...
+        )
+        ps_cube.redistribute("delay")
+        ps_cube.ps3D[:] = 0.0
+
+        # Save the power spectrum normalization factor in the attrs
+        ps_cube.attrs["ps_norm"] = ps_norm
+
+        # Create index map for (u,v) coordinates,k_parallel axis
+        # and uv-mask
+        fourier_axes = ["u", "v", "k_parallel", "uv_mask"]
```
ljgray, line 472: Same thing as mentioned before that these should probably be datasets

ljgray (changes requested):
I think I've gone through mostly everything for now. Overall looks good, most of my comments are just about code style (naming conventions, etc...) and some minor possible improvements.

draco/analysis/powerspec.py
```diff
@@ -0,0 +1,1568 @@
...
+from draco.util import tools
+
+f21 = units.nu21  # 21cm line frequency in MHz
+C = units.c  # Speed of light in m/s
+# We are setting cosmology to Planck2015 result
+# one can use Planck2018 too
+cosmo = Planck15
+
+# Lat/Lon of CHIME
+_LAT_LON = {
```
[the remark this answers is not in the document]
ssiegelx: I agree that this should be removed.  `SpatialTransformDelayMap` is the only task that uses these variables to convert between el and dec.  I would just add a setup method to that task, which accepts a telescope object, and then get the lat/lon from that object.  This is what we do elsewhere in draco.

draco/analysis/powerspec.py
```diff
@@ -0,0 +1,1568 @@
+"""Power spectrum estimation from ringmap."""
```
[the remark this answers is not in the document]
Arnab-half-blood-prince: I am keeping this new file as it is. This is because the previous `powerspectrum.py` file is generated by Richard and mainly for `m-mode` based power spectrum pipeline. I do not want to mix these two. We may change the name of this file as `powerspecmap.py`

draco/analysis/powerspec.py
```diff
@@ -0,0 +1,1568 @@
...
+                yspec = np.fft.fftshift(np.fft.ifft(data, axis=-1), axes=-1)
+                delay_spectrum.attrs["window_los"] = "None"
+                delay_spectrum.attrs["effective_bandwidth"] = 1.0
+
+            delay_spectrum.spectrum[bi] = yspec
+
+        return delay_spectrum
+
+
+class SpatialTransformDelayMap(task.SingleTask):
```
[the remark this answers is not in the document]
Arnab-half-blood-prince: Yes, I changed all these in the new containers.

draco/analysis/powerspec.py
```diff
@@ -0,0 +1,1568 @@
...
+        # Estimate the uv-mask
+        uv_mask = spatial_mask(
+            kx, ky, self.ew_min, self.ew_max, self.ns_bl, wl.min(), wl.max(), redshift
+        )
+
+        # Estimate the volume of the cube
+        vol_cube = vol_normalization(ra, dec, freq, redshift)
+
+        # Initialise the spatial transformed data cube container
+        vis_cube = containers.SpatialDelayCube(
```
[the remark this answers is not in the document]
Arnab-half-blood-prince: Done

draco/analysis/powerspec.py
Arnab-half-blood-prince: Keeping it as attrs to match with - https://github.com/radiocosmology/draco/blob/d8199ca8cc5d094020c3d50dc01313ea5c9e5523/draco/analysis/delay.py#L1041

draco/analysis/powerspec.py
Arnab-half-blood-prince: As the test shows power spectrum is not taking much time, so keeping it as `np.fft`

draco/core/containers.py
```diff
@@ -2534,6 +2534,229 @@ def freq(self):
...
+        return self.attrs["redshift"]
+
+    @property
+    def freq_center(self):
+        """Get the central frequency attrs."""
+        return self.attrs["freq_center"]
+
+
+class Powerspec3D(ContainerBase):
+    """Container for a 3D power spectrum."""
```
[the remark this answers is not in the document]
Arnab-half-blood-prince: Done

draco/core/containers.py
```diff
@@ -2534,6 +2534,229 @@ def freq(self):
...
+
+    _axes = ("pol", "delay", "kx", "ky")
+
+    _dataset_spec: ClassVar = {
+        "ps3D": {
```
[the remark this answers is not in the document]
Arnab-half-blood-prince: keeping it as it is now. `DelaySpectrum`  is a misnomer as that is actually a power spectrum.

draco/core/containers.py
```diff
@@ -2534,6 +2534,229 @@ def freq(self):
...
+        return self.index_map["v"]
+
+    @property
+    def ps_norm(self):
+        """Get the power spectrum normalizaiton attrs."""
+        return self.attrs["ps_norm"]
+
+
+class Powerspec2D(ContainerBase):
+    """Container for a 2D cylindrically averaged  power spectrum."""
```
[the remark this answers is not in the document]
Arnab-half-blood-prince: I am now changing the names to  `PowerSpectrum3D, PowerSpectrum2D and PowerSpectrum1D` . Primarily again do not want to mix things with m-mode pipeline

Arnab-half-blood-prince (commented):
I have now changed most of these suggestions. 

- I have also added cora.util.cosmology and all the containers will carry the cosmology information.
- I have removed the DelayTransformMapFFT as this can be done with new task in `draco.analysis.delay.DelayTransformFFT`
- Also changed the naming conventions of the containers.

draco/analysis/powerspec.py
```diff
@@ -0,0 +1,1504 @@
...
+    ndec = dec.size
+
+    # Mean resolution in RA and DEC and convert that to radian
+    res_ra_radian = np.deg2rad(np.mean(np.diff(ra)))
+    res_dec_radian = np.deg2rad(np.mean(np.diff(dec)))
+
+    DMz = dRperp_dtheta(redshift, cosmo=cosmo)  # in [h^-1 Mpc]
+
+    # Convert the RA and DEC resolution to Mpc unit
+    d_RA_Mpc = DMz * res_ra_radian
```
ssiegelx, line 1081:
```suggestion
    d_RA_Mpc = DMz * res_ra_radian * np.mean(np.cos(np.deg2rad(dec)))
```
I think you need a factor of cos(dec) in the conversion here.

draco/analysis/powerspec.py
```diff
@@ -0,0 +1,1504 @@
...
+    cosmo: Cosmology object
+        Default is cora.util.Cosmology() default.
+
+    Returns
+    -------
+    comoving transverse distance: float. unit in [h^-1 Mpc].
+    """
+    if cosmo is None:
+        cosmo = get_cosmo()
+    return cosmo.comoving_distance(z)
```
ssiegelx, line 857:
```suggestion
    return cosmo.proper_distance(z)
```
I think this is what you want here, although they should give the same result of Omega_K = 0.

draco/analysis/powerspec.py
```diff
@@ -0,0 +1,1504 @@
...
+
+    # Mean resolution in RA and DEC and convert that to radian
+    res_ra_radian = np.deg2rad(np.mean(np.diff(ra)))
+    res_dec_radian = np.deg2rad(np.mean(np.diff(dec)))
+
+    # Comoving distance
+    DMz = dRperp_dtheta(redshift, cosmo=cosmo)  # in [h^-1 Mpc]
+
+    # Convert the RA and DEC resolution to Mpc unit
+    dx_Mpc = DMz * res_ra_radian
```
ssiegelx, line 1177:
```suggestion
    dx_Mpc = DMz * res_ra_radian  * np.mean(np.cos(np.deg2rad(dec)))
```
As above, I think you need a factor of cos(dec) here.

draco/analysis/powerspec.py
```diff
@@ -0,0 +1,1504 @@
...
+     The weight to use during averaging.
+    axis: int, optional
+     Axis along which the average will be taken.
+
+    Returns
+    -------
+    d_avg : np.ndarray
+     The weighted average.
+    """
+    return np.nansum(d * w, axis=axis) / np.nansum(w, axis=axis)
```
ljgray, line 1207: I think this will be incorrect unless the NaNs exist in the same indices in both `d` and `w`, or the weights are zeroed wherever NaNs exist in `d`. That may be the case wherever it's used, but I'd think about switching to using the `where` keyword in `sum` so you get something like
```python
np.sum(d * w, axis=axis, where=~np.isnan(d))
```

ljgray (approved):
A few comments, but nothing major

draco/core/containers.py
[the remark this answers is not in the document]
ssiegelx: I agree with Liam that the container name `PowerSpectrum3D` already makes it clear that this is a 3D power spectrum, and the axes further reinforce that. Naming the dataset `ps`, `powerspectrum`, or `spectrum` would make it more general and allow for more consistent code across 1D, 2D, and 3D power spectra without requiring a lookup table for dataset names. This would be particularly useful in model-fitting code, where we might fit either 1D or 2D power spectra, and could also help consolidate some of the averaging tasks in the future.

draco/core/containers.py
```diff
@@ -2566,6 +2567,279 @@ def freq(self):
...
+            "initialise": True,
+            "distributed": True,
+        },
+        "mask": {
+            "axes": ["pol", "delay", "kperp"],
+            "dtype": bool,
+            "initialise": True,
+            "distributed": True,
+        },
+        "k_parallel": {
```
ssiegelx, line 2745:
```suggestion
        "kpara": {
```

draco/analysis/powerspec.py
ssiegelx, line 1207: Why do we need to use `nansum` instead of `sum` here? It doesn't seem like NaNs are used to represent missing data.  We're relying on `weight == 0.0`, as done elsewhere in the pipeline. If a NaN appears, wouldn't we want to catch it? That would likely indicate something has gone wrong in these calculations.

draco/analysis/powerspec.py
```diff
@@ -0,0 +1,1504 @@
...
+    bin_indx = np.digitize(ru, bins=kperp_bins)
+
+    # Define empty list to store the binned 2D PS
+    ps_2D = []
+    ps_2D_w = []
+
+    # Now bin in 2D ##
+    with np.errstate(divide="ignore", invalid="ignore"):
+        for i in np.arange(len(kperp_bins) - 1) + 1:
+            p = np.nansum(w[bin_indx == i] * ps_cube[bin_indx == i]) / np.sum(
```
ssiegelx, line 1401:
```suggestion
            p = np.sum(w[bin_indx == i] * ps_cube[bin_indx == i]) / np.sum(
```
Same reason as above.

draco/analysis/powerspec.py
```diff
@@ -0,0 +1,1504 @@
...
+    # averaged power spec and other relevant quantities
+    ps_3D = []
+    ps_3D_err = []
+    k3D = []
+    variance = []
+    # Average the modes falls in each bin
+    with np.errstate(divide="ignore", invalid="ignore"):
+        for i in np.arange(len(kbins) - 1) + 1:
+            w_b = w1D[indices == i]
+            p = np.nansum(w_b * p1D[indices == i]) / np.sum(w_b)
```
ssiegelx, line 1488:
```suggestion
            p = np.sum(w_b * p1D[indices == i]) / np.sum(w_b)
```
Same reason as above.

draco/analysis/powerspec.py
```diff
@@ -0,0 +1,1504 @@
...
+                )
+
+                weight_flat, _, _ = reshape_data_cube(
+                    W, u, v, u_min_lambda, u_max_lambda
+                )
+
+                # Cylindrical 2D power spectrum
+                pspec_2D.ps2D[pp, de], pspec_2D.weight[pp, de] = get_2d_ps(
+                    ps3D_flat[mask_flat],
+                    weight=weight_flat[mask_flat]
```
ssiegelx, line 540: You say you are squaring here to get inverse variance, but the docstring in the `setup` method seems to suggest the quantity being saved to the `weight` container is already a variance.  Is the quantity in the weight container actually a standard deviation?  If so, then the docstring should be updated.  If not, then I think this `**2` needs to be removed.

ssiegelx (commented):
Looks good overall, but have a few more suggestions.

draco/analysis/powerspec.py
Arnab-half-blood-prince: Changed.

draco/analysis/powerspec.py
Arnab-half-blood-prince: Done.

draco/analysis/powerspec.py
Arnab-half-blood-prince: Changed to sum as @ljgray suggested above

draco/core/containers.py
Arnab-half-blood-prince: Changed all these to `kpara`

draco/analysis/powerspec.py
```diff
@@ -0,0 +1,1504 @@
...
+
+        # Genearate an output container
+        out_map = containers.empty_like(rm)
+        out_map.map[:] = 0.0
+        out_map.weight[:] = 0.0
+
+        # store the map after applying the conversion factor
+        out_map.map[:].local_array[:] = (
+            rm.map[:].local_array
+            * factor[np.newaxis, np.newaxis, :, np.newaxis, np.newaxis]
```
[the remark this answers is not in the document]
Arnab-half-blood-prince: Done

draco/analysis/powerspec.py
Arnab-half-blood-prince: I added the cos(dec) term

draco/analysis/powerspec.py
```diff
@@ -0,0 +1,1504 @@
...
+        )
+        out_map.weight[:] = (
+            rm.weight[:].local_array
+            * factor[np.newaxis, :, np.newaxis, np.newaxis] ** 2
+        )
```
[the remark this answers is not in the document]
Arnab-half-blood-prince: Yes, I changed it accordingly

draco/analysis/powerspec.py
```diff
@@ -0,0 +1,1504 @@
...
+        ax_freq = data_axes.index("freq")
+        sfreq = rm.map.local_offset[ax_freq]
+        efreq = sfreq + rm.map.local_shape[ax_freq]
+        local_freq = rm.freq[sfreq:efreq]
+
+        # Estimate the conversion factor
+        factor = jy_per_beam_to_kelvin(local_freq, self.bl_max)
+
+        # Genearate an output container
+        out_map = containers.empty_like(rm)
```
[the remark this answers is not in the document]
Arnab-half-blood-prince: Changed it

draco/analysis/delay.py
```diff
@@ -1036,6 +1036,13 @@ def _create_output(
         # container
         delay_spec.attrs["freq"] = ss.freq
 
+        # Save the frequency tapering window name as an attribute in the output
+        # container. This is needed in the later stage to estimate effective bandwidth.
+        if self.apply_window:
```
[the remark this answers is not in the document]
Arnab-half-blood-prince: Changed

draco/core/containers.py
```diff
@@ -2566,6 +2567,302 @@ def freq(self):
...
+            "initialise": True,
+            "distributed": True,
+        },
+        "kpara": {
+            "axes": ["delay"],
+            "dtype": np.float64,
+            "initialise": True,
+            "distributed": False,
+        },
+        "kperp": {
```
ssiegelx, line 2758: I don't think you need the `kperp` dataset, since you are making `kperp` an axis in this container.  Alternatively, you could make the axis something like `uv`, `uv_dist`, or `uv_radius` for consistency with the 3D spectrum and then keep `kperp` as a dataset.

draco/core/containers.py
```diff
@@ -2566,6 +2567,302 @@ def freq(self):
...
+        """Get the redshift attrs."""
+        return self.attrs["redshift"]
+
+    @property
+    def freq_center(self):
+        """Get the central frequency attrs."""
+        return self.attrs["freq_center"]
+
+
+class PowerSpectrum3D(SpatialDelayCube):
```
ssiegelx, line 2696: This is a question for @ljgray.  If `PowerSpectrum3D` is a subclass of `SpatialDelayCube`, does it inherit all of the datasets?  In other words, will we end up creating an unused `data` array in this container?

ssiegelx (approved):

draco/core/containers.py
Arnab-half-blood-prince: Yes, I checked the datasets present in the `PowerSpectrum3D` container, and it has spectrum dataset as well as `data` dataset from the `SpatialDelayCube`.  That is what it shows in my output PS3D with this pipeline. Is there a way to avoid keeping another extra dataset? I think other than that particular dataset, all other are required in `PowerSpectrum3D`

ssiegelx:
Looks good to me.

draco/core/containers.py
ljgray: Yeah, both `_axes` and `_dataset_spec` are pseudo-inherited through the `axes` and `dataset_spec` properties in `ContainerBase`. Basically, when you access those public properties, it walks back through the class inheritance tree and gets all the axis/dataset_spec entries. However, if you access the private `_axes`/`_dataset_spec` properties, those are just static for the items defined on the child class that you're looking at.

I think these changes are fine. However, just keep in mind that the `data` dataset property is a bit unique because anything that inherits from `DataWeightContainer` can access its main dataset via `.data`. So, it's fine to re-use a dataset called `data`, but only for classes that we don't expect to inherit from any further.

ssiegelx: Okay should we just change this to something else now to avoid that issue if we ever want to make these DataWeight containers? @Arnab-half-blood-prince @ljgray

ljgray: Yeah, that would be a good idea if we think we might want to add weights to these containers in the future. I'm not sure of a better name for this dataset off the top of my head, though.

PR #246, opened 2023-07-23
fix(sidereal): mpiarray issues

jrs65:


draco/analysis/sidereal.py
```diff
@@ -703,8 +703,11 @@ def process(self, sdata):
             self.stack.weight[:] = 0.0
 
             self.count = 0
-            self.Ni_s = np.zeros(
-                (sdata.weight.local_shape[0], sdata.weight.local_shape[2])
+            self.Ni_s = mpiarray.zeros(
```
ljgray, line 706: Does this need to be an mpiarray? I don't think it goes through any redistrbutions

ljgray (approved):

draco/analysis/sidereal.py
jrs65: I agree. On reflection I don't think this needs to be an MPIArray. That said, I'm going to leave it as one because it currently works and I don't want to break it when I change it.

PR #248, opened 2023-07-24
fix: update cython code to account for the changed division operator

jrs65:


ljgray (approved):
Looks good as long as the tests pass

PR #249, opened 2023-08-04
feat(DebugInfo): a new task for printing useful debugging information

jrs65:


jrs65:
I quickly wrote this to have a compact way to have people give more useful and comprehensive debugging info for the pipeline. Is there anything else you think we should add?

rikvl:
Perhaps some very general info like:
- hostname
- whoami
- list of modules

ljgray:
Some basic info about the slurm job (number of nodes, memory, processes) could be useful (mostly for oom issues)

ljgray (approved):

rikvl (approved):

ketiltrout (approved):

PR #250, opened 2023-08-22
Timestream high pass filter

jrs65:
- feat(containers): add DataWeightContainer to abstract over generic inputs
- feat(containers): add VisBase as a very generic container for visibilities
- feat(HPFTimeStream): high pass filter a generic timestream


rikvl (commented):
I found two small potential typos.

As far as I can tell, the rest is fine

PR #251, opened 2023-08-24
fix(ringmapmaker): fix bug finding shortest baseline in find_basis

ssiegelx:
Previous version gives the correct result for CHIME, but can result in incorrect values for the grid spacing and indices for other telescopes due to a bug in how the shortest baseline is identified.


jrs65 (approved):

jrs65:
Yeah, that's a bug. Good catch!

PR #237, opened 2023-04-03
Wavelet Power Spectra

jrs65:
- refactor(DelaySpectrumEstimatorBase): remove unnecessary setup method
- refactor(delay): routine to flatten selected axes of a distributed dataset
- feat(wavelet): wavelet power spectrum estimation

PR #254, opened 2023-12-19
Delay cross 1

jrs65:
This contains all the parts of the cross delay PS branch that will cleanly merge.


ljgray (approved):
Think this needs a rebase after merging #237

PR #255, opened 2024-01-02
Delay cross power spectrum estimator

jrs65:
Here is a finalised version of the cross spectrum estimator that fits in with Simon's earlier refactoring. To make it work I've had to refactor the base code a bit further to remove some duplication. Notable changes:
- The output container creation has been moved from `_process_data` into a new `_create_output` method.
- The delay calculation code is moved into its own method `_calculate_delays`
- The data/weight manipulation code is moved into `_cut_data` method.

With these changes above it's reasonably simple to create a new cross spectrum estimator task.

Also in this PR I've merged in some changes I made to work with the absorber project of a mixin to abstract over containers with a "data" dataset and a corresponding weight dataset. That allows the general container transforms to chose the right datasets to use without explicitly specifying them.

### DIFF
--- a/draco/analysis/delay.py
+++ b/draco/analysis/delay.py
@@ -697,13 +697,15 @@ class DelayGeneralContainerBase(DelayTransformBase):
 
     Attributes
     ----------
-    dataset : str
-        Calculate the delay spectrum of this dataset (e.g., "vis", "map", "beam").
+    dataset : str, optional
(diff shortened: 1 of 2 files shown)


jrs65:
One more thing, for reviewing it may be easier to look through commit by commit...

ljgray (approved):
Looks good to me

PR #257, opened 2024-01-11
Ensure truncation is applied to all datasets

jrs65:
- feat(containers): routine for combining all dataset specs on a class
- fix(Truncate): ensure truncation is applied to all datasets in a container

### DIFF
--- a/draco/core/io.py
+++ b/draco/core/io.py
@@ -757,13 +757,12 @@ class Truncate(task.SingleTask):
         """
         # Check if dataset should get truncated at all
         if (self.dataset is None) or (dset not in self.dataset):
-            if dset not in container._dataset_spec or not container._dataset_spec[
-                dset
-            ].get("truncate", False):
+            cdspec = container._class_dataset_spec()
+            if dset not in cdspec or not cdspec[dset].get("truncate", False):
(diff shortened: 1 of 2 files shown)


draco/core/io.py
```diff
@@ -757,13 +757,12 @@ def _get_params(self, container, dset):
         """
         # Check if dataset should get truncated at all
         if (self.dataset is None) or (dset not in self.dataset):
-            if dset not in container._dataset_spec or not container._dataset_spec[
-                dset
-            ].get("truncate", False):
+            cdspec = container._class_dataset_spec()
+            if dset not in cdspec or not cdspec[dset].get("truncate", False):
```
ljgray, line 761: I think I'm just missing something subtle here. Wouldn't we want to use the instance `_dataset_spec` rather than the class `_dataset_spec` when deciding whether or not to truncate, since that could have been changed by the instance?

ljgray (commented):
This looks fine ~~, and seems good for readability. I think I'm just missing something about what behaviour it's fixing.~~

draco/core/io.py
ljgray: Ahhh. Of course right after writing this I've noticed the difference with `dataset_spec` and `_dataset_spec`

ljgray (approved):

PR #258, opened 2024-01-17
Enforce linting rule RUF012 

ljgray:
See [PEP 526](https://peps.python.org/pep-0526/). Mutable class attributes should be annotated with `typing.ClassVar`. I think that this makes sense in the cases I've implemented it, but I'm open to arguments.

Also, remove Sphinx < 5 version pin and set sphinx language to English


ketiltrout (approved):
This looks fine.

PR #262, opened 2024-02-14
Maximum likelihood delay power spectrum estimator

jrs65:
- feat(delay): maximum likelihood power spectrum estimator
- feat(delay): add option to use maximum likelihood estimator in tasks

### DIFF
--- a/draco/analysis/delay.py
+++ b/draco/analysis/delay.py
@@ -12,6 +12,7 @@ from draco.core.containers import ContainerBase, FreqContainer
 
 from ..core import containers, io, task
 from ..util import random, tools
+from .delayopt import delay_power_spectrum_maxpost
 
 
 class DelayFilter(task.SingleTask):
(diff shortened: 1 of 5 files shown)


draco/analysis/delay.py
```diff
@@ -734,30 +776,59 @@ def _evaluate(self, data_view, weight_view, out_cont, delays, channel_ind):
...
-                initial_S[lbi],
-                window=self.window if self.apply_window else None,
-                fsel=non_zero_channel,
-                niter=self.nsamp,
-                rng=rng,
-                complex_timedomain=self.complex_timedomain,
-            )
+                # If max-likelihood didn't converge in allowed number of iters, reflect this in the mask.
+                if not success:
+                    out_cont.datasets["spectrum_mask"][bi] = 1
```
ljgray, line 798: The indexing here is very subtle and probably worth making a comment about (I spent awhile reading it until I actually understood what is happening). When using `mpiarray.enumerate`, we get both the local array index `lbi` and the global array index `bi`. We use the local index `lbi` to get the `data`, `weight`, and `initial_S` datasets per iteration, which is expected. However, when setting the datasets in `out_cont`, we're using the global slice. This only works because we're actually indexing the `MemDataset` in `out_cont`, *not* the `MPIArray`, and `__getitem__` and `__setitem__` on `MemDatasetDistributed` operate on a global slice of the underlying array rather than the local data. So, for example, what is being done here is equivalent to doing `out_cont.datasets["spectrum_mask"][:][lbi]`, because indexing directly on the `MPIArray` is done to the local array. 

There isn't anything wrong with doing it this way, and it's consistent with what was done before. I just think there needs to be a comment here because it's quite confusing. I'm happy to add that comment later once all changes are pushed.

draco/analysis/delayopt.py
```diff
@@ -0,0 +1,593 @@
...
+    # likely from a numerical instability somewhere in the gradient/hessian computation. This has
+    # only been observed to affect a small number of (almost entirely masked) baselines that have
+    # very low weights.
+    # ValueError also occasionally gets thrown from (almost certainly) the same numerical
+    # instability: one of the minimization parameters blows up and causes an overflow when taking
+    # np.exp() of the minimization parameter to get the delay spectrum. The ValueError gets thrown
+    # again in the Cholesky decomposition when numpy runs a check_finite on the covariance (which
+    # at that point contains infs/nans).
+    except (la.LinAlgError, ValueError):
+        pass  # success already set to false if this executes.
```
ljgray, line 583: While this is true and works as expected, I think it makes the code a bit clearer to just explicitly set `success = False` here rather than just having the comment

ljgray (approved):

ljgray:
@jmaceachern I just restructured the commit history here to compartmentalize the various changes, and rebased to master

draco/analysis/delay.py
jmaceachern: Fixed in latest commit!

jmaceachern (approved):

PR #263, opened 2024-03-13
fix(GaussianNoise): fix bug in noise generation

sjforeman:
Commit https://github.com/radiocosmology/draco/commit/32da5d4553d1fb32d3268a050b231db3e8cb88c3 broke the call to `util.random.complex_normal` in `synthesis.noise.GaussianNoise`, and this fixes it. (Bug discovered by @PanChristeas)


jrs65 (approved):
Please fix the ruff errors and then this is good to merge.

PR #265, opened 2024-04-12
Add a new RFI Flagging task

ljgray:
Adds a new RFI mask and a couple of extra tools

# RFI Mask
2-stage RFI mask
- high-pass m filter to isolate transient rfi
- low-pass m filter and sum threshold over frequencies to isolate narrowband rfi

# Additional Tools
- Update `null_delay_filter` to optionally low or high pass
- Add a task to taper a 2D mask
- Add a task to calculate a baseline estimate for a 1D spectrum in the presence of outliers. Works quite well - ~too computationally expensive to extend to 2D so I didn't end up using it in the RFI filter~ This might not actually be true - going to investigate

### DIFF
--- a/draco/analysis/flagging.py
+++ b/draco/analysis/flagging.py
@@ -13,7 +13,10 @@ from typing import Union, overload
 import numpy as np
 import scipy.signal
 from caput import config, mpiarray, weighted_median
+from cora.util import units
+from skimage.filters import apply_hysteresis_threshold
 
+from ..analysis import delay
 from ..core import containers, io, task
 from ..util import rfi, tools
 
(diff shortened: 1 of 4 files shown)


draco/util/tools.py
```diff
@@ -534,3 +539,152 @@ def window_generalised(x, window="nuttall"):
...
+
+
+def arPLS_1d(y, mask=None, scale=1, end_frac=1e-2, max_iter=1000):
+    """Use arPLS to estimate a signal baseline.
+
+    Asymmetrically reweighted penalized least squares solve for a signal
+    baseline in the presence of high power outliers by heavily weighting
+    values below the signal and minimizing the weights of values above
+    the signal.
+
```
sjforeman, line 551: It would be useful to have a brief explanation of what the algorithm is actually doing. This seems to be a good reference: https://pybaselines.readthedocs.io/en/latest/algorithms/whittaker.html

draco/util/tools.py
```diff
@@ -534,3 +539,152 @@ def window_generalised(x, window="nuttall"):
...
+    tapered_mask[:, width:-width] = mask.astype(np.float64)
+    # Extend the edges
+    tapered_mask[:, :width] = tapered_mask[:, width][:, np.newaxis]
+    tapered_mask[:, -width:] = tapered_mask[:, -width - 1][:, np.newaxis]
+
+    if outer:
+        tapered_mask = 1.0 - tapered_mask
+
+    # First convolution
+    tapered_mask = np.isclose(
```
sjforeman, line 682: The purpose of the `isclose` line isn't obvious to me on first reading. A comment might be helpful.

draco/analysis/flagging.py
```diff
@@ -997,6 +1000,280 @@ def process(
...
+        dec = np.deg2rad(self.telescope.latitude)
+        lambda_inv = freq[:, np.newaxis] * 1e6 / units.c
+
+        # Maximum cut per frequency
+        return lambda_inv * baselines[:, 0].max() / np.cos(dec)
+
+    def _lpf_cut_hook(self, freq, baselines):
+        """Get a low-pass fringe rate cut for each frequency."""
+        # Corresponds to about 30 minutes
+        cut = 1 / np.deg2rad(7.5)
```
sjforeman, line 1243: The 7.5 feels like it should not be hard-coded...

draco/analysis/flagging.py
```diff
@@ -997,6 +1000,284 @@ def process(
...
+        power = np.zeros_like(weight[0], dtype=np.float64)
+
+        # Iterate over frequencies
+        for fsel in range(vis.shape[1]):
+            if np.all(mask[fsel]):
+                # Frequency is already masked
+                continue
+
+            # Apply a high-pass mmode filter. Scattered emission appears
+            # similar to an impulse function in time, so it's fourier transform
```
sjforeman, line 1139:
```suggestion
            # similar to an impulse function in time, so its fourier transform
```

draco/analysis/flagging.py
```diff
@@ -997,6 +1000,284 @@ def process(
         return mask_cont
 
 
+class RFIMModeMask(task.SingleTask):
+    """Apply high-pass and low-pass mmode filters to flag rfi.
+
```
sjforeman, line 1005: A brief description of the algorithm might be useful here.

sjforeman (approved):
Mostly looking good - just a few places where more explanation would be useful

draco/util/tools.py
ljgray: Oh cool, I never looked to see if there was an existing python implementation of this

draco/util/tools.py
sjforeman: I'm still confused about this. If the input mask is just ones and zeros, applying the taper will result in array with numbers <1, which `np.allclose` will uniformly set to zero. What am I missing here?

ljgray: Ok I'm looking at it again and I've confused myself. I think I need to poke around in a notebook and figure out what I was doing

ljgray: I've gone back through to better understand why this is done and clarified it for myself.

The idea is that the taper is always applied to the inside of the `True` regions during the convolution - i.e., any value that is 0 before convolving with the kernel will remain 0 after, and then the `1 - x` is applied afterwards if the taper should be outside. The first convolution effectively narrows the region where `x == 1` by `nwidth`, but leaves a taper which extends by `nwidth` outside the desired boundary. So, by doing `np.isclose(..., 1)` and then applying a second convolution, the taper is applied at the correct width. There are a couple of plots below to show what's happened
taper_step_1
taper_step_2

sjforeman: Got it, thanks for the extra explanation!

draco/util/tools.py
```diff
@@ -534,3 +539,166 @@ def window_generalised(x, window="nuttall"):
...
+    tapered_mask[:, width:-width] = mask.astype(np.float64)
+    # Extend the edges
+    tapered_mask[:, :width] = tapered_mask[:, width][:, np.newaxis]
+    tapered_mask[:, -width:] = tapered_mask[:, -width - 1][:, np.newaxis]
+
+    if outer:
+        tapered_mask = 1.0 - tapered_mask
+
+    # Tapering is always done on the interior of `tapered_mask` - i.e.
+    # any valud in `tapered_mask` which is zero here will remain zero,
```
sjforeman, line 694:
```suggestion
    # any value in `tapered_mask` which is zero here will remain zero,
```

sjforeman (approved):

PR #266, opened 2024-05-02
Rebinner

ljgray:
- Add a SiderealRebinner task
- Add a task to apply a gradient bin centring correction to rebinned data
- Update stacking tasks to track effective RA bin centres

### DIFF
--- a/draco/analysis/sidereal.py
+++ b/draco/analysis/sidereal.py
@@ -9,8 +9,6 @@ into  :class:`SiderealGrouper`, then feeding that into
 :class:`SiderealStacker` if you want to combine the different days.
 """
 
-from typing import Union
-
 import numpy as np
 import scipy.linalg as la
 from caput import config, mpiarray, tod
@@ -748,7 +746,7 @@ class RebinGradientFix(task.SingleTask):
     create the gradient.
     """
 
-    def setup(self, sstream_ref: Union[containers.SiderealStream, None] = None) -> None:
+    def setup(self, sstream_ref: containers.SiderealStream) -> None:
         """Provide the dataset to use in the gradient calculation.
 
         This dataset must have complete sidereal coverage.
@@ -773,23 +771,12 @@ class RebinGradientFix(task.SingleTask):
         Returns
         -------
         sstream
-            Input sidereal day with gradient correction applied
+            Input sidereal day with gradient correction applied, if needed
         """
-        if self.sstream_ref is None:
-            raise NotImplementedError("At the moment a ref stream must be given.")
-
-        if self.sstream_ref is not None:
-            self.sstream_ref.redistribute("freq")
-        sstream.redistribute("freq")
-
-        sv = sstream.vis[:].local_array
-        ssv = self.sstream_ref.vis[:].local_array
-        ssw = self.sstream_ref.weight[:].local_array
-
-        sra = self.sstream_ref.ra
-
+        # Allows a normal sidereal stream to pass through this task.
+        # Helpful for creating generic configs
         try:
-            ssra = sstream.datasets["effective_ra"][:].local_array
+            sra = sstream.datasets["effective_ra"][:].local_array
         except KeyError:
             self.log.info(
                 f"Dataset of type ({type(sstream)}) does not have an effective "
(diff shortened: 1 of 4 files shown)


draco/analysis/sidereal.py
```diff
@@ -477,6 +479,140 @@ def _regrid(self, vis, weight, lsd):
...
+            w = average_weight[fi, np.newaxis]
+            n = norm[fi, np.newaxis]
+            ssv[fi] = n * ((vis_data[fi] * w) @ Rt)
+
+            regrid_var = tools.invert_no_zero(weight[fi]) * (w @ Rt) ** 2
+            ssw[fi] = tools.invert_no_zero(n**2 * regrid_var)
+
+        return sdata
+
+    def _rebin_matrix(
```
ssiegelx, line 557: The `rebin_matrix` method could be moved into `util.regrid` since it does not require access to any class attributes.

draco/core/containers.py
```diff
@@ -1252,6 +1252,34 @@ def _mean(self):
         return self.datasets["vis"]
 
 
+class SiderealStreamRebin(SiderealStream):
```
ssiegelx, line 1255: This class does not define `vis`, `weight`, `input_flag`, etc., datasets in the `_dataset_spec` dictionary.  I am finding that if I attempt to load a `SiderealStreamRebin` from disk with selections along an axis, then the selection will only be applied to the datasets that are currently defined in `_dataset_spec`.  I think we need to either:
* include all possible datasets in `_dataset_spec`
* have `_dataset_spec` populated from subclasses, similar to what is done with the `_axes` attribute
* update the code that does the axis selection to get the lists of datasets from some other source (and update anywhere else in the code that uses this attribute)

draco/analysis/sidereal.py
```diff
@@ -477,6 +479,140 @@ def _regrid(self, vis, weight, lsd):
...
+        # Get the average weights over baselines
+        average_weight = weight.mean(axis=1)
+        # Get the median time sample width
+        width_t = np.median(np.abs(np.diff(timestamp_lsd)))
+        # Create the regular grid of RA samples
+        target_lsd = np.linspace(self.start, self.end, self.samples, endpoint=False)
+
+        R = self._rebin_matrix(timestamp_lsd, target_lsd, width_t=width_t)
+        Rt = ss.csr_array(R.T)
+
```
ssiegelx, line 531:
```suggestion
Rtsq = Rt.power(2)

```

draco/analysis/sidereal.py
```diff
@@ -477,6 +479,140 @@ def _regrid(self, vis, weight, lsd):
...
+        ssv = sdata.vis[:].local_array
+        ssw = sdata.weight[:].local_array
+
+        for fi in range(vis_data.shape[0]):
+            w = average_weight[fi, np.newaxis]
+            n = norm[fi, np.newaxis]
+            ssv[fi] = n * ((vis_data[fi] * w) @ Rt)
+
+            regrid_var = tools.invert_no_zero(weight[fi]) * (w @ Rt) ** 2
```
ssiegelx, line 552:
```suggestion
            regrid_var = (tools.invert_no_zero(weight[fi]) * w**2) @ Rtsq
```
In order to propagate the uncertainty from the raw time samples to the regridded time samples, we need to sum the variances of the raw samples multiplied by the square of any coefficient/weight that was applied when performing the rebinning.  I think what is being done here will calculate the square of the sum of the weights/coefficients, whereas we need the sum of the squared weights/coefficients.  Also the raw variances taken from `invert_no_zero(weight)` needs to be inside the matrix multiply to get the proper dimensions.

ssiegelx (changes requested):

draco/core/containers.py
ljgray: Looks like this is fixed in #260, I've updated that PR

draco/analysis/sidereal.py
```diff
@@ -477,6 +479,91 @@ def _regrid(self, vis, weight, lsd):
...
+        # Get the average weights over baselines
+        average_weight = weight.mean(axis=1)
+        # Get the median time sample width
+        width_t = np.median(np.abs(np.diff(timestamp_lsd)))
+        # Create the regular grid of RA samples
+        target_lsd = np.linspace(self.start, self.end, self.samples, endpoint=False)
+
+        R = regrid.rebin_matrix(timestamp_lsd, target_lsd, width_t=width_t)
+        Rt = ss.csr_array(R.T)
+        Rtsq = Rt**2
```
ssiegelx, line 539: Does this operation work for you?  When I try `Rt**2` with a `csr_array` I get an error that the matrix is not square.  I think it's attempting to do a matrix multiply of Rt with itself.  I need to do `Rt.power(2)` to get the element-wise operation.

ljgray: Ah - I didn't test it, I just assumed it would work since that's the standard behaviour with numpy arrays. I suppose that isn't consistent with sparse arrays. I'll fix it

ljgray:
Closed in favour of #269

ljgray:
Closed in favour of #269

PR #268, opened 2024-05-14
feat(flagging): mask rfi based on delay filtered visibilities

ssiegelx:
Adds three new tasks:
 - analysis.dayenu.DayenuDelayFilterFixedCutoff enables delay filtering of timeseries with rapidly changing frequency masks.
 - analysis.transform.ReduceChisq calculates a chi2-per-dof test statistic by comparing the delay filtered visibilities to the inverse noise variance stored in the weights.
 - analysis.flagging.RFIMaskChisqHighDelay generates a mask from this chi2-per-dof test statistic.

### DIFF
--- a/draco/analysis/dayenu.py
+++ b/draco/analysis/dayenu.py
@@ -189,6 +189,144 @@ class DayenuDelayFilter(task.SingleTask):
         return baseline_delay_cut + self.tauw
 
 
+class DayenuDelayFilterFixedCutoff(task.SingleTask):
+    """Apply a DAYENU high-pass delay filter to visibility data.
+
+    This task loops over time instead of baseline.  It can be used
+    to filter timeseries that have a rapidly changing frequency mask,
+    with the caveat that one must use the same delay cutoff for all baselines.
+
+    Attributes
+    ----------
+    epsilon : float
+        The stop-band rejection of the filter.  Default is 1e-12.
+    tauw : float
+        Delay cutoff in micro-seconds.  Default is 0.4 micro-seconds.
+    single_mask : bool
+        Apply a single frequency mask for all baselines.  Only includes
+        frequencies where the weights are nonzero for all baselines.
+        Otherwise will construct a filter for all unique single-time
+        frequency masks (can be significantly slower).  Default is True.
+    atten_threshold : float
+        Mask any frequency where the diagonal element of the filter
+        is less than this fraction of the median value over all
+        unmasked frequencies.  Default is 0.0 (i.e., do not mask
+        frequencies with low attenuation).
+    """
+
+    epsilon = config.Property(proptype=float, default=1e-12)
+    tauw = config.Property(proptype=float, default=0.400)
+    single_mask = config.Property(proptype=bool, default=True)
+    atten_threshold = config.Property(proptype=float, default=0.0)
+
+    def process(self, stream):
+        """Filter delays below some cutoff.
+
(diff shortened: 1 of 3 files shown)


ssiegelx:
Depends on #265 for the arPLS method.

draco/analysis/dayenu.py
```diff
@@ -189,6 +188,147 @@ def _get_cut(self, prod):
...
+        weight = stream.weight[:].local_array
+
+        # Identify baselines that are always flagged
+        temp = np.any(weight > 0.0, axis=(0, 2))
+        baseline_flag = np.zeros_like(temp)
+        self.comm.Allreduce(temp, baseline_flag, op=MPI.LOR)
+        baseline_mask = ~baseline_flag
+
+        if np.all(baseline_mask):
+            self.log.error("All data flaged as bad.")
```
ljgray, line 255:
```suggestion
            self.log.error("All data flagged as bad.")
```

draco/analysis/dayenu.py
```diff
@@ -189,6 +188,147 @@ def _get_cut(self, prod):
...
+            self.log.debug(f"Filter time {tt} of {ntime}. [{self.tauw:0.3f} microsec]")
+
+            # Construct the filter
+            try:
+                NF, index = highpass_delay_filter(
+                    freq, self.tauw, flag, epsilon=self.epsilon
+                )
+
+            except np.linalg.LinAlgError as exc:
+                self.log.error(f"Failed to converge while processing time {tt}: {exc}")
```
ljgray, line 291: I'm not too familiar with the behaviour of the dayenu filter - is it expected to throw `LinAlgError` every now and then? Or would it be better for the pipeline to fail here if there's an issue?

draco/analysis/flagging.py
```diff
@@ -997,6 +997,253 @@ def process(
...
+        # For each frequency, caculate a weighted median
+        y = np.ascontiguousarray(y.astype(np.float64))
+        w = np.ascontiguousarray((~m).astype(np.float64))
+
+        med_y = weighted_median.weighted_median(y, w)
+        med_m = np.all(m, axis=-1)
+        med_w = np.ascontiguousarray((~med_m).astype(np.float64))
+
+        # Estimate a baseline
+        baseline = tools.arPLS_1d(med_y, mask=med_m, scale=self.reg_arpls)
```
ljgray, line 1173:
```suggestion
        baseline = tools.arPLS_1d(med_y, mask=med_m, lam=self.reg_arpls)
```

This is on me, I changed this variable name in `arPLS_1d` to better reflect the standard variables used in the equation it's solving

ljgray (approved):
Looks good to me - just some minor comments

draco/analysis/dayenu.py
ljgray: Sounds good

ssiegelx:
I've confirmed that using the updated `DayenuDelayFilterFixedCutoff` task yields the same result as previously obtained using the three separate tasks.  If there are no objections I will go ahead and merge this PR and the corresponding one in ch_pipeline.

ljgray:
> I've confirmed that using the updated `DayenuDelayFilterFixedCutoff` task yields the same result as previously obtained using the three separate tasks. If there are no objections I will go ahead and merge this PR and the corresponding one in ch_pipeline.

Sounds good to me

PR #270, opened 2024-05-22
fix(sensitivity): fix bug in exclude_intracyl sensitivity calculation

ssiegelx:
Fixed bug in ComputeSystemSensitivity when exclude_intracyl was set to true.  The inter-cylinder baselines were not being identified correctly when calculating the `measured` sensitivity dataset due to the CHIME cylinder rotation.  This resulted in 30 MHz ripples as a function of frequency when calculating the normalized sensitivity (`measured / radiometer`) from which the RFI mask is derived.  These ripples would change slightly with each gain updates, causing discontinuities in the metric that we were flagging on.

### DIFF
--- a/draco/analysis/sensitivity.py
+++ b/draco/analysis/sensitivity.py
@@ -125,9 +125,14 @@ class ComputeSystemSensitivity(task.SingleTask):
 
         baseline_pol = np.core.defchararray.add(pol_a, pol_b)
 
+        ew_intra = 0.5 * self.telescope.cylinder_width
         if self.exclude_intracyl:
             baseline_flag = (
-                ew_position[prodstack["input_a"]] != ew_position[prodstack["input_b"]]
+                np.abs(
+                    ew_position[prodstack["input_a"]]
+                    - ew_position[prodstack["input_b"]]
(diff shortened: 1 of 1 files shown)


ssiegelx:
This solves a long standing question I had as to why the average value of the radiometer test (`measured / radiometer`) was 0.88 and showed significant 30 MHz structure as a function of frequency.  The intracylinder baselines were not being excluded properly in the calculation of the `measured` dataset due to the cylinder rotation appearing in the baseline distances.

[fix_intra_check.pdf](https://github.com/radiocosmology/draco/files/15397616/fix_intra_check.pdf)

Here is a comparison of the radiometer test for CSD 1892 before and after this fix.  After the fix, the average value is 1.02, indicating that the measured variance is very close to our expectation for radiometer noise. The residual 2% bias might be due to quantization effects.  Also there is no longer ripple as a function of frequency.  This ripple would change slightly every gain update, meaning that it would not be removed entirely by the per-frequency median over time subtraction and would cause difficulties for the RFI masking.

ljgray (approved):
That looks like a huge improvement

PR #272, opened 2024-05-28
fix(rfi): fix misalignment between mask and data in sumthreshold

ssiegelx:
Also removes unnecessary slices.


ljgray (approved):

ssiegelx:
For `m = 1` it will be properly aligned, but for other odd values of `m` there will be a shift that is proportional to the size of the window.

PR #274, opened 2024-05-30
ss/sens mask

ssiegelx:
Makes a number of modifications to RFISensitivityMask:
  * First generate a static frequency mask by looking for outlier
    channels in the quantile over time.
  * Iteratively determine both the baseline and the standard deviation about
    that baseline by applying weighted rolling medians that accounts for the
    current masks.
  * Remove the baseline before applying the sumthreshold method.
  * Use the frequency and time dependent standard deviation in the sumthreshold method.
  * Increase the default maximum window size used in sumthreshold.
  * Modify the sumthreshold method to allow for:
     - Full control over how the threshold changes with window size.
       Use the expected standard deviation times a constant value for the
       sensitivity-based masking.
     - Calculate the expected noise in the convolution by propagating estimates
       of the variance for the individual samples.

### DIFF
--- a/draco/analysis/flagging.py
+++ b/draco/analysis/flagging.py
@@ -1566,7 +1566,7 @@ class RFISensitivityMask(task.SingleTask):
         by more than this number of median absolute deviations.
     quantile_1d: float, optional
         The quantile to use along time to construct the static mask.
-        Defaults to 0.5 (the median).
+        Defaults to 0.15.
     win_f_1d : int, optional
         Number of frequency channels used to calculate a rolling median
         and median absolute deviation for the staic mask.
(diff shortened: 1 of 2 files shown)


draco/analysis/flagging.py
```diff
@@ -1768,6 +1785,34 @@ def _static_rfi_mask_hook(self, freq, timestamp=None):
...
+                absd_medt_y, medt_w
+            )
+        else:
+            mad_1d = self.MAD_TO_RMS * weighted_median.moving_weighted_median(
+                absd_medt_y, medt_w, self.win_f_1d
+            )
+
+        return absd_medt_y > (self.nsigma_1d * mad_1d), medt_y
+
     def _apply_sir(self, mask, baseflag, eta=0.2):
```
ljgray, line 1816: I'm curious about using the scale invariant rank operator over the time axis only. I think this would effectively extend frequency bands that are masked by the sum threshold and not by the MAD, while not increasing the amount of masked data too significantly. This is what's done in the second stage of the visibility-based masking, and using the operator in time only seemed to only increase the amount of masked data by around 1%.

ljgray (approved):
Looks great. Just one comment about the SIR operation

draco/analysis/flagging.py
ssiegelx: Thanks!  I gave this a try and it properly masked most of the residual RFI during transits that was not being picked up by the MAD algorithm.  However, it increased the total amount of masked data by 3%, most of which was obviously good data.

I ended up applying the SIR operator but then only using the extended mask during the bright source transits.  This mostly does what I want and increases the amount of data masked by less than 1%.

PR #280, opened 2024-06-20
Modifications to BlendStack and frequency mask

ljgray:


draco/analysis/flagging.py
```diff
@@ -2431,49 +2443,47 @@ def process(self, data):
...
-                .astype(np.float32)
-                .view(np.ndarray)
-            )
+            mask = ((weight[:] > 0) & (weight_stack[:] > 0)).astype(np.float32)
 
-            # ... get the median of the stack in this common subset
+            # Get the median of the stack in this common subset
             stack_med_real = weighted_median.weighted_median(
-                dset_stack.real.view(np.ndarray).copy(), mask
+                np.ascontiguousarray(dset_stack.real), mask
```
jmaceachern, line 2457: Obviously not critical, but just curious: what is the reason for the call to `ascontiguousarray`? Is this related to the change to using `local_array`'s above? Is the stack somehow not in C ordering in this case?

jmaceachern (approved):

draco/analysis/flagging.py
ljgray: It's because we look at the real and imaginary components individually, and those aren't contiguous on their own (you can imagine that storing the full complex dataset in a contiguous format would mean that taking only a component of that would no longer be contiguous). This was being implicitly done by the `.copy()` call, so I just changed that to make the behaviour clearer (without actually changing what's being done)

PR #281, opened 2024-06-24
Increase minimum supported python to 3.9

ljgray:
Python 3.8 reaches end-of-life in October 2024


ketiltrout (approved):
I've updated the branch protection rules for this workflow change:
* removed `run-tests (3.8)`
* added `run-tests (3.9)`

PR #282, opened 2024-07-17
Fix visibility gradient correction for quarterstacks

ljgray:
Multiple fixes for tracking effective ra and applying a gradient correction:
- Initialize the `effective_ra` dataset in the stack container
- Properly replace the effective ra of masked samples with the bin centre ra
- Handle NaN values in the visibility gradient where the ra step is 0

### DIFF
--- a/draco/analysis/sidereal.py
+++ b/draco/analysis/sidereal.py
@@ -641,6 +641,12 @@ class RebinGradientCorrection(task.SingleTask):
             if not np.any(weight[fi]):
                 continue
 
+            # We need to track the full time mask applied to each
+            # baseline. Because the `effective_ra` dataset is baseline
+            # dependent, the gradient could end up producing a slightly
+            # baseline dependent mask
+            fmask = np.zeros(weight.shape[-1], dtype=bool)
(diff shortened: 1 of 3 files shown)


ssiegelx (approved):
Looks good!

Arnab-half-blood-prince:
I tried to apply this task to make a quartarstack, where I used `/project/rpp-chime/chime/chime_processed/stacks/rev_03/all/sstack.h5` for the gradient correction. It works. But, if I use the same sstream as a requirement for the gradient correction, it is not working. It ignores the task and does not produce the rest of the products. I am not sure why this is so.

ljgray:
> I tried to apply this task to make a quartarstack, where I used `/project/rpp-chime/chime/chime_processed/stacks/rev_03/all/sstack.h5` for the gradient correction. It works. But, if I use the same sstream as a requirement for the gradient correction, it is not working. It ignores the task and does not produce the rest of the products. I am not sure why this is so. 

This was actually a bug with the caput pipeline runner. If you pull the newest caput it should be fixed.

PR #283, opened 2024-08-01
Convert build to pyproject.toml

ljgray:
- Converts most of the project setup to use `pyproject.toml`. 
- Removes versioneer and switches to `setuptools-git-versioning`
- Cleans up CI pipeline to use optional dependencies from project file


ljgray:
I believe this is failing because of a versioneer issue with cora. I might have to update cora and driftscan before draco

ketiltrout (approved):
Great!

PR #284, opened 2024-08-02
Avoid gradient correction for masked values

ljgray:


draco/analysis/sidereal.py
```diff
@@ -682,7 +682,7 @@ def process(self, sstream: containers.SiderealStream) -> containers.SiderealStre
         # Helpful for creating generic configs
         try:
             era = sstream.effective_ra[:].local_array
-        except KeyError:
+        except AttributeError:
```
ssiegelx, line 685: I think the `effective_ra` property will indeed return a `KeyError` if the dataset is missing:
```
    @property
    def effective_ra(self):
        """Get the effective_ra dataset if it exists, None otherwise."""
        if "effective_ra" in self.datasets:
            return self.datasets["effective_ra"]

        raise KeyError("Dataset 'effective_ra' not initialised.")
```

ssiegelx (approved):
Just a few questions

draco/analysis/sidereal.py
ljgray: Good catch. I'm not sure why I made that change.

draco/analysis/sidereal.py
```diff
@@ -710,12 +710,6 @@ def process(self, sstream: containers.SiderealStream) -> containers.SiderealStre
             if not np.any(weight[fi]):
                 continue
 
-            # We need to track the full time mask applied to each
```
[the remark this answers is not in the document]
ljgray: It was a while ago, but I think I remember having issues where a baseline that was flagged for most but not all times would end up masking the entire dataset, or something of that sort. I think that the only way that the mask could change for a single baseline is if two adjacent `effective_ra` samples are identical, which shouldn't be able to happen for any baselines.

PR #291, opened 2024-09-13
ss/daily filter

ssiegelx:
This contains all of the modifications to draco needed to implement the daily foreground filtering pipeline, which beamforms a Timestream along north-south baselines and then applies a high-pass filter along the frequency axis.  It primarily updates various tasks to work with the HybridVisStream.


ssiegelx:
Apologies for putting this all in a single PR.  We probably want to hold off on merging this until we have tested on the 100 days and confirmed that it works.

ljgray (approved):
I've gone through the changes here and I think it all looks good, other than a couple of very minor comments. It would be nice to get this branch merged in since everything here seems to be more or less complete and tested, and any extra changes we end up needing could be a separate PR to make it easier to review.

The commits should probably be cleaned up with an interactive rebase before merging.

ljgray:
Also, there are a few places where the `ruff` linting is failing

draco/analysis/ringmapmaker.py
```diff
@@ -111,11 +111,12 @@ def process(self, sstream):
...
         # Extract the right ascension to initialise the new container with (or
         # calculate from timestamp)
-        ra = (
-            sstream.ra
-            if "ra" in sstream.index_map
-            else self.telescope.lsa(sstream.time)
-        )
+        if "ra" in sstream.index_map:
+            ra = sstream.ra
+        elif "lsd" in sstream.attrs:
```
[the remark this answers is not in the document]
ssiegelx: The SiderealGrouper task keeps a small amount of data that extends beyond the boundaries of the sidereal day. If you convert to RA here, then the RA axis will wrap, which causes issues later on when passed to the SiderealRebinner task.  This line sets the timestamps before the sidereal day to RAs less than 0 deg and timestamps after the sidereal day to RAs greater than 360 degrees.  This behaves correctly when passed to SiderealRebinner.

draco/analysis/dayenu.py
```diff
@@ -403,6 +403,137 @@ def process(self, stream):
...
 
 
+class DayenuDelayFilterHybridVis(task.SingleTask):
+    """Apply a DAYENU high-pass delay filter to hybrid beamformed visibilities.
+
+    Attributes
+    ----------
+    tau_width : float or np.ndarray[nstopband,]
+        The half width of the stop-band region in micro-seconds.
+    tau_centre : float or np.ndarray[nstopband,]
```
Arnab-half-blood-prince, line 413: tau_centre should be tauc

draco/analysis/dayenu.py
```diff
@@ -403,6 +403,137 @@ def process(self, stream):
         return out
 
 
+class DayenuDelayFilterHybridVis(task.SingleTask):
+    """Apply a DAYENU high-pass delay filter to hybrid beamformed visibilities.
+
+    Attributes
+    ----------
+    tau_width : float or np.ndarray[nstopband,]
```
Arnab-half-blood-prince, line 411: tau_width should be tauw

draco/analysis/flagging.py
```diff
@@ -1592,6 +1637,71 @@ def mask_2d(self, y, w):
...
+
+            # Generate a mask using sumthreshold
+            stmask = rfi.sumthreshold(
+                dy,
+                self.max_m,
+                start_flag=mask,
+                threshold1=nsigma,
+                remove_median=False,
+                correct_for_missing=True,
+                rho=1.0,
```
Arnab-half-blood-prince, line 1694: Is rho would be `rho = self.rho` ?

Arnab-half-blood-prince (commented):
looks good to me. I just have only a few comments. Most of these new tasks, I checked and it is working.

draco/analysis/flagging.py
ssiegelx: Rho determines how the threshold is lowered with each iteration and is used to determine self.threshold in the setup method.  For chi-squared masking and sensitivity masking, I've moved this iteration outside of the sumthreshold function so that we can also update the median and median absolute deviation as we lower the threshold.

Arnab-half-blood-prince (approved):

PR #296, opened 2024-10-16
Only initialize dirty_beam dataset if it will be used

ljgray:
Otherwise, the dataset will remain empty which can introduce NaN values and takes up memory when saved.

This appears to be the source of NaN values in some of the beamformed datasets in the daily filtered pipeline


ssiegelx:
Ah yeah, good catch!

ssiegelx (approved):

PR #297, opened 2024-10-22
Allow selection from index map values

ljgray:
Adds the ability for `SelectionsMixin` to make selections based on index_map entries. This has to be explicitly enabled with the `allow_index_map` attribute, and will raise an error if an item is not found in the index map.

### DIFF
--- a/draco/analysis/transform.py
+++ b/draco/analysis/transform.py
@@ -1354,22 +1354,35 @@ class Downselect(io.SelectionsMixin, task.SingleTask):
             Container of same type as the input with specific axis selections.
             Any datasets not included in the selections will not be initialized.
         """
-        # Re-format selections to only use axis name
-        for ax_sel in list(self._sel):
-            ax = ax_sel.replace("_sel", "")
-            self._sel[ax] = self._sel.pop(ax_sel)
+        sel = {}
+
+        # Parse axes with selections and reformat to use only
+        # the axis name
+        for k in self.selections:
+            *axis, type_ = k.split("_")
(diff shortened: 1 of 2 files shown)


draco/analysis/transform.py
```diff
@@ -1354,10 +1354,19 @@ def process(self, data: containers.ContainerBase) -> containers.ContainerBase:
...
-        for ax_sel in list(self._sel):
-            ax = ax_sel.replace("_sel", "")
-            self._sel[ax] = self._sel.pop(ax_sel)
+        # Parse axes with selections and reformat to use only
+        # the axis name
+        for k in self.selections:
+            *axis, type_ = k.split("_")
+            axis_name = "_".join(axis)
+
+            sel = self._sel.pop(f"{axis_name}_sel")
```
ssiegelx, line 1363: One issue that I encountered is if I run `Downselect` more than one time in a single job (when I am iterating over files) then this line will crash because `{axis_name}_sel` no longer exists and has been replaced with `{axis_name}`.

To fix this I changed `pop` to `get` and then populated a new dictionary that is instantiated at the start of the `process` method.  This new dictionary is what is used to specify selections in the lines below this for loop.

ssiegelx (changes requested):

draco/analysis/transform.py
ljgray: Good catch, I'll fix that here

ssiegelx (approved):
Look good, thanks!

PR #300, opened 2024-12-08
ac/apply_filter

Arnab-half-blood-prince:
Apply previously saved filter to HybridVis data

### DIFF
--- a/draco/analysis/dayenu.py
+++ b/draco/analysis/dayenu.py
@@ -579,6 +579,22 @@ class ApplyDelayFilterHybridVis(task.SingleTask):
         hv.redistribute(["ra", "time"])
         self.sstream.redistribute(["ra", "time"])
 
+        # Validate that both hybrid beamformed visibilites match
+        if not np.array_equal(self.stream.freq, hv.freq):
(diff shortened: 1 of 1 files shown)


draco/analysis/dayenu.py
```diff
@@ -534,6 +534,93 @@ def process(self, stream):
...
+        -------
+        hv_filt: containers.HybridVisStream
+          The filtered dataset.
+        """
+        # Distribute over products
+        hv.redistribute(["ra", "time"])
+        self.sstream.redistribute(["ra", "time"])
+
+        npol, nfreq, new, nel, ntime = hv.vis.local_shape
+
```
ssiegelx, line 583: It might be a good idea to check that the various axes (freq, ew, pol, time) match between the filter and the hybrid-beamformed visibilities using `np.array_equal` similar to what is done at the start of `DeconvolveHybridMBase`.  This would guard against a situation where different subsets of the data have been loaded for the two containers.

ssiegelx (approved):
I think this looks good.  I just have one suggestion.

draco/analysis/dayenu.py
Arnab-half-blood-prince: done

PR #310, opened 2024-12-13
Add DelayTransform compression and increase SiderealStream chunk size

ljgray:
Still need to test to make sure truncation effects are not significant

### DIFF
--- a/draco/core/containers.py
+++ b/draco/core/containers.py
@@ -1134,7 +1134,7 @@ class SiderealStream(
             "distributed_axis": "freq",
             "compression": COMPRESSION,
             "compression_opts": COMPRESSION_OPTS,
-            "chunks": (64, 128, 128),
+            "chunks": (32, 512, 2048),
             "truncate": {
                 "weight_dataset": "vis_weight",
             },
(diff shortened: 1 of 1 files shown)


draco/core/containers.py
```diff
@@ -1182,7 +1182,7 @@ class SiderealStream(
             "distributed_axis": "freq",
             "compression": COMPRESSION,
             "compression_opts": COMPRESSION_OPTS,
-            "chunks": (64, 128, 128),
+            "chunks": (32, 512, 2048),
```
ljgray, line 1185: This is consistent with the daily filtering pipeline, we just never changed the defaults

ssiegelx (approved):
Did the truncation significantly impact the delay transform?  What does it use to determine the truncation level since there is no weight dataset?

ljgray:
> Did the truncation significantly impact the delay transform? What does it use to determine the truncation level since there is no weight dataset?

It hasn't been tested yet, so I'm not going to merge until that's done. I think it will just use a constant variance increase which would be supplied in the `Truncate` task

PR #311, opened 2024-12-16
Fix incorrect low pass null delay filter

ljgray:


draco/analysis/delay.py
```diff
@@ -2203,17 +2203,17 @@ def null_delay_filter(
     # to be the fault of MKL (see https://github.com/scipy/scipy/issues/10032 and links
     # therein). This seems to be limited to the `gesdd` LAPACK routine, so we can get
     # around it by switching to `gesvd`.
-    u, sig, vh = la.svd(F, full_matrices=False, lapack_driver=lapack_driver)
+    u, sig, _ = la.svd(F, full_matrices=False, lapack_driver=lapack_driver)
```
sjforeman, line 2206: Don't we need `full_matrices=True` here? If we're keeping all low-delay modes and not applying a mask or window, then we want `p @ p.T.conj()` to be the identity, and this only works when `full_matrices=True`. But then, if `full_matrices=True`, I believe that the old and new versions are equivalent.

sjforeman (changes requested):

draco/analysis/delay.py
ljgray: I made this PR awhile ago, so I just had another look. From what I can tell, you're correct that the old implementation requires `full_matrices=True`.

From a quick test, it looks like this new change produces the same `proj` matrix, regardless of whether `full_matrices` is True or False, as using the old implementation with `full_matrices=True`.

PR #312, opened 2024-12-16
Truncation weight broadcasting and get weights from a property

ljgray:
Pulls some useful functionality from #259.

These commits are required when to broadcast weights when compressing a `HybridVisStream`, but were previously contained in a much larger PR

### DIFF
--- a/draco/core/io.py
+++ b/draco/core/io.py
@@ -854,13 +854,12 @@ class Truncate(task.SingleTask):
                         val, specs["fixed_precision"]
                     ).reshape(old_shape)
             else:
-                invvar = (
-                    np.broadcast_to(
-                        data[specs["weight_dataset"]][:], data[dset][:].shape
-                    )
-                    .copy()
-                    .reshape(-1)
-                )
+                if hasattr(data, specs["weight_dataset"]):
+                    invvar = getattr(data, specs["weight_dataset"])
+                else:
+                    invvar = data[specs["weight_dataset"]][:]
+
+                invvar = np.broadcast_to(invvar, data[dset][:].shape).copy().reshape(-1)
                 invvar *= (2.0 if np.iscomplexobj(data[dset]) else 1.0) / specs[
                     "variance_increase"
                 ]
--- a/draco/core/io.py
+++ b/draco/core/io.py
@@ -801,6 +801,56 @@ class Truncate(task.SingleTask):
 
         return params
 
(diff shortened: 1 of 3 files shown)


draco/core/io.py
```diff
@@ -854,16 +855,27 @@ def process(self, data):
...
+                    invvar = getattr(data, specs["weight_dataset"])
+                else:
+                    wdset = data[specs["weight_dataset"]]
+                    # Add missing axes to the weights dataset if
+                    # needed and if possible
+                    waxes = wdset.attrs.get("axis", [])
+                    daxes = data[dset].attrs.get("axis", [])
+                    # Add length-one axes
+                    slobj = tuple(
+                        slice(None) if ax in waxes else np.newaxis for ax in daxes
```
ssiegelx, line 870: Does expanding the dimensions need to be moved outside of the else block?  I think even if weight is an attribute of the container, it could still have a subset of axes compared to the dataset.  For example if someone provided `weight` as the `weight_dataset` for the `HybridVisStream`.

ssiegelx (approved):
Thanks, it will be great to have this merged.  I just have one question.

draco/core/io.py
ljgray: Ah, that's a good point. I'll make that change

ljgray: Right. I actually realized why I did it this way. It uses the axes of the `weight_dataset` dataset to figure out which axes to add, rather than just the dimension of the weight array. So, if the weights are extracted from a class attribute, there won't necessarily be a dataset container with those axes specified, so it won't be able to figure out how to add the extra dimensions. There are two options:
1. Keep it the way it is, but maybe swap the order (i.e. check for a dataset _first_, then try falling back to using an attr)
2. Change the logic to add extra axes based on the dimension of the arrays rather than the specified axes. The only risk here is the chance of having two adjacent axes with the same length, which could result in broadcasting over the wrong axis.

Thoughts?

ssiegelx: I think in most cases the `weight` property returns a `memh5.MemDataset`, which should still have an `attrs["axis"]`.  We could try to use that.  If it's not available we could see if they can be broadcast against each other outright and if not raise an error.

ljgray: I've changed this around

ssiegelx (approved):

PR #313, opened 2024-12-16
Skip applying dayenu filter and add options when applying existing filter

ljgray:
This PR does two things:
1. Adds an option to generate but not apply a dayenu high-pass filter. At the moment, this is only implemented for hybrid vis datasets, but should probably be consistent everywhere
2. Extend the task which applies an existing filter so that it can be used with different loop options. Previously, only a single filter could be applied to multiple dataset, since the filter was provided in the `setup` method which can only run once. This behaviour is preserved in the new task. The alternate behaviour is to provide both the filter and the dataset to filter in the `process` method, which lets us split up the process of generating and applying a filter to each dataset


ssiegelx (approved):
Looks good

PR #315, opened 2024-12-24
fix(dayenu): do not apply atten_threshold to fully masked samples

ssiegelx:
Also ensure there are no masked frequencies in the data that were unmasked during filter generation.

### DIFF
--- a/draco/analysis/dayenu.py
+++ b/draco/analysis/dayenu.py
@@ -615,17 +615,35 @@ class ApplyDelayFilterHybridVis(task.SingleTask):
         # loop over products
         for tt in range(ntime):
             t0 = time.time()
+            self.log.debug(f"Filter time {tt} of {ntime}.")
 
             for xx in range(new):
-                self.log.debug(f"Filter time {tt} of {ntime}, baseline {xx} of {new}.")
 
                 for pp in range(npol):
+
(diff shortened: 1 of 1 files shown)


ljgray (approved):

draco/analysis/dayenu.py
```diff
@@ -615,18 +615,38 @@ def process(self, hv, source):
...
+                    # are also unmasked in the data
+                    valid_freq = np.any(np.abs(NF) > 0.0, axis=0)
+                    missing_freq = np.flatnonzero(valid_freq & ~flag)
+                    if missing_freq.size > 0:
+                        self.log.warning(
+                            "Missing the following frequencies that were "
+                            "assumed valid during filter generation: "
+                            f"{missing_freq}"
+                        )
+                        weight[pp, :, xx, tt] = 0.0
```
Arnab-half-blood-prince, line 647: don't you want, the missing freq would be 0 in the weight?

ssiegelx: If there are missing frequencies, then applying the filter won't work at all, since those frequencies were not taken into account during filter generation.  This will just set the weight to zero for all frequencies if that happens.

Arnab-half-blood-prince (approved):

PR #316, opened 2024-12-30
feat(io): class for loading files from attributes of an input container

ssiegelx:
Useful for loading auxiliary files when iterating over the primary data products.  For example, as one iterates over LSDs during quarter stacking, this can be used to load the appropriate rfi mask for each sidereal day.

### DIFF
--- a/draco/core/io.py
+++ b/draco/core/io.py
@@ -561,6 +561,44 @@ class LoadFilesFromParams(BaseLoadFiles):
 LoadBasicCont = LoadFilesFromParams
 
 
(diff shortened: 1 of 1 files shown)


ssiegelx:
I'm currently using this during quarterstacking to load the rfi mask corresponding to each hybrid beamformed visibility.  This is similar to `ch_pipeline.core.io.LoadFileFromTags` but provides more flexibility in how the path is constructed.

ljgray (approved):

ljgray:
This seems really useful

PR #318, opened 2025-01-09
DPSS inpainting and FFT delay transform

ljgray:
I want to look more into doing simultaneous frequency-ra interpolation, but this code is mostly complete and tested, so I think it's ready to be a PR.

This PR implements DPSS-based inpainting over a single axis, following the implementation in HERA's paper (https://arxiv.org/abs/2411.10529). It's been tested to work as-expected over both frequency and RA, with the caveat that it does not convert the RA axis to radians, which needs to be taken into account when selecting an appropriate cutoff. 

### util.dpss
Implements all the actual inpainting code. Only the diagonal of the final covariance is returned. 

### analysis.interpolate
Implements inpainting as a task. Includes a base task, which applies a constant cutoff over a single axis, and tasks which try to calculate a per-baseline foreground cutoff for frequency and ra. 

### analysis.delay
Implements a simple FFT delay transform using our current delay transform framework, with an optional argument to return the delay power spectrum instead. The power spectrum argument is also added to the Wiener filter delay transform code. I hope to eventually do another minor refactor of this module.

### analysis.transform
`analysis.delay.stokes_I` is moved here, with a slight modification to use the telescope polarisation property for improved readability. A standalone task to extract stokes I visibilities is also added, which was needed because, when inpainting, we want to inpaint _after_ computing stokes I.

### DIFF
--- a/draco/analysis/delay.py
+++ b/draco/analysis/delay.py
@@ -4,7 +4,7 @@ from typing import Optional, TypeVar, Union
 
 import numpy as np
 import scipy.linalg as la
-from caput import config, memh5, mpiarray
+from caput import config, fftw, memh5, mpiarray
 from cora.util import units
 from numpy.lib.recfunctions import structured_to_unstructured
 
@@ -397,6 +397,7 @@ class DelayTransformBase(task.SingleTask):
             "nuttall",
             "blackman_nuttall",
             "blackman_harris",
+            "tukey-0.2",
         ],
         default="nuttall",
     )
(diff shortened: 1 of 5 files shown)


draco/analysis/delay.py
```diff
@@ -979,29 +985,34 @@ class DelayPowerSpectrumGeneralEstimator(
...
+            "This means that _all_ frequencies will be "
+            "included in the delay transform, includeing "
+            "those which are masked."
+        )
+        self.freq_frac = -1.0
+
+        if self.powerspectrum:
+            # TODO: this seems to work, but I'm not sure if
+            # it's the correct way to do this
+            self._create_output = DelayGibbsSamplerBase()._create_output
```
ketiltrout, line 1010: I don't think it's _incorrect_, per se, but I think it might cause confusion, in that people trying to debug code later might not realise that the `_create_output` defined just below might be completely skipped in the program flow unexpectedly.

I don't know if there are implications for delaying this check, but, could you just have:
```
if self.powerspectrum:
  return DelayGibbsSamplerBase()._create_output(self, ss, delays, coord_axes)
```
at the top of the `_create_output` defined below?

It would also save you from having to repeat this diversion in `DelaySpectrumWienerEstimator.setup`.

ljgray: I agree that it's confusing, and I think your suggestion would be the right way to go about it. However, this line ends up getting somewhat refactored in #321 (the `_create_output` methods are defined independently of other functionality), so I think that PR is where we should consider changes.

ketiltrout: That's fair.

draco/util/dpss.py
```diff
@@ -0,0 +1,522 @@
...
+    xinp, winp = filter(x, Ni, A, W, Si)
+
+    xinp[W] = x[W]
+    winp[W] = Ni[W]
+
+    return xinp, winp
+
+
+def atleast_Nd(x: np.ndarray, N: int, lax: int = -1):
+    """Ensure that an array is at least `n` dimensional.
```
ketiltrout, line 443:
```suggestion
    """Ensure that an array is at least `N` dimensional.
```

draco/util/dpss.py
```diff
@@ -0,0 +1,522 @@
...
+        Array to expend dimensions.
+    N
+        Desired dimension of `x`.
+    lax
+        Axis to the left of where the new axes
+        are added. Default is -1, so new axes are
+        added to the end.
+    inverse
+        If True, return the slice that recovers the
+        original array dimension.
```
ketiltrout, line 462: There is no such parameter

ketiltrout (approved):
I haven't matched things up equation-to-equation, so it's possible I've glossed over something from the paper, but it looks like it should be doing what is advertised.

ljgray:
@ketiltrout I've made those fixes.

I've also changed the `fftw` dependency to be mandatory. I don't think it should be a requirement for `caput`, but `draco` already has some heavier dependencies and is only really going to be used in an analysis environment, so I think it's ok (but open to arguments against it)

ketiltrout:
> I've also changed the fftw dependency to be mandatory.

Yeah. That's reasonable here in the analysis code.  Making it optional in `caput` was because `caput` is also used in the realtime pipeline for ephemeris calculations.

pyproject.toml
```diff
@@ -24,6 +24,7 @@ requires-python = ">=3.9"
 dependencies = [
     "caput @ git+https://github.com/radiocosmology/caput.git",
     "caput[compression] @ git+https://github.com/radiocosmology/caput.git",
+    "caput[fftw] @ git+https://github.com/radiocosmology/caput.git",
```
ketiltrout, line 27: Does pyproject allow you to combine these?
```suggestion
    "caput[compression,fftw] @ git+https://github.com/radiocosmology/caput.git",
```

ljgray: Let's find out

PR #320, opened 2025-02-01
Adding hyforesbandpass module to implement hyfores

hcwang96:
The hyforesbandpass module estimates bandpass gains from daily hybrid beamformed visibilities, save the estimated gains, and the correct and delay filter the daily hybrid beamformed visibilities. Additional containers were added to the container module.

### DIFF
--- a/draco/analysis/hyforesbandpass.py
+++ b/draco/analysis/hyforesbandpass.py
@@ -355,6 +355,7 @@ class DelayFilterHyFoReSBandpassHybridVis(task.SingleTask):
         """
         return units.c / (freq * 1e6 * self.min_ysep) - 1.0
 
+
 class DelayFilterHyFoReSBandpassHybridVisMask(DelayFilterHyFoReSBandpassHybridVis):
     """Same as DelayFilterHyFoReSBandpassHybridVis but adding a pixel mask.
 
@@ -506,9 +507,9 @@ class DelayFilterHyFoReSBandpassHybridVisMask(DelayFilterHyFoReSBandpassHybridVi
         )  # generate a mask along the el axis to mask out aliased regions
 
         # Apply the pixel mask
-        post_vis *= mask[:,:,np.newaxis,:,:]
-        vis *= mask[:,:,np.newaxis,:,:]
-        
+        post_vis *= mask[:, :, np.newaxis, :, :]
+        vis *= mask[:, :, np.newaxis, :, :]
+
         self.log.debug("Start computing the estimated gains.")
         t0 = time.time()
         for pp in range(npol):
@@ -608,6 +609,7 @@ class DelayFilterHyFoReSBandpassHybridVisMask(DelayFilterHyFoReSBandpassHybridVi
         bp_gain_win.window[:] = W
 
         return bp_gain_win
+
 
 class HyFoReSBandpassHybridVis(DelayFilterHyFoReSBandpassHybridVis):
     """Same as DelayFilterHyFoReSBandpassHybridVis but does not implement the Delay filter.
@@ -776,6 +778,7 @@ class HyFoReSBandpassHybridVis(DelayFilterHyFoReSBandpassHybridVis):
         bp_gain_win.window[:] = W
 
         return bp_gain_win
+
 
 class DelayFilterHyFoReSBandpassHybridVisClean(task.SingleTask):
     """Second pipeline task of HyFoReS: Subtract foreground residuals using the estimated bandpass gains .
--- a/draco/analysis/hyforesbandpass.py
+++ b/draco/analysis/hyforesbandpass.py
@@ -1,0 +1,520 @@
+import numpy as np
+import ch_util
+from ch_util import ephemeris
+from astropy import constants as const
+import h5py
+from draco.core import containers
+import pickle
+from scipy import linalg as la
+import os
+import scipy
+from draco.analysis.ringmapmaker import find_grid_indices
+
+import time
+import scipy.interpolate
+from caput import config
+from cora.util import units
+from mpi4py import MPI
+
+from ..core import io, task
+from ..util import tools
+from . import transform
+
+##################################################################
+# Comments:
+# vis has the shape (pol, freq, ew, el, time)
+# filter has the shape (pol, freq, freq, ew, time)
+# vis_weight has the shape (pol, freq, ew, time)
+# time is basically ra
+#
+# HyFoReS
+#
+# Pipeline task 1:
+#
+# inputs: vis, filter
+#
+# step 1: get estimated signal (apply delay filter to vis): loop over pol, ew, time: post_vis[pol, :, ew, :, time] = filter[pol, :, :, ew, time].dot(vis[pol, :, ew, :, time]) --> distributed time axes work independently
+#
+# step 2: estimate bandpass gains: 
+#                               Both numerator and denominator are 0 if the freq channel is masked
+#                               Numerator:   loop over pol, freq: yN[pol, freq] = vis[pol, freq,...].vdot(post_vis[pol, freq,...]) --> sum over distributed time axes
+#                               Denominator: loop over pol, freq: yD[pol, freq] = vis[pol, freq,...].vdot(vis[pol, freq,...])      --> sum over distributed time axes
+#                               Estimated g: y[:, :] = yN[:,:]/yN[:,:] 
+#
+# step 3: compute the window matrix:
+#                               Numerator: loop over pol N[pol, :, :] = sum over ew, time: vis[pol,:, ew, :, time].conj().dot(vis[pol,:, ew, :, time].T())*filter[pol,:,:,ew, time] # the last product is element-wise product --> sum over distributed time axes
+#                                                  # this can be computed under the pol loop of the previous step
+#                               Dominator: loop over pol, freq_1: D[pol, freq_1] = vis[pol, freq_1, ...].vdot(vis[pol, freq_1, ...]) --> sum over distributed time axes # note this is the same as yD above, do so no need to compute again
+#                               Window:    loop over pol, freq_1: W[pol,freq_1,:] = N[pol, freq_1, :]/D[pol, freq_1]
+#
+# outputs: the estimated gains and window (new container required)
+#
+#
+# Pipeline task 2:
+#
+# inputs: vis, filter, estimate gains and window
+#
+# step 4: pseudo-invert the window:
+#                               Singular values:  loop over pol: u[pol], s_val[pol], vh[pol] = LA.svd(W[pol,:,:]) # need the singular value to determine the pseduo-inverse cut off: cut_off = 3e-2 by default
+#                               Pseudo-inverse:   loop over pol: W_pinv[pol,:,:], rank[pol] = LA.pinv(W[pol,:,:], atol = cut_off, return_rank = True) # return rank should be Nfreq - Nfreq_mask - Nfreq_delay_cut
+#                               Unwindowed gains: loop over pol: g[pol, :] = W_pinv[pol,:,:].dot(y[pol, :])
+# 
+# step 5: subtract foregrounds:
+#                               cleaned vis: loop over pol, ew, time: vis_cleaned[pol, :, ew, :, time] = post_vis[pol, :, ew, :, time] - filter[pol, :, :, ew, time].dot(np.diag(g[pol,:]).dot(vis[pol, :, ew, :, time])) --> distributed time axes work independently
+#                               # post_vis[pol, :, ew, :, time] is computed the same way as step 1
+#
+# output: cleaned vis
+#
+# Required Draco containers: VisBandpassWindowBaseline, VisBandpassCompensateBaseline
+#
+###################################################################
+
+
+class DelayFilterHyFoReSBandpassHybridVis(task.SingleTask): 
+    """Apply HyFoReS to estimate bandpass gains and compute their window matrix from unfiltered hybrid beam formed visibilities. 
+
+    This task builds on ApplyDelayFilterHybridVis. HyFoReS uses the unfiltered visibilities 
+    as estimated foregrounds and use the delay filtered visibilities as estimated signals. It
+    cross correlates the two to estimate the bandpass errors in the postfiltered data.
+
+    Attributes
+    ----------
+    atten_threshold : float
+        Used by the DAYENU filter.
+        Mask any frequency where the diagonal element of the filter
+        is less than this fraction of the median value over all
+        unmasked frequencies.  Default is 0.0 (i.e., do not mask
+        frequencies with low attenuation).
+    """
+
+    atten_threshold = config.Property(proptype=float, default=0.0)
+
+    def setup(self, manager):
+        """Extract the minimum baseline separation from the telescope class.
+
+        Parameters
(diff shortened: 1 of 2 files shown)


draco/analysis/hyforesbandpass.py
```diff
@@ -0,0 +1,520 @@
...
+        W = np.zeros_like(N_sum)
+        W = N_sum*tools.invert_no_zero(D_sum[:, :, :, np.newaxis])
+
+        # put the estimated gains and window in a container
+        bp_gain_win = containers.VisBandpassWindowBaseline(
+                pol= hv.pol,
+                ew = hv.ew,
+                freq= hv.freq,
+                )
+        bp_gain_win.bandpass[...] = y # new question: do I need [...] here? should check out other examples
```
ljgray, line 288: You do need a slice here, but typically you would just use `[:]` when slicing into an entire array. `...` is used to expand more complicated slices to an arbitrary number of indices - for example, if I wanted to index the second-last axis of an array with arbitrary dimension, I could do `arr[..., 0, :]`

draco/analysis/hyforesbandpass.py
```diff
@@ -0,0 +1,520 @@
...
+
+        Returns
+        -------
+        horizon_limit : np.ndarray[nfreq,]
+            This is the value of sin(za) where the southern horizon aliases.
+            Regions of sky where ``|sin(za)|`` is greater than or equal to
+            this value will contain aliases.
+        """
+        
+        return scipy.constants.c / (freq * 1e6 * self.min_ysep) - 1.0
```
ljgray, line 330: This is nitpicking, but for consistency we used `units.c` from `cora.util`, which is already imported in this module. Alternatively, it's ok if you use `scipy.constants.c`, but you should make sure that's used consistently in this module and then remove the `from cora.util import units` import

draco/analysis/hyforesbandpass.py
```diff
@@ -0,0 +1,520 @@
...
+
+
+# Add the option to compensate the window or not.
```
ljgray, line 333: If this is a todo item, add `TODO: ` at the start. It's just a common practice, and will be highlighted correctly when reading the code

draco/analysis/hyforesbandpass.py
```diff
@@ -0,0 +1,520 @@
...
+        the window matrix. Default is 3e-2.
+    atten_threshold : float
+        Used by the DAYENU filter.
+        Mask any frequency where the diagonal element of the filter
+        is less than this fraction of the median value over all
+        unmasked frequencies.  Default is 0.0 (i.e., do not mask
+        frequencies with low attenuation).
+    """
+
+    # if the cutoff is 0.0, then do not compensate the window
```
ljgray, line 356: This comment should be moved to the variable description in the `Attributes` section

draco/analysis/hyforesbandpass.py
```diff
@@ -0,0 +1,520 @@
...
+
+        npol, nfreq, new, nel, ntime = hv.vis.local_shape ### Q:what is local_shape? Shape of the data distributed to a node, not the total shape?
+
+        # Step 4
+        # step 4: pseudo-invert the window:
+        # Singular values:  loop over pol: u[pol], s_val[pol], vh[pol] = LA.svd(W[pol,:,:]) # need the singular value to determine the pseduo-inverse cut off: cut_off = 3e-2 by default
+        # Pseudo-inverse:   loop over pol: W_pinv[pol,:,:], rank[pol] = LA.pinv(W[pol,:,:], atol = cut_off, return_rank = True) # return rank should be Nfreq - Nfreq_mask - Nfreq_delay_cut
+        # Unwindowed gains: loop over pol: g[pol, :] = W_pinv[pol,:,:].dot(y[pol, :])
+        # Compensate the window for the estimated bandpass gains
+        y = bp.bandpass[...]
```
ljgray, line 411:
```suggestion
        y = bp.bandpass[:]
```

draco/analysis/hyforesbandpass.py
```diff
@@ -0,0 +1,520 @@
...
+        else:
+
+            self.log.debug("Start compensating the window")
+
+            for pp in range(npol):
+
+                for xx in range(new):
+
+                    u, s_val[pp, xx], vh = la.svd(W[pp, xx, :,:]) #Q6: is scipy.linalg imported as LA somewhere in the pipeline? #la lower case # create new module in draco analysis
+                    W_pinv[pp, xx, :, :], rank[pp, xx] = la.pinv(W[pp, xx, :, :], atol = self.cutoff, return_rank = True)
```
ljgray, line 432: If you don't need to save out `W_pinv`, this will be dramatically faster by using a `solve` instead, i.e. `g = la.solve(W, y)`. Off the top of my head, I don't know exactly what the structure of `W` is, but if is hermitian positive-definite, using `WL = la.cho_factor(w)` followed by `g = la.cho_solve(Wl, y)` is extremely fast. Otherwise, just using `la.solve` will still be significantly faster and more numerically stable than a pseudo-inverse.

draco/analysis/hyforesbandpass.py
```diff
@@ -0,0 +1,520 @@
...
+                    g[pp, xx, :] = np.dot(W_pinv[pp, xx, :, :], y[pp, xx, :])
+
+            self.log.debug("Gain window compensated")
+        
+        # put unwinded bandpass gains in a container. This container is not needed in later pipeline tasks. Save for debug purposes only.
+        comp_bandpass = containers.VisBandpassCompensateBaseline(
```
ljgray, line 438: Similar to above, this should be initialised using the `axes_from` and `attrs_from` arguments

draco/analysis/hyforesbandpass.py
```diff
@@ -0,0 +1,520 @@
...
+            self.log.debug("Skip compensating the window")
+        else:
+
+            self.log.debug("Start compensating the window")
+
+            for pp in range(npol):
+
+                for xx in range(new):
+
+                    u, s_val[pp, xx], vh = la.svd(W[pp, xx, :,:]) #Q6: is scipy.linalg imported as LA somewhere in the pipeline? #la lower case # create new module in draco analysis
```
ljgray, line 431: It doesn't look like this svd is being used anywhere

draco/analysis/hyforesbandpass.py
```diff
@@ -0,0 +1,520 @@
...
+                        nonzero_diag_flag = diag > 0.0
+                        if np.any(nonzero_diag_flag):
+                            med_diag = np.median(diag[nonzero_diag_flag])
+                            flag_low = diag > (self.atten_threshold * med_diag)
+                            weight[pp, :, xx, tt] *= flag_low.astype(weight.dtype) ### this masking is only done on the weight
+
+            self.log.debug(f"Took {time.time() - t0:0.3f} seconds in total.")
+
+        # Problems saving to disk when distributed over last axis
+        hv.redistribute("freq") ### why redistribute again? How does this work when the data are aready distributed over the ra axis on different nodes?
```
ljgray, line 518: Typically we keep datasets distributed over frequency because most operations we do are frequency-independent, and it's the fastest axis to distribute over when writing files to disk.

draco/core/containers.py
```diff
@@ -3340,3 +3340,227 @@ def copy_datasets_filter(
         if isinstance(dest_dset, memh5.MemDatasetDistributed):
             # Redistribute back to the original axis
             dest_dset.redistribute(original_ax_id)
+
+
+    # New container for the estimated bandpass gains and window
+class VisBandpassWindow(FreqContainer):
+    """Container for bandpass gains and their window estimated by running bandpass HyFoReS on hybrid beam-formed visibilities"""
+
+    _axes = ("pol", "freq") #Q8: do I need to declare _axes? What is the point of this? might have to add pol
```
ljgray, line 3349: You will only have to declare the polarisation axis here, since the frequency axis is derived from `FreqContainer`

draco/core/containers.py
```diff
@@ -3340,3 +3340,227 @@ def copy_datasets_filter(
...
+
+    _dataset_spec: ClassVar = {
+        "bandpass": {
+            "axes": ["pol", "freq"],
+            "dtype": np.complex128,
```
ljgray, line 3354: I would check if 64-bit precision is required here. Visibilities are 32-bit, and we only upcast to 64-bit in a handful of places where some complex operation requires the higher precision.

draco/core/containers.py
```diff
@@ -3340,3 +3340,227 @@ def copy_datasets_filter(
...
+        },
+        "window": {
+            "axes": ["pol", "freq", "freq"],
+            "dtype": np.complex128,
+            "initialise": True,
+            "distributed": False,
+        },
+    }
+
+    #_data_dset_name = "stack"
```
ljgray, line 3366: These can just be removed instead of commented out

draco/core/containers.py
```diff
@@ -3340,3 +3340,227 @@ def copy_datasets_filter(
...
+
+    _dataset_spec: ClassVar = {
+        "comp_bandpass": {
+            "axes": ["pol", "freq"],
+            "dtype": np.complex128,
+            "initialise": True,
+            "distributed": False,
+        },
+        "sval": {
+            "axes": ["pol", "freq"], # want an axis that has the same length as the freq axis "mode"
```
ljgray, line 3392: What is the frequency axis here? If it's a series of modes with axis length equal to frequency, rather than the actual frequencies, you should create a new axis with a descriptive name

draco/core/containers.py
```diff
@@ -3340,3 +3340,227 @@ def copy_datasets_filter(
...
+    @property
+    def sval(self):
+        """Get the sval dataset."""
+        return self.datasets["sval"]
+
+    
+class VisBandpassWindowBaselineRA(FreqContainer):
+    """Container for bandpass gains and their window estimated by running bandpass HyFoReS on hybrid beam-formed visibilities"""
+
+    _axes = ("pol", "ew", "ra", "freq") #Q8: do I need to declare _axes? What is the point of this? might have to add pol
```
ljgray, line 3492: For data with an ra axis, you should inherit from `SiderealContainer`. In this case, you can inherit from both `VisBandpassWindowBaseline` and `SiderealContainer`, and then you do not have to define `_axes` and you _only_ have to add the `_dataset_spec`

draco/core/containers.py
```diff
@@ -3340,3 +3340,227 @@ def copy_datasets_filter(
...
+    def comp_bandpass(self):
+        """Get the comp_bandpass dataset."""
+        return self.datasets["comp_bandpass"]
+
+    @property
+    def sval(self):
+        """Get the sval dataset."""
+        return self.datasets["sval"]
+
+class VisBandpassWindowBaseline(FreqContainer):
```
ljgray, line 3424: This class can inherit directly from `VisBandpassWindow`, which already implements the `bandpass` and `window` properties, so you would just have to define the `_dataset_spec` and `_axes = ("ew",)`.

draco/core/containers.py
```diff
@@ -3340,3 +3340,227 @@ def copy_datasets_filter(
...
+        """Get the bandpass dataset."""
+        return self.datasets["bandpass"]
+
+    @property
+    def window(self):
+        """Get the window dataset."""
+        return self.datasets["window"]
+
+
+class VisBandpassCompensateBaseline(FreqContainer):
```
ljgray, line 3458: This can inherit from `VisBandpassCompensate`

draco/core/containers.py
```diff
@@ -3340,3 +3340,227 @@ def copy_datasets_filter(
...
+    def bandpass(self):
+        """Get the bandpass dataset."""
+        return self.datasets["bandpass"]
+
+    @property
+    def window(self):
+        """Get the window dataset."""
+        return self.datasets["window"]
+
+class VisBandpassCompensateBaselineRA(FreqContainer):
```
ljgray, line 3530: Same comment as above with inheriting from `VisBandpassCompensateBaseline` and `SiderealContainer`

ljgray (changes requested):
This implementation looks good overall. I've made a bunch of comments, but most of them should be minor fixes to bring things in line with chime coding standards. I think there are a couple of comments with places where things could be sped up, or code/variable that aren't actually used.

One general thing - unless a comment about a single line of code is one or two words, it's standard to place the comment on the line(s) above the code, rather than in the same line. This is because python standards enforce maximum line lengths, so when a comment is on the same line it will force the code and comment to extend to multiple lines in a way that's difficult to read. Comments lines should also typically be shorter than code lines (max ~70 characters I think?), so long comments can be extended to multiple lines.

draco/analysis/hyforesbandpass.py
```diff
@@ -0,0 +1,520 @@
...
+                        diag = np.abs(np.diag(NF))
+                        nonzero_diag_flag = diag > 0.0
+                        if np.any(nonzero_diag_flag):
+                            med_diag = np.median(diag[nonzero_diag_flag])
+                            flag_low = diag > (self.atten_threshold * med_diag)
+                            weight[pp, :, xx, tt] *= flag_low.astype(weight.dtype) ### this masking is only done on the weight
+                            # Now apply this masking to the filtered visibilities as well
+                            atten_mask = flag_low.astype(vis.dtype)
+                            post_vis[pp, :, xx, :, tt] *= atten_mask[:, np.newaxis]
+
```
ssiegelx, line 214: Rather than have three separate loops over pol, ew baseline, and freq/time, I'm wondering if we could just move everything into the first loop?  I think this would reduce run time, reduce memory usage since you would not have to create a `post_vis` array that is the same shape as the input visibilities, and simplify the code.

I think it would involve inserting something like on this line after you've dealt with the attenuation threshold:
```
post_vis = np.matmul(NF, tvis)
w = (weight[pp, :, xx, tt, np.newaxis] > 0) & el_mask
wtvis = w * tvis
yN[pp, xx, :] += np.sum(wtvis * post_vis, axis=-1)
D[pp, xx, :] += np.sum(wtvis * tvis, axis=-1)
N[pp, xx, :, :] += np.matmul(np.conj(wtvis), wtvis.T) * NF
```

ssiegelx (approved):
I'm approving this. There are a few things I’d recommend addressing before merging, though I don't think they're strictly necessary:

* A few comments from the last round haven't been addressed.
* Many of the class docstrings exceed 88 characters. These should be shortened or split onto secondary lines.
* For the module-level docstring: I think it's helpful to describe what each task does and how it should be used, but I’d suggest removing the pseudo-code; it’s a bit confusing as-is.  If you prefer to keep it, consider shortening each line to stay within the 88-character limit.
* There are several places with commented-out code.  If you'd like to keep this, I suggest adding comments to clarify what the code is and why it's there.

I’ve made a small modification on top of this branch that adds a config property to apply the gains to the data while bypassing the filter. I’m using this to generate an unfiltered dataset that’s otherwise processed identically to the filtered one, which is necessary for generating noise realizations. I can either push this change to this branch now or open a separate PR once this is merged.

ljgray (approved):
I'm approving as well, with pretty much the same message as Seth. I'd like to see the docstring and comment lengths addressed before merging, at least.

I think, at some point, we should revisit some of the for loops in here, because I suspect that we can both vectorize some operations and cut down on the number of repeated loops for different steps in the process. However, this doesn't have to happen immediately.

PR #322, opened 2025-02-25
Tasks and containers for directional RFI mask

yukari-u:
The current RFI flagging allows only frequency and time/RA masks. I added two containers, **LocalizedRFIMask(freq, el, time)** and **LocalizedSiderealRFIMask(freq, el, ra)**, in core/containers.py, as well as three tasks, **SiderealMaskConversion**, **ReduceMaskEl**, and **ApplyLocalizedRFIMask**, in analysis/flagging.py, to enable elevation-specific directional RFI flagging in addition to freq and time/RA.

### DIFF
--- a/draco/analysis/flagging.py
+++ b/draco/analysis/flagging.py
@@ -2981,3 +2981,259 @@ def destripe(x, w, axis=1):
     bsel = tuple(bsel)
 
     return x - stripe[bsel]
+
+
+class SiderealMaskConversion(task.SingleTask):
+    """Convert the axis of an RFI mask from time to ra.
+
+    Attributes
+    ----------
+    spread_size : int
+        The number of cells to flag before and after a detected true value. This ensures
+        conservative flagging, preventing missed detections due to axis alignment issues.
+        Default is 1.
+    npix : int
+        The number of pixels used to cover the full RA range from 0 to 360.
+         Defualt is 4096.
+    """
+
+    spread_size = config.Property(proptype=int, default=1)
+    npix = config.Property(proptype=int, default=4096)
+
+    def setup(self, manager):
+        """Set the local observers position.
+
+        Parameters
+        ----------
+        manager : :class:`~caput.time.Observer`
+            An Observer object holding the geographic location of the telescope.
+            Note that :class:`~drift.core.TransitTelescope` instances are also
+            Observers.
+        """
+        # Need an Observer object holding the geographic location of the telescope.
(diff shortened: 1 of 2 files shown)


draco/analysis/flagging.py
```diff
@@ -2981,3 +2981,259 @@ def destripe(x, w, axis=1):
     bsel = tuple(bsel)
 
     return x - stripe[bsel]
+
+
+class SiderealMaskConversion(task.SingleTask):
+    """Convert the axis of an RFI mask from time to ra.
+
```
ljgray, line 2988: It would be helpful to have a quick explanation here about the method this task uses to convert from time to RA.

draco/analysis/flagging.py
```diff
@@ -2981,3 +2981,259 @@ def destripe(x, w, axis=1):
...
+        Parameters
+        ----------
+        rfimask : containers.LocalizedRFIMask
+            Container for holding a mask indicating channels that are free from RFI events.
+            Its axes are freq, el, and time.
+
+        Returns
+        -------
+            out : containers.LocalizedSiderealRFIMask
```
ljgray, line 3027:
```suggestion
        out : containers.LocalizedSiderealRFIMask
```

draco/analysis/flagging.py
```diff
@@ -2981,3 +2981,259 @@ def destripe(x, w, axis=1):
...
+            Boolean mask that can be applied to a ringmap
+            with the task `ApplyLocalizedRFIMask` to mask contaminated samples.
+            Its axes are freq, ra, and el.
+
```
ljgray, line 3031:
```suggestion
```

draco/analysis/flagging.py
```diff
@@ -2981,3 +2981,259 @@ def destripe(x, w, axis=1):
...
+        output.mask[:] = final_mask
+
+        # Return output container
+        return output
+
+
+class ApplyLocalizedRFIMask(task.SingleTask):
+    """Apply a localised (el-sensitive) RFI mask to the data by zeroing the weights.
+
```
ljgray, line 3155: This docstring should mention that the data stream and mask datasets must have overlapping RA/El samples, but those samples are not assumed to be the same.

ljgray (changes requested):
This looks great overall! I just have a couple of comments about details to include in the task docstrings.

ljgray (approved):

PR #323, opened 2025-03-03
feat(noise): task that generates multiple gaussian noise datasets

ssiegelx:
Add MultipleGaussianNoiseDatasets task to iteratively generate many Gaussian noise realizations from an underlying weight dataset provided on setup. Introduces an in_place config property to GaussianNoiseDatasets to support this feature.


ljgray (approved):
This looks good. I think the commit history just needs to be cleaned up via rebase, not sure why it's showing some of these old commits.

Arnab-half-blood-prince (approved):
looks good.

ljgray (approved):
Looks good

PR #324, opened 2025-03-06
Speed up mmode and BeamformNS

ljgray:
There are a few critical improvements here:
## BeamformNS
- Use the `out` kwarg where possible to avoid unnecessary duplicate arrays
- Use a faster method to find masked baselines

## MModeTransform
- Use `.local_array` to avoid `mpiarray` read/writes
- Remove `for` loops during mmode transform and write
- Use fast `fftw` module for m transform
- Apply m-mode normalization when writing to the `mmode` array - this was surprisingly almost ⅓ of the function runtime

### DIFF
--- a/draco/analysis/transform.py
+++ b/draco/analysis/transform.py
@@ -7,7 +7,7 @@ from typing import overload
 
 import numpy as np
 import scipy.linalg as la
-from caput import config, mpiarray, pipeline
+from caput import config, fftw, mpiarray, pipeline
 from caput.tools import invert_no_zero
 from numpy.lib.recfunctions import structured_to_unstructured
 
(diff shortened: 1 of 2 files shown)


ssiegelx:
I'm testing this now.  One thought: since this is an optional dependency, should we default to numpy if fftw fails to import?

ljgray:
> One thought: since this is an optional dependency, should we default to numpy if fftw fails to import?

Yeah that's a good idea. I added an extra commit

draco/analysis/transform.py
```diff
@@ -664,16 +665,15 @@ def _make_marray(ts, mmodes=None, mmax=None, dtype=None):
...
-
-        for mi in range(1, mlim_neg + 1):
-            mmodes[mi, 1, i] = m_fft[..., -mi].conj()
+    # Move the M axis to the front
+    m_fft = np.moveaxis(m_fft, -1, 0)
+    # Write the positive and negative m's
+    mmodes[:, 0] = m_fft[: mlim + 1]
+    mmodes[:, 1] = m_fft[::-1][: mlim_neg + 1]
+    # Conjugate and write in-place
+    mmodes[:, 1] = np.conjugate(mmodes[:, 1], out=mmodes[:, 1])
```
ssiegelx, line 676:
```suggestion
    mmodes[:npos, 0] = m_fft[:npos]
    mmodes[1:nneg, 1] = m_fft[-1:-nneg:-1]
    # Conjugate and write in-place
    np.conjugate(mmodes[1:nneg, 1], out=mmodes[1:nneg, 1])
```
where
```
npos = mlim + 1
nneg = mlim_neg + 1
```
The m=0 mode is stored in the positive-m array.  The negative-m array is filled starting at m=1.

ssiegelx (approved):
This all looks good.  I agree that we should check whether changing the precision of the beamforming matrix has any significant impact on the results.

ljgray:
@ssiegelx I did a brief test comparing 32 and 64 bit precision. For an unfiltered ringmap, the typical absolute error is on the order 10^-5, and the relative error 10^-6. The maximum absolute error is ordering 10^-2 and maximum relative order 10^1 (note the positive 1). 

What are your thoughts?

ssiegelx:
Yeah seeing a factor of 10 error is a bit troubling.  Is this a noise realization or a map of the data?

ljgray:
> Yeah seeing a factor of 10 error is a bit troubling. Is this a noise realization or a map of the data?

This was just a map of the data.

Arnab-half-blood-prince:
What will happen to keep the data to 64bit? If that gives us the best result, I think we should go ahead with that. We may hit by some extra time, but that is fine.

ssiegelx:
It may behave better with noise realizations.  We could have a config property that can be switched to perform the computation in float32 for the noise realizations, but defaults to float64.  I'm also fine with just switching back to float64.

ljgray:
The overall runtime of the task increases by about a factor of 3 for 64 bit matmul. I lean towards just sticking with 64-bit, but it wouldn't be difficult to add a config property

ssiegelx:
Hmm that's a significant reduction, so I would be inclined to add the config property.  But that's contingent on the errors being reduced for the noise realizations.  Would that be easy for you to test?  If not I should have some time tomorrow morning.

ljgray:
> @ssiegelx I got around to testing this (with just a single noise realization). Mean absolute error: order 10^-9 Mean relative error: order 10^-6 Max absolute error: order 10^-7 Max relative error: order 1
> 
> Overall lower, but the max relative error is still somewhat high

However, the large relative error seems to only be for values very close to zero, so maybe that's fine.

ssiegelx (approved):

PR #325, opened 2025-03-06
Update BlendStack to support different container types

ljgray:
Support `Ringmap` and `HybridVisStream` data. Still needs testing.


yukari-u:
I tested BlendStack with some data and confirmed it works properly. I only updated the doctoring.

PR #326, opened 2025-03-09
feat(ringmapmaker): ability to skip calculating the redundancy

ssiegelx:
Calculating the redundancy is computationally expensive and is only required if using a natural weighting scheme in `BeamformNS`.  This PR adds a config property `save_redundancy` to `MakeVisGrid` that when set to False will skip the calculation.

The PR also speeds up MakeVisGrid significantly by accessing local arrays outside of the for loop, preventing repeated MPIArray calls.  It fixes a longstanding bug in how redundancy was assigned that affected intracylinder baselines.  It also resolves some warnings related to MPIArray operations in RADependentWeights.

### DIFF
--- a/draco/analysis/ringmapmaker.py
+++ b/draco/analysis/ringmapmaker.py
@@ -1165,8 +1165,12 @@ class RADependentWeights(task.SingleTask):
             )
             raise RuntimeError(msg)
 
+        # Ensure containers are distributed over the same axis
+        hybrid_vis.redistribute("freq")
+        ringmap.redistribute("freq")
+
         # Extract the variance of the hybrid visibilities from the weight dataset
-        var = tools.invert_no_zero(hybrid_vis.weight[:].view(np.ndarray))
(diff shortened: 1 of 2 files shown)


draco/analysis/ringmapmaker.py
```diff
@@ -223,7 +234,8 @@ def process(self, gstream):
 
         gsv = gstream.vis[:].local_array
         gsw = gstream.weight[:].local_array
-        gsr = gstream.redundancy[:]
+        if self.weight == "natural":
+            gsr = gstream.redundancy[:]
```
ljgray, line 238: Because the redundancy dataset is no longer guaranteed to exist, think there should be an extra check that either switches to a different weighting method or raises an informative error (probably the better option).

ljgray (approved):
Looks good, just one comment

Arnab-half-blood-prince (approved):
looks good.

PR #328, opened 2025-03-11
feat(ringmapmaker): create ReconstructVisWeight task

ssiegelx:
Generates visibility weights that will match hybrid beamformed
visibility weights after beamforming north-south.  Useful for
generating noise realizations of hybrid beamformed visibilities
that maintain proper correlation along the el axis.

### DIFF
--- a/draco/analysis/transform.py
+++ b/draco/analysis/transform.py
@@ -86,7 +86,58 @@ class FrequencyRebin(task.SingleTask):
         return sb
 
 
-class CollateProducts(task.SingleTask):
+class TelescopeStreamMixIn:
+    """A mixin providing functionality for creating telescope-defined sidereal streams.
+
+    This mixin is designed to be used with pipeline tasks that require certain
+    index maps in order to create SiderealStream containers compatible with the
+    baseline configuration provided in a telescope instance.
+    """
+
+    def setup(self, tel):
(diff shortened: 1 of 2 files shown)


draco/analysis/transform.py
```diff
@@ -155,6 +136,45 @@ def setup(self, tel):
...
+    Parameters
+    ----------
+    weight : string ('natural', 'uniform', or 'inverse_variance')
+        How to weight the redundant baselines when stacking:
+            'natural' - each baseline weighted by its redundancy (default)
+            'uniform' - each baseline given equal weight
+            'inverse_variance' - each baseline weighted by the weight attribute
+    """
+
+    weight = config.Property(proptype=str, default="natural")
```
ljgray, line 163: We may as well change this to an `enum` and not bother removing the check in `setup` - that way an error could get caught by `caput-pipeline lint`

ljgray (approved):
Looks good.

draco/analysis/transform.py
ssiegelx: Oh yeah, good point.

Arnab-half-blood-prince (commented):
It looks good.  also keeping this image here to make sure what we see..here I have plotted the stream weight and the reconstructed weight from hybrid vis (made from the same stack).

image

PR #330, opened 2025-03-13
Task to mix exactly two datasets in a single iteration

ljgray:
This is untested so far, but I think it should work.

### DIFF
--- a/draco/analysis/transform.py
+++ b/draco/analysis/transform.py
@@ -1421,7 +1421,12 @@ class MixData(task.SingleTask):
         self._tags = []
         self._wfunc = tools.invert_no_zero if self.invert_weight else lambda x: x
 
-    def process(self, data: containers.SiderealStream | containers.RingMap):
+    def process(
+        self,
+        data: (
+            containers.SiderealStream | containers.HybridVisStream | containers.RingMap
+        ),
+    ):
         """Add the input data into the mixed data output.
 
         Parameters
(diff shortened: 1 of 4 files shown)


draco/analysis/transform.py
```diff
@@ -1534,6 +1534,26 @@ class Jackknife(MixData):
     require_nonzero_weight = config.Property(proptype=bool, default=True)
 
 
+class MixTwoDatasets(MixData):
```
ssiegelx, line 1537: If we are adding an HI simulation to the daily data during the quarterstacking stage, we will iterate over days, mix in the simulation, estimate/correct the gains, and then add the mixed data to the quarterstack.  I think the mixed data needs to be returned as an output of `process`, but here it looks like it will be returned as an output of `process_finish`. If it's output at `process_finish`, will the pipeline then loop back and start the next day in the list?

ljgray: Oh yeah, you're right, it will have to move to `process`

ssiegelx: And will `process_finish` be called for each day, or will the pipeline add all of the days to the mix before finally calling `process_finish`?

ljgray: I've made some changes which should address this

draco/analysis/transform.py
```diff
@@ -1488,7 +1488,7 @@ def process(self, data: containers.SiderealStream | containers.RingMap):
 
         self._data_ind += 1
 
-    def process_finish(self) -> containers.SiderealStream | containers.RingMap:
+    def _make_output(self) -> containers.SiderealStream | containers.RingMap:
```
ssiegelx, line 1491: We will need to mix the `HybridVisStream`.  I think this can just be added to the type list (both here and in the `process` method of `MixData`) and it should work without further modification to the actual pipeline task.

PR #331, opened 2025-03-14
Fix applying window and fft of map.

Arnab-half-blood-prince:
Apply the window to the data before applying the data cut. Also use numpy fft when working with ringmap, as fftw only supports complex to complex transform

### DIFF
--- a/draco/analysis/delay.py
+++ b/draco/analysis/delay.py
@@ -1081,6 +1081,12 @@ class DelayTransformFFT(DelayGeneralContainerBase, DelayGibbsSamplerBase):
 
             data = data_view.local_array[lbi]
             weight = weight_view.local_array[lbi]
+            
+            # Apply the window if requested
+            # Do this before applying data cut
+            # as that will change the data shape
+            if self.apply_window:
+                data *= window            
 
             # Apply data cuts
             t = self._cut_data(data, weight)
@@ -1088,12 +1094,15 @@ class DelayTransformFFT(DelayGeneralContainerBase, DelayGibbsSamplerBase):
                 continue
             data, _, _, nzt = t
 
-            # Apply the window if requested
-            if self.apply_window:
-                data *= window
+            # Take the inverse Fourier transform
+            # Note fftw only supports complex to
+            # complex transform. For real value ringmap
+            # use numpy fft.
+            if data.dtype is complex:
+                y_spec = fftw.ifft(data, axes=-1)
+            else:
+                y_spec = np.fft.ifft(data, axis=-1)
 
-            # Take the inverse fourier transform
-            y_spec = fftw.ifft(data, axes=-1)
             # fftshift over the transform axis
             y_spec = np.fft.fftshift(y_spec, axes=-1)
 
--- a/draco/analysis/delay.py
+++ b/draco/analysis/delay.py
@@ -1081,12 +1081,12 @@ class DelayTransformFFT(DelayGeneralContainerBase, DelayGibbsSamplerBase):
 
             data = data_view.local_array[lbi]
             weight = weight_view.local_array[lbi]
-            
+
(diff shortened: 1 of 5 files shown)


draco/analysis/delay.py
```diff
@@ -1082,18 +1082,27 @@ def _evaluate(self, data_view, weight_view, out_cont, delays, channel_ind):
             data = data_view.local_array[lbi]
             weight = weight_view.local_array[lbi]
 
+            # Apply the window if requested
+            # Do this before applying data cut
+            # as that will change the data shape
+            if self.apply_window:
```
ljgray, line 1088: Yeah this makes sense. I think the reason it got through is because with `freq_frac=-1` no frequency cuts are applied, so the data was unchanged along that axis.

draco/analysis/delay.py
```diff
@@ -1082,18 +1082,27 @@ def _evaluate(self, data_view, weight_view, out_cont, delays, channel_ind):
...
             data, _, _, nzt = t
 
-            # Apply the window if requested
-            if self.apply_window:
-                data *= window
+            # Take the inverse Fourier transform
+            # Note fftw only supports complex to
+            # complex transform. For real value ringmap
+            # use numpy fft.
+            if data.dtype is complex:
```
ljgray, line 1101:
```suggestion
            if np.iscomplexobj(data):
```

Numpy dtypes don't agree with python, so this will end up always returning False. Just use the `numpy` built-in instead.

ljgray (approved):
Yeah this looks good. I'll update `fftw` to support r2c ffts eventually, but this is fine. Just the one change on the `dtype` check.

draco/analysis/delay.py
```diff
@@ -1082,18 +1082,27 @@ def _evaluate(self, data_view, weight_view, out_cont, delays, channel_ind):
...
-            if self.apply_window:
-                data *= window
+            # Take the inverse Fourier transform
+            # Note fftw only supports complex to
+            # complex transform. For real value ringmap
+            # use numpy fft.
+            if np.iscomplexobj(data):
+                y_spec = fftw.ifft(data, axes=-1)
+            else:
+                y_spec = np.fft.ifft(data, axis=-1)
```
ketiltrout, line 1104:
```suggestion
                y_spec = np.fft.irfft(data, axis=-1)
```
Since you know the data are purely real in this case, using `irfft` should be faster, since that function will exploit the Hermitian nature of the Fourier transform.

ketiltrout (approved):
> I'll update fftw to support r2c ffts eventually

This is a good idea, though how much faster is `fftw` than numpy?  Is it fast enough that it's better to convert to complex and use FFTW than using `numpy.fft`?

draco/analysis/delay.py
Arnab-half-blood-prince: Although the Map is in real space (ra,dec,freq), we want to transform that to complex Fourier field (u,v,delay), and for power spectrum, we want to take complex correlation of two such fields and then take the real part of that. So, I think we want to keep the full complex quantity. This is what we do in the Wiener filter too, by providing complex_timedomain=True.

ketiltrout: Oh, right.  This is an inverse transform from real.  You're right, `irfft` isn't what you want.

But `ihfft` should do the trick.  It takes a purely real spectrum and outputs a (Hermitian-symmetric) complex signal.

`ifft` will work, too, but it will ignore the symmetric nature of the output (i.e. in terms of computational load, it will compute the inverse FFT as if the input were complex).

Arnab-half-blood-prince: keeping it as `ifft` for now, as this is what we are following in other parts of the code, for example the Fourier definition in Gibbs and Weiner filter code is this.

PR #333, opened 2025-03-19
Update timestream synthesis tasks to take time input from another dataset

ljgray:
Adds tasks to interpolate a sidereal dataset onto a time axis provided by another container.
1. Main tasks iterates sidereal stream and time axis
2. Second task iterates time axis for a fixed sidereal stream
3. Third task iterates sidereal stream for a fixed time axis

Also, refactors the existing task.

Interpolation is just done using the lanczos kernel, with no attempt made to compute weights (under the assumption that we don't care). If weights are needed, we can figure out how to do that.

Seems to work.

### DIFF
--- a/draco/synthesis/stream.py
+++ b/draco/synthesis/stream.py
@@ -6,11 +6,14 @@ expand any redundant products with :class:`ExpandProducts` and finally generate
 a set of time stream files with :class:`MakeTimeStream`.
 """
 
+import inspect
+
 import numpy as np
-from caput import config, mpiarray, mpiutil, pipeline
+from caput import config, mpiarray, mpiutil, pipeline, tools
 from cora.util import hputil
 
 from ..core import containers, io, task
+from ..util import regrid
 
 
 class SimulateSidereal(task.SingleTask):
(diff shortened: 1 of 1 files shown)


draco/synthesis/stream.py
```diff
@@ -243,6 +246,162 @@ def process(self, sstream):
...
+        # Provide both the time and ra axes - whichever one isn't
+        # used will be ignored
+        out = OutputContainer(axes_from=sstream, attrs_from=sstream, time=time, ra=tra)
+
+        # Make sure datasets are distributed correctly
+        sstream.redistribute("freq")
+        out.redistribute("freq")
+
+        # Make the interpolation array
+        R = regrid.lanczos_forward_matrix(sstream.ra, tra, self.lanczos_width).T.copy()
```
ssiegelx, line 323:
```suggestion
        R = regrid.lanczos_forward_matrix(sstream.ra, tra, self.lanczos_width, periodic=True).T.copy()
```

draco/synthesis/stream.py
```diff
@@ -243,6 +246,162 @@ def process(self, sstream):
...
+
+        # Move the RA axis to the back
+        axind = list(sstream.data.attrs["axis"]).index("ra")
+        data = np.moveaxis(sstream.data[:].local_array, axind, -1)
```
ssiegelx, line 327: For both supported container types, `ra` is already along the last axis.  Is this intended as a safeguard for potential future container types? Also, does this operation have any effect if `axind` is already the last axis, or is it effectively a no-op?

draco/synthesis/stream.py
```diff
@@ -243,6 +246,162 @@ def process(self, sstream):
...
+
+        # Apply the interpolation matrix
+        data_ts = data @ R
+
+        # Move the time-like axis back to where it belongs
+        out.data[:].local_array[:] = np.moveaxis(data_ts, -1, axind)
+        # Just set the weights to one
+        out.weight[:].local_array[:] = 1.0
```
ssiegelx, line 335: I would probably just implement the weight propagation.  I think it would follow something like:
```
waxind = list(sstream.weight.attrs["axis"]).index("ra")
var = np.moveaxis(tools.invert_no_zero(sstream.weight[:].local_array), waxind, -1)
out.weight[:].local_array[:] = np.moveaxis(tools.invert_no_zero(var @ R ** 2), -1, waxind)
```

draco/synthesis/stream.py
ssiegelx: For a `HybridVisStream` generated directly from a `TimeStream`, if the time axis extends beyond the bounds of a sidereal day, it will take on values that are either negative or greater than 360.  This ensures that lines like:
```
time = self.observer.lsd_to_unix(lsd + tra / 360.0)
```
still produce the correct time. So, I believe that in addition to setting `periodic = True`,  we also need to wrap `tra` between 0 and 360 before passing it to this task, i.e., `tra % 360.0`.

ssiegelx (approved):
Thanks for putting this together.  It looks great.  I just have a few comments.

draco/synthesis/stream.py
ljgray: Yeah, my thinking was just if we ever wanted to use it on a ring map or something. moveaxis returns a view, so if the axes are the same it's just a no-op

PR #338, opened 2025-03-25
Fix baseline ordering error in RFI mask

ljgray:
In #318, the `stokes_I` function was changed to maintain the standard visibility axis order. This task, which make use of that function, was never updated to accommodate this change.

I think that this is the only other place where this function is used.


ketiltrout (approved):
I think this looks okay.

PR #346, opened 2025-04-25
Ac/ps imaginary

Arnab-half-blood-prince:
This will estimate full complex correlation between two fields and gives real and imaginary part of the power spectrum


ssiegelx:
This looks good.  I've just pushed one set of changes that ensures the weights are real valued and do not inherit the complex data type from the spectrum.

ssiegelx (approved):

PR #347, opened 2025-04-25
feat(ringmapmaker): add option to skip deconvolution

ssiegelx:
Add the `skip_deconvolution` config property to `DeconvolveHybridMBase`.  If set to True, then the map maker will apply the weighting and beam model in m-mode space, but will not attempt to deconvolve the beam model.

This also adds config properties `copy_weight` and `copy_tag` to `ApplyDelayFilterHybridVis`.  If set to True, then these quantities will be copied over from the container holding the filter.  This is useful when applying the filter to the simulation, since you would like the weights to to take the exact values present in the data beyond this task.  However, maybe these operations should be moved to their own tasks?  I think in principal, `MixData` could do this, but it will create an unnecessary copy, whereas here we would just like to do this in place and overwrite the existing weights.


draco/analysis/dayenu.py
```diff
@@ -655,20 +667,30 @@ def process(self, hv, source):
...
+                            if np.any(nonzero_diag_flag):
+                                med_diag = np.median(diag[nonzero_diag_flag])
+                                flag_low = diag > (self.atten_threshold * med_diag)
+                                weight[pp, :, xx, tt] *= flag_low.astype(weight.dtype)
 
             self.log.debug(f"Took {time.time() - t0:0.3f} seconds in total.")
 
+        # If requested copy over the weight dataset
+        if self.copy_weight:
+            weight[:] = source.weight[:].local_array.copy()
```
ljgray, line 692: Is the `.copy` is needed here? Since you're writing into an existing and distinct array, I don't think that there would be any shared memory even without the copy.

ljgray (approved):
One comment but otherwise I think this is fine. I think that we would only benefit from a separate task for changing the weights if it would be used elsewhere, so for now I'm happy to just keep it within this task.

What is the use case for skipping the beam deconvolution?

Arnab-half-blood-prince (approved):

draco/analysis/dayenu.py
ssiegelx: Good call.  I've removed this copy.

ssiegelx:
Currently in the power spectrum analysis, we set the regularization parameter epsilon so large that the inverse covariance matrix C_inv is approximately equal to epsilon.  In this limit, we are effectively not performing any deconvolution of the beam.

This choice makes sense for our case: the main motivation for attempting beam deconvolution would be to redistribute foreground power to lower delays before applying a foreground filter. However, we are already working with foreground-filtered data, so there is little additional benefit to trying to deconvolve the beam at this stage. Attempting to deconvolve the beam will introduce additional noise amplification without helping our foreground mitigation.

The reason the `skip_deconvolution` property is needed (rather than simply recommending a large a large value of epsilon) is the our current method still normalizes each declination stripe independently by the peak beam response at that declination.  At declinations where the peak beam response is small (i.e., at large zenith angles), we end up dividing by a small number, amplifying the noise in those regions.

When `skip_deconvolution` is set, the map instead uses a single normalization, defined by the `reference_declination` property.  The data are then weighted by the beam response times the inverse variance, without any per-declination renormalization.  I think this gives optimal weighting for the power spectrum estimation.

PR #348, opened 2025-05-04
feat(transform): improve polarisation handling

ssiegelx:
* Extend `SelectPol` to extract arbitrary Stokes parameters (I, Q, U, V)
* Add `PolWeightedAverage` task to compute optimal pseudo-Stokes I from XX and YY.
* Modify `MixData` to support optional mixing of tags via `tag_coeff`.
* Add `broadcast_weights` utility for matching weight and data shapes.

### DIFF
--- a/draco/analysis/transform.py
+++ b/draco/analysis/transform.py
@@ -1051,34 +1051,53 @@ class ShiftRA(task.SingleTask):
 
 
 class SelectPol(task.SingleTask):
-    """Extract a subset of polarisations, including Stokes parameters.
+    """Extract a subset of Stokes parameters from beamformed data.
 
-    This currently only extracts Stokes I.
+    Supports extraction of Stokes I, Q, U, and V from beamformed data for
+    linear polarisations (XX, YY, reXY, imXY). Assumes beamformed data are
+    already calibrated and normalized.
 
     Attributes
     ----------
-    pol : list
-        Polarisations to extract. Only Stokes I extraction is supported (i.e. `pol =
-        ["I"]`).
+    pol : list of str
+            List of Stokes parameters to extract.
+            Must be a subset of ['I', 'Q', 'U', 'V'].
     """
 
     pol = config.Property(proptype=list)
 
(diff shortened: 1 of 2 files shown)


draco/analysis/transform.py
```diff
@@ -1087,31 +1106,206 @@ def process(self, polcont):
...
-                if np.issubdtype(out_dset.dtype, np.integer):
-                    out_dset[:] //= 2
                 else:
-                    out_dset[:] *= 0.5
+                    out_dset[oslc] /= nsum
+
+        return outcont
+
+
+class PolWeightedAverage(task.SingleTask):
```
ljgray, line 1214: If it isn't too difficult, I think we should combine this task with `transform.StokesIVis`, which is just doing an unweighted sum of XX and YY for vis and weights only.

ljgray (approved):
Looks good. I don't think that my comment necessarily needs to be addressed right now, but we should at least make a note to combine them at some point.

draco/analysis/transform.py
```diff
@@ -1087,31 +1106,206 @@ def process(self, polcont):
...
+            return axis, slc
+
+        # Create output container
+        outcont = containers.empty_like(polcont, pol=np.array(["I"]))
+
+        for name in polcont.datasets.keys():
+            if name not in outcont.datasets:
+                outcont.add_dataset(name)
+
+        outcont.redistribute("freq")
```
tristpinsm, line 1277: should there be a corresponding call on `poldata` to ensure both are distributed along the same axis?

ssiegelx: Good call.  This is now fixed.

draco/analysis/transform.py
ssiegelx: It’s not immediately clear to me what the best approach is here. One option might be to check whether a polarization axis exists and, if not, create one by selecting the appropriate subset of baselines. I’ve opened PR #348 to track this, but I plan to postpone further work until after the power spectrum analysis is finalized.

PR #349, opened 2025-05-04
feat(flagging): refactor mask logic, add taper support

ssiegelx:
* Replace CombineMasks with a GeneralCombineMasks task that can construct
arbitrary combinations of masks.
* Create ApplyTaper task, which applies a taper to a RingMap.
* Introduce GeneralCombineTapers task for arbitrary taper combination.
* Add MaskFromTaper task to generate binary masks from taper thresholds.
* Define new container RingMapTaper and HorizonLimit container.
* Do not create datasets if they already exist in DayenuDelayFilterHybridVis.
* Fix bug in FitBeamForm.

### DIFF
--- a/draco/analysis/flagging.py
+++ b/draco/analysis/flagging.py
@@ -2334,44 +2334,251 @@ class ApplyGenericMask(task.SingleTask):
 MaskBeamformedOutliers = ApplyGenericMask
 
 
-class CombineMasks(task.SingleTask):
-    """Combine an arbitrary number of masks (conservatively).
+class GeneralCombineMasks(task.SingleTask):
+    """Combine multiple masks using a user-specified logical expression.
 
-    All of the given masks must be of the same type and that type must have a `mask` dataset.
-    Any flagged value in any of the provided masks will be flagged in the output mask.
+    The input is a list of containers with `mask` datasets. Each mask is assigned
+    a variable name (`A`, `B`, `C`, ..., `Z`) in the order they appear. The logical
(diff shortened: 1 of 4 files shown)


draco/analysis/flagging.py
```diff
@@ -2334,46 +2334,257 @@ def process(self, data: containers.ContainerBase, mask: containers.ContainerBase
...
+
+        # Assign variables A, B, C, ..., one per mask
+        namespace = {
+            chr(ord("A") + i): mask.datasets[self._dataset_name][:]
+            for i, mask in enumerate(masks)
+        }
+
+        # Evaluate the logical expression
+        self.log.info(f"Evaluating mask combination expression: '{self.expression}'")
+        result = eval(self.expression, {}, namespace)
```
ljgray, line 2389: It's worth checking what's in `self.expression` before running `eval` since someone could in theory execute some arbitrary python code. This should be an easy check, since `self.expression` can only be a sequence of letters and logical symbols. 

I think you can also just omit the `globals` argument and provide `namespace` by keyword so it's more immediately obvious what's being passed to `eval` (I didn't know off the top of my head, so I had to look up the definition of `eval` to see what the extra arguments were).

ssiegelx: Okay I'm now validating the provided expression by comparing to a set of allowed characters and operators.  I'm also passing locals directly as a keyword argument to eval.

ljgray (approved):

PR #350, opened 2025-05-06
feat(container): add compression option to RFI mask containers

yukari-u:
Add compression option to LocalizedRFIMask and SiderealLocalizedRFIMask containers.


ljgray (approved):
Looks great@

PR #352, opened 2025-05-08
(1/?) - Move pipeline infrastructure and some basic utilities to `caput`

ljgray:
Requires radiocosmology/caput#298.

This PR is the first of several.

Here, we just move the pipeline infrastructure, namely the base task and container classes, into caput.

`truncate` is removed entirely, in favour of the implementation in `caput` (it seems like this version isn't used anywhere, and the `caput` implementation is more developed).

I've moved a few basic tasks, like file I/O, Truncate, RandomTask,... and some generic utilities, like creating a generalized window function. I think it sort of makes sense to have those in `caput` instead of `draco`, but I haven't entirely convinced myself.


ketiltrout (approved):
I think this is fine, per se, but see my comments in the caput PR about `caput.tasks.io`.

You should also update the `autosummary` in the `draco.core.__init__.py` docstring.

PR #356, opened 2025-05-28
feat: track freq-freq covariance matrix

ssiegelx:
Applying the Dayenu filter introduces off-diagonal terms in the
frequency-frequency noise covariance matrix.  Update the following
tasks to calculate this covariance matrix:

    - DayenuDelayFilterHybridVis
    - DayenuDelayFilterHybridVis
    - DelayFilterHyFoReSBandpassHybridVisClean

Also update the following tasks to propagate this covariance matrix
through the pipeline:

    - SiderealRebinner
    - SiderealStacker
    - RADependentWeights
    - SelectPol

The covariance matrix is saved as a new dataset named either
freq_cov or complex_freq_cov in the HybridVisStream or RingMap,
similar to how the filter itself is saved.

Creates two new tasks:

    - ConstructWienerDelayTransform
    - ApplyWienerDelayTransform

that implement a Wiener-filtered delay transform that:
    - handles frequency masks with non-uniform coverage
    - accounts for filtering applied to the data
    - uses the tracked frequency-frequency noise covariance
    - estimates the signal covariance from simulations

This approach avoids the need to estimate the delay power spectrum
via Gibbs+Wiener or NRML methods, instead assuming 21 cm signal + noise
and relying on the propagated frequency-domain covariances.

Creates a ReconstructVisFreqCov task that generates a FreqNoiseModel container
with the ingredients to synthesize noise with a specified frequency-frequency
covariance. It reuses logic from ReconstructVisWeight, which is now refactored
into a shared base class, ReconstructVisNoiseBase.

Creates FreqCorrelatedNoise and MultipleFreqCorrelatedNoise tasks that use
the FreqNoiseModel container to synthesize noise with the correct correlations
in frequency and el.

Finally, creates a TaperDelayTransform task that can apply a taper to the
delay transform.  The existing task will not work for this purpose due to
the way the axes are collaped in the DelayTransform container.

### DIFF
--- a/draco/analysis/dayenu.py
+++ b/draco/analysis/dayenu.py
@@ -423,6 +423,8 @@ class DayenuDelayFilterHybridVis(task.SingleTask):
     apply_filter : bool
         Apply the filter that was generated. If False, `save_filter`
         must be True.
+    save_filter : bool
+        Save the filter that was applied to the output container.
     calculate_cov : bool
         Calculate the frequency-frequency noise covariance due to filtering.
     """
--- a/draco/analysis/dayenu.py
+++ b/draco/analysis/dayenu.py
@@ -423,8 +423,8 @@ class DayenuDelayFilterHybridVis(task.SingleTask):
     apply_filter : bool
         Apply the filter that was generated. If False, `save_filter`
         must be True.
-    save_filter : bool
-        Save the filter that was applied to the output container.
+    calculate_cov : bool
+        Calculate the frequency-frequency noise covariance due to filtering.
     """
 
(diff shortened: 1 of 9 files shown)


ssiegelx:
rev_09 is the version that Arnab is using and rev_12 is the new version with the covariance matrix.

The file size increased by 20% (from 93G in rev_09 to 109G in rev_12 for the band b sidereal stack).

The run time appears to have decreased by 25%, which I don't quite understand.  This is just comparing the time that equivalent jobs took to run to create the quarterstacks.

PR #359, opened 2025-06-09
feat(beamform): more flexible ringmap-based source stacking

ssiegelx:
Update RingMapBeamForm and RingMapStack2D to handle partial maps of the sky.  Sources that fall outside of the RA and Dec range covered by the map will be excluded from the beamformed output or 2D stack.

### DIFF
--- a/draco/analysis/beamform.py
+++ b/draco/analysis/beamform.py
@@ -953,9 +953,12 @@ class RingMapBeamForm(task.SingleTask):
 
         src_ra, src_dec = self._process_catalog(catalog)
 
+        # Get the pixel indices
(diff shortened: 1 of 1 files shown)


ssiegelx:
Previous versions of RingMapBeamForm and RingMapStack2D assumed the RA axis spanned 0–360 deg and fully enclosed the catalog in both RA and elevation.  If this wasn't true, outputs would be generated but using incorrect pixel coordinates.

This update fixes that behavior. The updated RingMapBeamForm was used for the stacking paper results, but hadn't yet been pushed. RingMapStack2D is also updated here, though it still needs further testing

Arnab-half-blood-prince (approved):

PR #363, opened 2025-06-17
Mask samples with negative autos

ljgray:
In a few cases, pipeline jobs have failed due to NaN values in the sensitivity metric. I've traced this back to negative autocorrelation values introducing NaNs when taking a square root. I don't think there's any possible way for this to be physical, so we should mask these values early-on in the daily pipeline.

This task hasn't been tested at scale yet, but I think it should work fine.

### DIFF
--- a/draco/analysis/flagging.py
+++ b/draco/analysis/flagging.py
@@ -627,6 +627,42 @@ class SanitizeWeights(task.SingleTask):
         return data
 
 
+class NegativeAutosMask(task.SingleTask):
+    """Flag in frequency-time if any autocorrelation is negative."""
+
+    def process(self, data: containers.VisContainer) -> containers.RFIMask:
+        """Extract autos and flag if any auto is negative.
+
+        Parameters
+        ----------
+        data
+            Timestream dataset containing visibilities and
+            the relevant index map entries.
+
+        Returns
+        -------
+        mask
+            time-frequency mask
+        """
+        data.redistribute("freq")
+
+        # Extract autocorrelations
+        prodstack = data.prodstack
+        autos = data.vis[:, prodstack["input_a"] == prodstack["input_b"]].real
+
+        # Flag if any auto is negative and gather the mask
+        mask = np.any(autos < 0.0, axis=1).allgather()
+
+        self.log.debug(
+            f"{100.0 * mask.mean():.2f}% of data flagged due to negative autos."
+        )
+
+        mask_cont = containers.RFIMask(axes_from=data, attrs_from=data)
+        mask_cont.mask[:] = mask
+
+        return mask_cont
+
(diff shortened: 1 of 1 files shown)


draco/analysis/flagging.py
```diff
@@ -627,6 +627,63 @@ def process(self, data):
...
+
+        # Determine stack axis
+        stack_new, _ = tools.redefine_stack_index_map(
+            self.telescope, data.input, data.prod, data.stack, data.reverse_map["stack"]
+        )
+        # Construct prodstack axis
+        ps = data.prod[stack_new["prod"]]
+        conj = stack_new["conjugate"]
+
+        prodstack = ps.copy()
```
ssiegelx, line 667: Since we're only checking whether `prodstack["input_a"] == prodstack["input_b"]`, it shouldn't matter if the inputs are swapped here.

draco/analysis/flagging.py
```diff
@@ -627,6 +627,63 @@ def process(self, data):
...
+            the relevant index map entries.
+
+        Returns
+        -------
+        mask
+            time-frequency mask
+        """
+        data.redistribute("freq")
+
+        # Determine stack axis
```
ssiegelx, line 659: I don’t think `redefine_stack_index_map` changes which elements of the stack axis are autocorrelations.  It just updates which feeds are used to represent the "typical" autocorrelations, making sure they correspond to good feeds with valid positions, polarisations, etc. Since we’re not using any of the feed info here, I think we could skip this step and just use `data.prodstack` directly. That would also mean we don’t need to provide a telescope instance at setup. Let me know if I’m overlooking something!

draco/analysis/flagging.py
```diff
@@ -627,6 +627,63 @@ def process(self, data):
...
+        # Construct prodstack axis
+        ps = data.prod[stack_new["prod"]]
+        conj = stack_new["conjugate"]
+
+        prodstack = ps.copy()
+        prodstack["input_a"] = np.where(conj, ps["input_b"], ps["input_a"])
+        prodstack["input_b"] = np.where(conj, ps["input_a"], ps["input_b"])
+
+        # Extract the autocorrelations
+        autos = data.vis[:, prodstack["input_a"] == prodstack["input_b"]].real
```
ssiegelx, line 672: I was worried about what would happen here if there were no autocorrelations in data, but after some testing it seems like numpy handles this fine.

ssiegelx (approved):
I think the construction of the `prodstack` axis could be simplified here, but also fine leaving it as is if you prefer.

ljgray:
> I think the construction of the `prodstack` axis could be simplified here, but also fine leaving it as is if you prefer.

Done

PR #364, opened 2025-06-20
Preserve gain mask through smoothing when applying gains

ljgray:


jmaceachern (commented):
Looks good. I do have one question:

Sorry if I'm a bit out of the loop on this, but doesn't smoothing the weights like this mix masked values in the weights (= 0) into neighbouring unmasked values in the weights?

I guess this sort of apodizes the weights (makes them smoothly roll off to 0 near the masked regions), so maybe this is actually desirable? But I think setting the masked values back to 0 after smoothing (as has been implemented in this PR) would introduce discontinuities again. Is this a potential issue? 

Again, sorry if I'm out of the loop on this!

jmaceachern:
Got it. That makes sense. Weighted median filter is a great idea. I can approve this if you want to keep this PR as a minimal change and do the weighted median in a separate PR?

draco/core/misc.py
```diff
@@ -115,20 +113,33 @@ def process(self, tstream, gain):
...
                     if weight_arr is not None:
+                        # Smooth
                         shp = weight_arr.shape
-                        weight_arr = ss.medfilt2d(
-                            weight_arr.reshape(-1, shp[-1]), kernel_size=[1, l]
+                        weight_arr = weighted_median.moving_weighted_median(
+                            weight_arr.reshape(-1, shp[-1]), weights=wmask, size=(1, l)
                         ).reshape(shp)
+                        # Ensure flagged values remain flagged
+                        weight_arr[wmask == 0] = 0.0
```
jmaceachern, line 142: This is probably not an issue, but I think `wmask` elements are type float64 and they're being compared to 0 with `==`. If you think that's a concern, maybe it's safer to save a copy of `weight_arr <= 0.0` as type `bool` before you start editing `weight_arr`?

jmaceachern: But I really think this is fine. Just wanted to point it out.

jmaceachern (approved):

draco/core/misc.py
ljgray: I _think_ that casting from boolean False to float should preserve that float evaluating to zero. Either way, technically there's already a float comparison to zero earlier `(weight_arr > 0.0)`, so I don't think there should be any issues.

jmaceachern: Good point about casting and the earlier comparison. Makes sense.

PR #368, opened 2025-07-14
adding noise scale factor

Arnab-half-blood-prince:
This will generate a scale factor based on even minus odd jackknife map and save it in a DelayTransform container and later that scale factor can be used to scale the noise delay transform map.


Arnab-half-blood-prince:
okay..I changed the tasks. I am now saving the scale factor into a RingMap container and apply the scale factor to the DelayTransform of noise.  However, when I merge the master to this, I see all the files have changed, but it should not. I am not sure why it happened. I can close this PR and regenerate another one.

Arnab-half-blood-prince:
But it seems the changes in all other files are none, at least it did not change anything

PR #373, opened 2025-08-13
build(pyproject): add dnspython to requirements

rikvl:
I think this gives a dependency on `dnspython`
https://github.com/radiocosmology/draco/blob/7307bfb2b3a613daddb8ecc73069ca2efd8b0004/draco/core/misc.py#L376


rikvl:
(Ran into this when trying to run `caput-pipeline` on Fir)

PR #381, opened 2025-10-04
Static frequency mask estimation from visibility data

ljgray:
The overall goal of this PR is to implement a task to come up with a static RFI mask from visibility data directly, without having to rely on a hard-coded set of bad channels.

This produces a flagging metric by reducing over baselines using the existing `ReduceChisq` code as follows:
- The chi2 is computed over cross-pol baselines only, and excludes baselines < 5 meters.
- Weighting is done by dividing by baseline redundancy (effectively up-weighting baselines with lower redundancy), which highlights faint frequency bands which are only otherwise visible in the high-delay chisq metric.

Flagging is then done using a simple 1D MAD metric over the 1D median in time of the chisq metric. Only nighttime data is included in the median (the hook is implemented in a `ch_pipeline` PR). It uses an updated version of the `arPLS` baseline-fitting code, which is better able to handle the lack of hard-coded static mask.

This PR does the following:
- Updated the `scale_invariant_rank` operation to be a bit faster
- Adds options to the `MaskBaselines` task to mask polarisation and to optionally combine the mask using an `and` operation
- Extends the `ReduceBase` task with a hook for custom weights
- Updates the `penalized_least_squares` implementation to support different weighting schemes
- Replaces the old narrowband vis flagging task with this new static RFI mask task

This code all works and is tested. This method does not appear to overflag, although there is absolutely room for improvement.

### DIFF
--- a/draco/analysis/beamform.py
+++ b/draco/analysis/beamform.py
@@ -1442,7 +1442,6 @@ class HybridVisBeamForm(tasklib.base.ContainerTask):
 
         # Loop over sources and fringestop
         for ss, (idec, sdec) in enumerate(zip(nearest_dec, np.radians(src_dec))):
-
             in_range = np.flatnonzero(valid[ss])
             if (in_range.size == 0) or not valid_src[ss]:
                 continue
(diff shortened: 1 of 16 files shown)


tristpinsm:
sorry, was hoping to take a look this pm

ljgray:
> sorry, was hoping to take a look this pm

No worries. It won't break anything, so it's really just a review of what's being implemented in this task, which I'd ideally like to improve on in the future anyway.

PR #393, opened 2026-01-20
Robust check for compiler openmp support

ljgray:
Provides a more robust check for compiler openmp support. Tests are passing, and I've confirmed that it behaves as expected on macos using `clang` (i.e., without openmp).

Also, pinned `sphinx<9`, as there are a number of breaking changes introduced.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,7 +50,7 @@ classifiers = [
 [project.optional-dependencies]
 lint = ["black", "ruff"]
 test = ["pytest", "wheel"]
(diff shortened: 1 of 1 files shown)


ljgray:
> This looks fine, but the envar `DRACO_NO_OPENMP` should be documented somewhere if it isn't already. I've left a suggestion below (which you don't need to take).

Good point - I just added this because we already have equivalent envars for cora and caput (CORA_NO_OPENMP, ...). However, I don't think that any of those are actually documented...

ketiltrout (approved):
Yeah, that's a cleaner way to do it.

PR #400, opened 2026-07-13
Skip DPSS interpolation where cholesky fails

ljgray:
Minor fix that I forgot to merge awhile back


draco/util/dpss.py
```diff
@@ -220,7 +224,11 @@ def solve(
         # Cholesky decomposition. This is faster than
         # doing a standard solve, and significantly
         # faster than an inverse
-        CiL = la.cho_factor(Ci.astype(cho_dtype), lower=False, check_finite=False)
+        try:
+            CiL = la.cho_factor(Ci.astype(cho_dtype), lower=False, check_finite=False)
+        except np.linalg.LinAlgError:
+            logger.debug(f"Error inverting Ci in iteration {ii}/{xp.shape[0] - 1}")
+            continue
```
ketiltrout, line 233: What's the value of `Ni[ii]` if the rest of this `ii` step is skipped?  Is it going to cause trouble down the line?

ketiltrout: It may at least be worth a comment before the `continue` indicating what the effect of the cholesky decomp failing will be.

ketiltrout (approved):

